Issue #127, opened 2019-03-22
Spaces are not supported in mocks resource paths


lbroudoux:
When having REST API with resource path like `/pastry/Eclair%20Cafe`, the exposed Mock is not able to compare this resource path to the dispatch criteria being `/pastry/Eclair Cafe`.

The whole resource path should be URL decoded.

lbroudoux:
Fixed.

Issue #126, opened 2019-03-20
OpenAPIImporter.java detection of yaml vs json is flawed


sercasti:
Using this logic:
```
        while ((line = reader.readLine()) != null) {
            line = line.trim();
            // Check is we start with json object or array definition.
            if (line.startsWith("{") || line.startsWith("[")) {
               isYaml = false;
               break;
            }
            if (line.startsWith("---")) {
               isYaml = true;
               break;
            }
         }
```
To detect the file format is prone to fail. My import job started failing this morning for my api.yaml file when a new revision was commited, as it looks like this now:
```
openapi: 3.0.2
info:
    title: Five
paths:
    /api/notifications:
        summary: Notifications
        description: Return a user list notifications
        get:
            summary: Returns a list of notifications
            responses:
                '200':
                    description: List of notifications
                    content:
                        application/json:
                            schema:
                                type: array
                                items:
                                    $ref: '#/components/schemas/Notification'
                            examples:
                                return one notification:
                                    value: |-
                                        [
                                            {
```

Those two lines starting with [ and { made microcks think I'm using a json file, and fail to execute the ObjectMapper.

What made this worse is that code on line 93 is swallowing the real exception, so I had to debug the code to figure out the root cause:
```
      } catch (Exception e) {
         throw new IOException("OpenAPI spec file");
      }
```

lbroudoux:
Thanks for opening this issue ;-)

lbroudoux:
Now fixed with new unit test.

PR #133, opened 2019-04-02
Remove the OpenAPI version check at importer instantiation time

### LINKED ISSUE #132: OpenAPI Version Check is Too Strict
When the OpenAPI version is quoted, Microcks fails to upload the API, despite displaying a partial success message:
```
openapi: '3.0.0'
info:
  description: 'Super bare bones API'
  version: '1.0.0'
  title: 'Example Microcks API'
```
Results in: 
image
In this case there is no new entry in the API & Services tab, which is unexpected.

Unquoting the version:
```
openapi: 3.0.0
info:
  description: 'Super bare bones API'
  version: '1.0.0'
  title: 'Example Microcks API'
```
Results in:
image
In this case the new service appears in the API & Services tab as expected.

This seems related to #107 but not entirely since in this case the version *is* valid

yahehe:
Resolves #132 

The OpenAPI version is checked before the OpenAPIImporter is instantiated, so it is known to be valid. Additionally, the presence of quotations in the spec causes uploads to report successfully despite the absence of a mocked server. It looks like this is probably due to a file pointer overrun where isYaml remains null, though this has not been verified.

### DIFF
--- a/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
+++ b/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
@@ -71,7 +71,7 @@ public class OpenAPIImporter implements MockRepositoryImporter {
                isYaml = false;
                break;
             }
-            else if (line.startsWith("---") || line.startsWith("openapi: 3")) {
+            else if (line.startsWith("---") || line.matches("(openapi: ["']?3.*))") {
                isYaml = true;
                break;
             }
--- a/src/test/resources/io/github/microcks/util/openapi/cars-openapi-with-json.yaml
+++ b/src/test/resources/io/github/microcks/util/openapi/cars-openapi-with-json.yaml
@@ -85,4 +85,4 @@ components:
       value: laurent
 tags:
 - name: car
-  description:
+  description: Car Ownership


lbroudoux:
Thanks for the PR. However your changes invalidate #127 and break unit tests.

src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
```diff
@@ -71,7 +71,7 @@ public OpenAPIImporter(String specificationFilePath) throws IOException {
                isYaml = false;
                break;
             }
-            else if (line.startsWith("---") || line.startsWith("openapi: 3")) {
+            else if (line.startsWith("---")) {
```
lbroudoux: Revert #126 that was legit because starting with dashes is not mandatory in YAML

lbroudoux: Something like `|| line.startsWith("openapi: ")` would not invalidate #126 and make the check less strict. Would you want to update the PR ? Or I'll push something on my side ?

yahehe: Oooh okay, yeah that makes sense. I can update this today, sorry I didn't notice the failing test

lbroudoux:
Thanks for PR!

PR #136, opened 2019-04-25
Spec body should also use UTF8 encoding

askfor:
Currently, when saving spec body, no encoding was specified. So the spec may be saved in wrong encoding.


lbroudoux:
Thank you very much @askfor for this contribution ! Greatly appreciated !

Issue #138, opened 2019-05-03
Hardcoded Scheduler to update definitions should be configurable


sercastiGL:
I'm working on a team that uses mock responses intensively, and waiting for 2 hours for microcks to update service definitions is just painful. This value should be configurable:
 https://github.com/microcks/microcks/blob/2e982996d4be02e88213185a8f0a4f6916070ab9/src/main/java/io/github/microcks/task/ImportServiceDefinitionTask.java#L72

I'll post a PR to fix this, if you agree

lbroudoux:
Thanks @sercastiGL for this enhancement! Will have a look at it as soon as possible.

lbroudoux:
Now merged.

PR #139, opened 2019-05-03
Make services update scheduler configurable, see https://github.com/m…

sercasti:
Fix fox https://github.com/microcks/microcks/issues/138


lbroudoux:
Thanks @sercasti for this contribution !

PR #155, opened 2019-06-13
Picking mock result based on Accepted content type

pietiepb:
The mock is ignoring the accept header of the request, although they are supported by OAS3.0.

With this little change, the response with the matching vendor specific type is returned. If it is not found or not set, the default behavior is executed, return the first one.


lbroudoux:
Hi @pietiepb ! Thanks a lot for the PR and sorry for delay but I was taking a summer break. Now I'm back and will merge it very soon. Thanks again !

Issue #119, opened 2019-02-07
Separate URLS for keycloak redirection and keycloak internal validation


sercastiGL:
This is related to https://github.com/microcks/microcks/issues/14

a) The config entries for KeyCloak URL should be separate, one for the login redirection, and a different other variable for internal communication, just like these guys did:
https://github.com/hobbit-project/platform/issues/18#issuecomment-341440232
      - KEYCLOAK_AUTH_URL=http://www.mypublicinternetpage.com/auth
      - KEYCLOAK_DIRECT_URL=http://myinternatIP:8080/auth

The solution to #14 will not work as my web instance will not run on localhost, as this a server/public instance, so I need, a public URL for login redirection, and a private URL for internal communication. 

b) https://github.com/microcks/microcks/issues/14 mentions "Keycloak protection is disabled", but there is no example on how to do that. There should be an additional       
        - CHECK_REALM_URL=false

sercastiGL:
Steps to reproduce on a LInux VM (Red Hat)

1. `uname -a`
Linux fivedevserver.synapse.com GNU/Linux
2. `hostname -f`
fivedevserver.synapse.com
3. `git clone https://github.com/microcks/microcks.git`
4. `cd microcks/install/docker-compose`
4. `cp keycloak-realm/microcks-realm-sample.json keycloak-realm/microcks-realm-sample.json.bak`
5. `yum install jq`
6. `jq '.applications |= map(if .name == "microcks-app-js" then .redirectUris = [ "http://fivedevserver.synapse.com:8080/*" ] else . end) | .sslRequired |= "none"' keycloak-realm/microcks-realm-sample.json.bak > keycloak-realm/microcks-realm-sample.json`
7. `perl -i.bak -pe 's|KEYCLOAK_URL=http://localhost:8180/auth|KEYCLOAK_URL=http://fivedevserver.synapse.com:8180/auth|' microcks.yml`
8. `perl -i.bak -pe 's|keycloak.ssl-required=external|keycloak.ssl-required=none|' config/application.properties`
9. `sudo docker-compose -f microcks.yml up -d`

lbroudoux:
Reopen this issue to finalize installation procedure changes

PR #296, opened 2020-11-22
Simple docker-compose

hguerrero:
Update the docker-compose setup:

- change to docker-compose.yaml
- remove the https config
- mount realm from volume 
- use generic name for host hostname


lbroudoux:
Hi @hguerrero 

And thanks for PR ! I've tested everything on my Mac and its way easier than I was previously achieved ... I'd like to make a test on a Linux box to be sure that everything is also ok and I'll merge it.

install/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: '2'
+
+services:
+  mongo:
+    image: mongo:3.4.23
+    container_name: microcks-db
```
lbroudoux, line 6: Could you change to the old name that is `microcks-mongo`? That way we know it's the container attached to the `mongo` alias.

install/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    volumes:
+      - "~/tmp/microcks-data:/data/db"
+
+  keycloak:
+    image: jboss/keycloak:10.0.1
+    container_name: microcks-sso
```
lbroudoux, line 12: Could you change to the old name that is `microcks-keycloak`? That way we know it's the container attached to the `keycloak` alias.

lbroudoux (changes requested):
Just 2 changes regarding container names.

install/docker-compose/docker-compose.yml
hguerrero: This name is shown only when interacting with docker like in `docker ps` the alias used in the network level is still `mongo`. Is there a specific need to expose the implementation detail on db and sso ?

hguerrero:
@lbroudoux I think in the case of Linux box you will need to configure the host IP address instead to make it work as I don't recall there is a entry created in the `/etc/hosts` file.

hguerrero:
@lbroudoux this PR now has the updated config required to address #119 and set up the correct external Keycloak URL and frontend hostname.

lbroudoux:
Yes just saw it. Perfect !

install/docker-compose/docker-compose.yml
lbroudoux: I use to have many mongo containers running or in the `docker ps` history and love to see that one is coming from microcks. But I'll change it afterwards. No problem ;-)

lbroudoux (approved):

lbroudoux:
Thanks a lot @hguerrero !!

PR #352, opened 2021-02-22
Fix podman-compose (missing config volume)

nmasse-itix:
The podman-compose files were broken due to a dangling "config" symlink (removed from docker-compose)


lbroudoux:
Thanks a lot @nmasse-itix for the fix!

Issue #320, opened 2021-01-04
Microcks UI is not loading on behind Azure application gateway Ingress


meyyazhagan:
I tried to deploy Microcks on AKS along with application gateway as an Ingress.
I could see all the pods are running and ingress is route the microcks requests to host.

Below are the log outputs,
**POD status**
NAME                                           READY   STATUS    RESTARTS   AGE
microcks-5f6d6dc94b-9p95p                      1/1     Running   0          18d
microcks-keycloak-69985858b9-vg82r             1/1     Running   3          18d
microcks-keycloak-postgresql-bd87d8fc7-pz8sq   1/1     Running   0          18d
microcks-mongodb-59f64665c4-fctk5              1/1     Running   0          17d
microcks-postman-runtime-5bf9944956-wm4lr      1/1     Running   0          18d

**Ingress microcks Annotations**
Name:             microcks
Namespace:        XXX
Address:          X.X.X.X
Default backend:  default-http-backend:80
Rules:
  Host                             Path  Backends
  XXX.org
                                   /microcks/*   microcks:8080 (172.16.1.126:8080)
Annotations:                       appgw.ingress.kubernetes.io/appgw-ssl-certificate: default
                                   appgw.ingress.kubernetes.io/backend-path-prefix: /
                                   appgw.ingress.kubernetes.io/request-timeout: 120
                                   appgw.ingress.kubernetes.io/ssl-redirect: true
                                   kubernetes.io/ingress.class: azure/application-gateway
                                   meta.helm.sh/release-name: microcks
                                   meta.helm.sh/release-namespace: XXX
Events:                            <none>

In the depth dive, I came to know that the URL is redirection is not happening .
Meaning, the context path for Microcks need to be set to run the microcks behind the reverse proxy.

```
When hit the URL https://xxx.org/microcks/ and getting below exceptions.

Loading failed for the <script> with source “https://XXX.org/main.3bf80884829c016c441a.js”. microcks:15:1
Loading failed for the <script> with source “https://XXX.org/runtime.0b05ef57c964121efcc0.js”.

**Microcks path is missing in URL** 
URL suppose to be  https://XXX.org/microcks/runtime.0b05ef57c964121efcc0.js
but getting URL is https://XXX.org/runtime.0b05ef57c964121efcc0.js  (Microcks is missing)
Looks context path is taken correctly to load the JS & CSS.
```

Can anyone help me with more lights how to context path on Microcks application?
Seeking for support, please advise.

[...]

PR #357, opened 2021-03-05
Add configurable base context to angular and spring app

fhopfensperger:
This PR allows to configure a subcontext (sub http path) for the Angular and Spring boot app and Keycloak. For example, the services then listen to the following urls: `https://example.com/mock-server/auth/` & `https://example.com/mock-server/microcks/`

Work is still in progress, but I already have a working deployment up and running on my Azure Kubernetes cluster with Istio. 

Signed-off-by: Florian Hopfensperger <florian.hopfensperger@de.ibm.com>

### DIFF
--- a/webapp/src/main/resources/config/application.properties
+++ b/webapp/src/main/resources/config/application.properties
@@ -1,22 +1,17 @@
 # Application configuration properties
 spring.jackson.serialization.write-dates-as-timestamps=true
 spring.jackson.default-property-inclusion=non_null
-
 tests-callback.url=${TEST_CALLBACK_URL:http://localhost:8080}
 postman-runner.url=${POSTMAN_RUNNER_URL:http://localhost:3000}
 async-minion.url=${ASYNC_MINION_URL:http://localhost:8081}
-
 network.proxyHost=${PROXY_HOST:}
 network.proxyPort=${PROXY_PORT:}
 network.proxyUsername=${PROXY_USERNAME:}
 network.proxyPassword=${PROXY_PASSWORD:}
 network.nonProxyHosts=${PROXY_EXCLUDE:localhost|127.0.0.1|*.svc.cluster.local}
-
 validation.resourceUrl=http://localhost:8080/api/resources/
 services.update.interval=${SERVICES_UPDATE_INTERVAL:0 0 0/2 * * *}
 mocks.rest.enable-cors-policy=${ENABLE_CORS_POLICY:true}
-
-
 # Keycloak adapter configuration properties
 keycloak.auth-server-url=${KEYCLOAK_URL:http://localhost:8180/auth}
 keycloak.realm=microcks
@@ -24,7 +19,9 @@ keycloak.resource=microcks-app
 keycloak.use-resource-role-mappings=true
 keycloak.bearer-only=true
 keycloak.ssl-required=external
-
+keycloak.cors=true
+keycloak.cors-max-age=3600
+keycloak.cors-allowed-methods=POST, PUT, DELETE, GET, OPTIONS
 keycloak.security-constraints[0].authRoles[0]=admin
 keycloak.security-constraints[0].authRoles[1]=manager
 keycloak.security-constraints[0].authRoles[2]=user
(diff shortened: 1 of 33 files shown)


webapp/src/main/webapp/angular.json
```diff
@@ -80,9 +80,7 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "microcks-ui:build",
-            "proxyConfig": "proxy.conf.json"
-          },
+            "browserTarget": "microcks-ui:build"          },
```
lbroudoux, line 83: `proxy.conf.json` is necessary while in development mode. I use `ng serve` to serve the frontend with hot-reload and backend is ran with `mvn spring-boot:run`

lbroudoux (commented):
Had some questions about required changes ...

fhopfensperger:
Hi @lbroudoux 
First of all thank you for the great product. You really surprised me now because at the moment I'm still working on the helm chart when my daily business allows it.

Main reason for the PR was the fact that I have a Kubernetes cluster that has a single external DNS. The DNS entry is routed to the Istio ingress gateway, which then decides which service to route the request to based on sub-path. So unfortunately I cannot implement what has been described in #320, because I don't "have" an "typical" ingress controller. Also when I started the PR, there was no solution for the issue. 😅

Right now I have the implementation running in an AKS cluster and am in the process of testing the same locally on my minikube and CRC with and without sub path.

> Hi @fhopfensperger
> 
> Could you explain the motivation with this PR? We had previous discussion (like for example in #320) about that and ended-up with it not being a harde requirement because you could mostly always find a gateway configuration to handle that. Is there some specific issues we wouldn't be able to handle ?
> 
> PR seems to involve a large set of resources...

fhopfensperger:
> Had some questions about required changes ...
> Also it would be nice removing unnecessary formatting changes (like introduced spaces or new lines) to ease-up a complete review later.

Would love to fix the formatting issues, what type of formatter do you use for the Typescript and Java code?

webapp/src/main/resources/config/application.properties
```diff
@@ -34,29 +31,25 @@ keycloak.security-constraints[0].securityCollections[0].patterns[1]=/api/service
...
 async-api.default-frequency=3
-
 # Kafka configuration properties
 spring.kafka.producer.bootstrap-servers=${KAFKA_BOOTSTRAP_SERVER:localhost:9092}
 spring.kafka.producer.key-serializer=org.apache.kafka.common.serialization.StringSerializer
-spring.kafka.producer.value-serializer=io.github.microcks.event.ServiceViewChangeEventSerializer
\ No newline at end of file
+spring.kafka.producer.value-serializer=io.github.microcks.event.ServiceViewChangeEventSerializer
+# Expose Angular environment configuration
+spring.resources.static-locations=classpath:/META-INF/resources/,classpath:/resources/,classpath:/static/,classpath:/public/,file:/deployments/assets/
```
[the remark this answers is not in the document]
fhopfensperger: By default ([baeldung](https://www.baeldung.com/spring-mvc-static-resources#spring-boot)) Spring doesn't serve static files from `/deployments/asssets/*` . 
I use this path to mount a `ConfigMap` (which holds the Angular `environment` configuration) as a volume to the running pod. Which then allow the Angular App to get the Backend URL and other configuration properties.

https://github.com/fhopfensperger/microcks/blob/4e3abe87d93030588c2ebc6bd50f115db4a5282c/install/kubernetes/microcks/templates/configmap.yaml#L390
https://github.com/fhopfensperger/microcks/blob/4e3abe87d93030588c2ebc6bd50f115db4a5282c/webapp/src/main/webapp/src/environments/environment-loader.ts#L8
https://github.com/fhopfensperger/microcks/blob/4e3abe87d93030588c2ebc6bd50f115db4a5282c/install/kubernetes/microcks/templates/deployment.yaml#L114

webapp/src/main/webapp/angular.json
fhopfensperger: Actually i do the same with `ng serve` and `mvn spring-boot:run` but I changed that the Angular services do not use the hardcoded `/api/` path instead the (https://github.com/fhopfensperger/microcks/blob/4e3abe87d93030588c2ebc6bd50f115db4a5282c/webapp/src/main/webapp/src/app/services/importers.service.ts#L27) services load the actual address + path from the ` environment`. Because of this I have configured the `environment.ts` file and removed the not longer needed `proxy.conf.json`:
```
export const environment = {
  production: false,
  apiUrl: 'http://localhost:8080/',
  baseContext: '/'
};
```

webapp/src/main/resources/config/application.properties
```diff
@@ -1,30 +1,27 @@
...
 keycloak.auth-server-url=${KEYCLOAK_URL:http://localhost:8180/auth}
 keycloak.realm=microcks
 keycloak.resource=microcks-app
 keycloak.use-resource-role-mappings=true
 keycloak.bearer-only=true
 keycloak.ssl-required=external
-
+keycloak.cors=true
+keycloak.cors-max-age=3600
+keycloak.cors-allowed-methods=POST, PUT, DELETE, GET, OPTIONS
```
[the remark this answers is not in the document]
fhopfensperger: I noticed that all APIs that are secured by Keycloak could no longer be called locally by the Angular app in development mode. Background is (see comment on proxy.conf), the angular app is accessible at http://localhost:4200 and the spring boot app is accessible at http://localhost:8080. So we would have a CORS issue. [Link to the Keycloak issue.](https://lists.jboss.org/pipermail/keycloak-user/2016-November/008208.html)

webapp/src/main/webapp/angular.json
fhopfensperger: But something just comes to my mind, we could define in the environment only the `/` as `apiUrl` and keep the proxy.conf. Maybe we can fix the CORS issue as well. If you want I can try it out.

fhopfensperger:
Short update from my side. The current implementation (Rest Mocking functionality) has been successfully tested on the following platforms and ingress types:
- Azure Kubernetes Service: Istio Ingressgateway with workload routing based on a base context like  `/mock-server/microcks/` using `VirtualServices`
- Minikube: Ingress Nginx with and without base context
- OpenShift on IBM Cloud: OpenShift Router with TLS edge-termination; with and without base context

I have not tested the `Async` and the `Testing` feature. Could you help me with this? If you want to test the current implementation yourself, feel free to use the container image `docker.io/fhopfensperger/microcks:1.1.0`. (Please ignore the tag that I just randomly selected).

keiranbatten-nzta:
@fhopfensperger @lbroudoux what is the reason for the inactivity on this PR?

lbroudoux:
Hi,
We didn't have any other requests for this feature and got focused on some other ones. Is this something that would be still useful?

lbroudoux:
I just updated to branch `1.7.x` where current development goes to refresh the impact of the Pull Request

keiranbatten-nzta:
Hi,

Thanks for your response. I would definitely find this useful. We are looking to deploy in Azure Kubernetes Service using application gateway ingress controller and the only option we have is to run it at a different context path than root. Keycloak now supports this using the --hostname-url and --admin-hostname-url options in the start command or similarly by setting environment variables.

mgvinuesa:
I think this pull request is pretty interesting, allow configure base path helps in some organization where using the root is not allowed.

Issue #356, opened 2021-03-03
SOAP 1.2 calls always return "text/xml;charset=UTF-8" in the Content-Type


fgoforth:
Steps to recreate:
Create a mock SOAP 1.2 service in SoapUI and save the project
Import the project into Microcks
Create a SOAP client using Spring Boot that requires SOAP 1.2 requests and responses
Use the Spring Boot client to make a call to the imported mock SOAP service
The client will fail because it was expecting "application/soap+xml" in the content type and received "text/xml;charset=UTF-8"

Expected result
The Microcks service should have returned "application/soap+xml" as the Content-Type per SOAP 1.2 specifications.

Suggested short term fix:

The SoapController class, on line 194, purposefully puts "text/xml;charset=UTF-8" into the Content-Type.  Instead, check for the request header Content Type; if it starts with "application/soap+xml", return that, otherwise return the current hard coded value.

Suggested long term fix:
See if there is a way, upon importing the SoapUI project to determine the SOAP 1.1 and 1.2 requirements and return an implementation that handles it appropriately.  I tried hard coding the Content-Type in SoapUI for the response, but it was ignored by Microcks even though it showed my hard coded value for the response in the mock service request tab.

Java 11
SoapUI 5.6.0;
Windows 10 Enterprise
Docker Desktop with Linux Containers WSL 2

PR #358, opened 2021-03-06
Handle SOAP 1.2 requests (Issue #356)

fgoforth:
This fix does a simple check to see if the request Content-Type was SOAP 1.2 based and returns that Content-Type value if true.  It assumes the creator of the Mock SOAP service has used the correct SOAP envelope namespace value to match the response type (1.1="xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/", 1.2=xmlns:soap="http://www.w3.org/2003/05/soap-envelope").

It would be helpful if the SoapUI instructions included a paragraph about the above + some other things I had to discover the hard way (e.g. SoapUI script dispatching needs to return a specific response and cannot use scripting to throw an error; when looking for an attribute name, you have to use the namespace value if present)

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/SoapController.java
+++ b/webapp/src/main/java/io/github/microcks/web/SoapController.java
@@ -189,9 +189,16 @@ public class SoapController {
             response = responses.get(0);
          }
 
-         // Set Content-Type to "text/xml".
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
Thanks a lot for the PR! That's great!

PR #362, opened 2021-03-18
fix: Changed kubeVersion requirement

195858:
Required to be able to run helm charts against EKS, GKE - to avoid errors like this:

Error: chart requires kubeVersion: >=1.17.0 which is incompatible with Kubernetes v1.19.6-eks-49a6c0

### DIFF
--- a/install/docker-compose/docker-compose-async.yml
+++ b/install/docker-compose/docker-compose-async.yml
@@ -1,0 +1,122 @@
+version: '2'
+
+networks:
+  default:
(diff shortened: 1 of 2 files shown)


195858:
Unfortunately I included an additional file in this PR, I'll reopen with the change as intended instead.

lbroudoux:
Thanks ! I also had a WIP for adding the async part within the docker-compose so better to merge efforts on a dedicated PR for that.

PR #370, opened 2021-03-19
Use same password as operator install

hguerrero:
change the docker-compose user's password to use the same as the operator.


lbroudoux:
Thanks @hguerrero for the update!

PR #378, opened 2021-04-06
Replacement character in test case id URL cause problems with APIs using _ in param names

195858:
The "_" replacement used caused problems with APIs using underscore in the parameter names. Switched to safer alternative - "!"


195858:
Commit id 2e46fe7 - those are the two relevant file changes.

lbroudoux (approved):
Thanks !

lbroudoux:
Thanks a lot for this PR!

Issue #367, opened 2021-03-19
Suggestion - add config param to switch on/off addition of API name, version to AsyncAPI channels/REST URLS 


195858:
> > I'd argue information like this should be included in the channel names in the AsyncAPI - it should include an API namespace and version. Similar to REST/OpenAPI conventions.
> > In other words - this problem should be addressed by a well designed topic namespace. Otherwise you'd need to re-publish
> 

> Agree that's in true for production and public communication purposes. However I do not see any other mean to avoid such collisions at development time... We applied the same mechanisms with REST mocks - adding a prefix to paths defined within OpenAPI contract

Separated this in to a different issue.

Understand why this is done in dev phase - our use case is slightly different.

We are looking at this from a Dev Portal/Onboarding perspective.

A developer wants to test drive Async APIs either as a quick "try-it-out" similar to Swagger ... so quickly open a subscription to see live events coming through in a sandbox.
Then next step - developed an app based on the AsyncAPI and want to test drive that they receive events and process them correctly.

The idea would be to provide a sandbox broker environment that is fed by microcks mocks. IN this case we actually want the mocks to look exactly like the actual API as exposed in a production system.

Understand if you don't want to support a use case like this - if you think it's a valid case then way to toggle this behaviour would be helpful.

PR #384, opened 2021-04-08
Configurable Async API Test Behaviour

195858:
Motivation for these changes are described in #367.
The changes are configurable, the default behaviour has not changed.

- Added an option to use a "bare" namespace for Async API channels - this omits the service name and version prefix and uses the Async API channel names as-is. We require this behaviour for our use case. It only applies to the Async API behaviour. It is relevant in two places - the example message view in the service page where the relevant mock-topic is displayed and in the async minion where the topic names are generated for message publication.
- Added an option to specify for which operation type (PUBLISH, SUBSCRIBE) mock messages should be scheduled. The async-api.default-frequency is used for the operations of the type/method in the new property, the other operations are set to 0 (mock messages are disabled).  If the property is omitted the current behaviour (schedule for both types) is preserved.
The reason for this change: an Async API spec describes the interaction between a client and a broker from the brokers point of view. Publish means "the broker publishes this to the client", subscribe means "the broker subscribes/receives  messages form the client". So in order to mock the "broker" side we essentially only want to publish mock messages for the "publish" operation type.


commons/util/pom.xml
```diff
@@ -52,20 +52,24 @@
       <artifactId>avro</artifactId>
       <version>${avro.version}</version>
     </dependency>
-
+    <dependency>
```
195858, line 55: After moving the URIBuilder I had to adjust the dependencies - it required the microcks-model as well as the spring-boot-starter-web dependency to build successfully.

minions/async/src/main/java/io/github/microcks/minion/async/producer/BindingProducerManager.java
```diff
@@ -0,0 +1,110 @@
...
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+import io.github.microcks.domain.EventMessage;
+import io.github.microcks.minion.async.AsyncMockDefinition;
+import io.github.microcks.util.URIBuilder;
+import io.github.microcks.util.el.TemplateEngine;
+import io.github.microcks.util.el.TemplateEngineFactory;
+
+public abstract class BindingProducerManager {
```
195858, line 15: I  introduced this base class to remove duplicate code (topic string creation) and in order to simplify the ProducerManager class (the case statement switching between bindings).

There's a template method publishMessages that makes sure topic strings are created and messages are sent.
Sub-classes can extend the topic generation (see the KafkaProducer) and need to implement a publishOne() method.

Issue #451, opened 2021-07-16
How to start microcks with kafka using docker


tillias:
Hello folks, many thanks for a very nice product!

What is the easiest way to start microcks with kafka using docker compose? I'd like to mock few asyncAPIs locally 

Many thanks in advance.

tillias:
I have tried adding few env variables, but it doesn't work

  app:
    depends_on:
      - mongo
      - keycloak
      - postman
    image: quay.io/microcks/microcks:latest
    container_name: microcks
    ports:
      - "8080:8080"
      - "9090:9090"
    environment:
      - KAFKA_INSTALL=false
      - FEATURES_ASYNC_ENABLED=true
      - FEATURES_ASYNC_KAFKA_URL=kafka:9092

lbroudoux:
Whilst begin reluctant at first on implementing this - as it is an additional configuration to support with many consumed resources - there's frequent demand to support this. We'll provide an alternative docker-compose file with the async features on.

tillias:
Hi Laurent, if it helps, there is Kompose: https://kubernetes.io/docs/tasks/configure-pod-container/translate-compose-kubernetes or https://github.com/kubernetes/kompose

So probably it is easier to make docker-compose first and then translate it into openshift / kubernetes?

lbroudoux:
Hi @tillias ! Thanks for the link. I found since our last conversation that I had already included the property to configure the Keycloak URL from the minion side. I am finalizing some tests but things seem to work well till now. I should commit some stuffs in few minutes ;-)

lbroudoux:
So to use if you just have to `cd install/docker-compose` and then execute:
```sh
$ docker-compose -f docker-compose-with-async.yml up -d
Creating network "docker-compose_default" with the default driver
Creating microcks-zookeeper       ... done
Creating microcks-db              ... done
Creating microcks-sso             ... done
Creating microcks-postman-runtime ... done
Creating microcks                 ... done
Creating microcks-kafka           ... done
Creating microcks-async-minion    ... done
```

After few seconds, you should have something like:
```sh
$ docker ps
CONTAINER ID        IMAGE                                              COMMAND                  CREATED             STATUS              PORTS                                                                NAMES
6563591f360c        quay.io/microcks/microcks-async-minion:latest      "/deployments/run-ja…"   7 seconds ago       Up 5 seconds        8080/tcp                                                             microcks-async-minion
[...]

PR #458, opened 2021-07-27
#451 added kafdrop for visual monitoring of kafka topics

tillias:


lbroudoux (changes requested):
Thanks Ilia for the PR! 

I would prefer to set the new block into a dedicated file (something like `kafdrop-addon.yml` or something) to avoid consuming even more resources.

This way it allows people to choose between:
* minimalistic async config: `docker-compose -f docker-compose.yml -f docker-compose-async-addon.yml up -d`
* async config + visualization: `docker-compose -f docker-compose.yml -f docker-compose-async-addon.yml -f kafdrop-addon.yml up -d`

Would you please move it?

tillias:
Great idea, done. I'm not sure what is happening with travis, it hangs after login into https://hub.docker.com

lbroudoux (approved):
Thanks a lot!

Issue #481, opened 2021-09-06
Allow resources values overrides for Keycloak and MongoDB


john873950:
We ran into issues with our deployment as our cluster default resources were too low. deployment.yaml hard-codes some resources values and doesn't pick any values up from values.yaml. It would be great if these could be defaulted in the values file so that they can easily be overridden.

john873950:
If this is okay as an issue, I should have a PR for it ready soon.

lbroudoux:
It's definitely OK as an issue and as the PR is very welcomed 😉 
Also, it would be very nice to explain how the default resources were too low for your cluster conditions. Was it a startup delay issues that makes all the probe fail ? Or poor performance ? How do you find it was related to resource values ? Everything relevant would be super interesting to help future similar troubleshootings. Thanks !

john873950:
I think that microcks didn't have defaults for most things and therefore would be heavily affected by the cluster settings. I believe our cluster itself has low defaults to stop costs spiralling.
For example - for mongodb you currently specify a memory limit in deployment.yaml: https://github.com/microcks/microcks/blob/master/install/kubernetes/microcks/templates/deployment.yaml#L255 but nothing else is specified so I assume it'll fall back to defaults set by your cluster admin? And because it is hard coded in the deployment.yaml there is no way of overriding it.
I've been deploying fine with these settings for microcks, keycloak and mongo:
    requests:
      cpu: 250m
      memory: 400Mi
    limits:
      cpu: 500m
      memory: 512Mi

john873950:
Pull request: https://github.com/microcks/microcks/pull/486

lbroudoux:
Thanks a lot for this PR. Keeping this issue open for now to track reporting of this within the Operator.

PR #486, opened 2021-09-09
Allow resource overrides

john873950:
https://github.com/microcks/microcks/issues/481


lbroudoux (approved):
Thanks a lot for this nice PR!

Issue #507, opened 2021-10-28
Allow setting of annotations on pvcs from the values file


john873950:
We need to be able to annotate our keycloak pvc with a resource-policy of keep. This is so we don't loose all our keycloak configuration between reinstalls.

It would be great if we can use the values.yaml to set pvc annotations and pick those up in claims.yaml.

lbroudoux:
Thanks @john873950 for raising this one. Considering it for next release.

john873950:
I think this PR should allow it: https://github.com/microcks/microcks/pull/514

lbroudoux:
Now merged ;-)

PR #514, opened 2021-11-25
Allow claims annotations to be set in values file

john873950:
See https://github.com/microcks/microcks/issues/507

### DIFF
--- a/commons/model/pom.xml
+++ b/commons/model/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <artifactId>microcks</artifactId>
     <groupId>io.github.microcks</groupId>
-    <version>1.4.2-SNAPSHOT</version>
+    <version>1.4.1-fix-2</version>
     <relativePath>../../pom.xml</relativePath>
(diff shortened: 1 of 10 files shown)


lbroudoux:
Thanks a lot for this contribution, it's awesome!

I've got 2 questions:
* Shouldn't the `helm.sh/resource-policy` be defined as default? What is the best practice?
* Could you please re-target this PR to the `1.5.x` branch? This is where current dev are going.

lbroudoux (approved):
Great! Thanks.

john873950:
Regarding defining a default, I tried to make the out-of-the-box behaviour the same as before so as to not confuse other users. The default behaviour is that when running helm uninstall it will delete the pvc (including all your keycloak configuration). I think that should be the default (otherwise you get stuff left lying around when you've uninstalled). There's an argument that it's annoying losing your keycloak config but then you can override the resource policy to fix that.

But it's up to you :)

lbroudoux:
Agreed!

PR #566, opened 2022-02-25
Fix podman-compose support

nmasse-itix:
This PR is enabling support for podman-compose 1.x.

Since podman-compose v1.0, the --transform_policy option is not accepted anymore and at the same time, not needed anymore.


nmasse-itix:
I also updated keycloak that was outdated according to the adapter version used in Microcks.

lbroudoux (approved):
Thanks a lot for this fix!

Issue #614, opened 2022-06-10
Response of wrong operation returned when no exact dispatching


imod:
we have noticed the following strange behaviour, which fully blocks us now:
Given two operations:

`GET /pastry/{name}`
`GET /pastry/{name}/details`

both have there own examples. If I now call `GET /pastry/mycaffee/details` - where `mycaffee` is the value for name, but don't have an example defined for this value - I will get the mock response from `GET /pastry/{name}`.
So it seems, that if there is no matching mock for a given value, it just ignores the path and drops back to the path higher in the hierarchy.

imod:
@lbroudoux many thanks, do you know if the microcks-operator accepts sha as image idenitfier - e.g. for this exact change it would be:
```
spec:
   version: sha256:bf561691961bf24a8e04fe27ba4c5fc396fee173b96d8b4f564614bc2eec9703
```

lbroudoux:
Unfortunately no. The `spec.version` is used as a tag to build the images from various components (main, minion and postman runtime)... So specifying a sole image sha256 cannot work.

At the moment I don't see any opportunity - nor free time for me - to release 1.5.3 sorry... However `nightly` is now plugged on `1.6.x` branch. So as long as we do not rollback the PR, the fix must not disappear from new `nightly` images.

Are you heading to production in coming days/weeks?

imod:
so far this is fine, no problem. Just wanted to understand.
But one thing coming up on our site soon is, that we will not be able to connect to the internet anymore - so we will need to provide all the images from an internal registry. This is a security measure, to prevent from supply chain attacks.

lbroudoux:
OK noticed! Maybe we could add a `spec.registry` in the CRD. Once specified it would be used in place of the default `quay.io`. What do you think?

PR #615, opened 2022-06-10
Fallback matcher must not match slashes

imod:
fixes https://github.com/microcks/microcks/issues/614

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/RestController.java
+++ b/webapp/src/main/java/io/github/microcks/web/RestController.java
@@ -143,7 +143,7 @@ public class RestController {
                if (operation.getResourcePaths() != null) {
                   // Produce a matching regexp removing {part} and :part from pattern.
                   String operationPattern = getURIPattern(operation.getName());
-                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "(.)+");
+                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "([^/])+");
                   operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "\\/(.+)");
                   if (resourcePath.matches(operationPattern)) {
                      rOperation = operation;
--- a/webapp/src/main/java/io/github/microcks/web/RestController.java
+++ b/webapp/src/main/java/io/github/microcks/web/RestController.java
@@ -144,7 +144,7 @@ public class RestController {
                   // Produce a matching regexp removing {part} and :part from pattern.
                   String operationPattern = getURIPattern(operation.getName());
                   operationPattern = operationPattern.replaceAll("\\{.+\\}", "([^/])+");
-                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "\\/(.+)");
+                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "([^/])+");
                   if (resourcePath.matches(operationPattern)) {
                      rOperation = operation;
                      break;
--- a/webapp/src/main/java/io/github/microcks/web/RestController.java
+++ b/webapp/src/main/java/io/github/microcks/web/RestController.java
@@ -144,7 +144,7 @@ public class RestController {
                   // Produce a matching regexp removing {part} and :part from pattern.
                   String operationPattern = getURIPattern(operation.getName());
                   operationPattern = operationPattern.replaceAll("\\{.+\\}", "([^/])+");
-                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "([^/])+");
+                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "\\/([^/]+)");
                   if (resourcePath.matches(operationPattern)) {
                      rOperation = operation;
                      break;
--- a/webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
+++ b/webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
@@ -101,7 +101,7 @@ public class RestControllerIT extends AbstractBaseIT {
       ResponseEntity<String> response = restTemplate.getForEntity("/rest/pastry-details/1.0.0/pastry/Millefeuille/details", String.class);
       assertEquals(200, response.getStatusCode().value());
       try {
-    	  JSONObject details = new JSONObject(response.getBody());
+          JSONObject details = new JSONObject(response.getBody());
           String description = details.getString("description");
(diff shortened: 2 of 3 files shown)


imod:
@lbroudoux I don't understand why the tests fail in this repo, but work on my fork for the exact same commit: https://github.com/imod/microcks/actions/runs/2476011578

lbroudoux:
Thanks for the PR! Seems like it fails on my machine too... Will try to investigate.

lbroudoux:
I think that failure is just due to your test being too deterministic ...
When no precise response is found, it defaults to first matching operation and mime-type. But 'first' is not precise in MongoDB terms. Depending on how the import was done and Mongo internals (or version, or arch), results may vary. The important thing is that you receive a response coming from the details operation with `details` in the description. I think the integration test should be simplified just to check that criterion.

webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
```diff
@@ -88,4 +88,46 @@ public void testSwaggerMocking() {
...
+         JSONAssert.assertEquals("{"
+         		+ "		\"name\":\"Millefeuille\","
+         		+ "		\"description\":\"Detail - Delicieux Millefeuille pas calorique du tout\","
+         		+ "		\"size\":\"L\","
+         		+ "		\"price\":4.4,"
+         		+ "		\"status\":\"available\","
+         		+ "		\"street\":\"freestreet 3\","
+         		+ "		\"city\":\"Paris\""
+         		+ "}",
+               response.getBody(), JSONCompareMode.LENIENT);
```
lbroudoux, line 128: I think this part should be simplified just to check that we have a response coming from the `details` operation. Something like `description.startsWith("Detail -")` should be enough.

webapp/src/main/java/io/github/microcks/web/RestController.java
```diff
@@ -143,7 +143,7 @@ public ResponseEntity<?> execute(
                if (operation.getResourcePaths() != null) {
                   // Produce a matching regexp removing {part} and :part from pattern.
                   String operationPattern = getURIPattern(operation.getName());
-                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "(.)+");
+                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "([^/])+");
```
lbroudoux, line 146: Shouldn't we have something similar on next line too?

lbroudoux (changes requested):
Need to make the test more robust by simplifying assertions.

imod:
hmm, not sure I follow...
I think there is only one example of `details` imported at all, therefore I also expect this one to be returned - but maybe I don't understand it completely

webapp/src/main/java/io/github/microcks/web/RestController.java
imod: yeah, thats probably true

lbroudoux:
Actually I see 2 examples for the `details` responses in the OpenAPI file you attached to integration test. One for `Millefeuille` and another for `Eclair Cafe`...

imod:
@lbroudoux I simplified the test and adjusted the pattern for the second case too. The test are fine on my branch again, but fail on this one with an error which is related to a problem of the infrastructure. Maybe you can trigger the action again?
https://github.com/microcks/microcks/actions/runs/2487395986

imod:
@lbroudoux I resolved all the open points

webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
```diff
@@ -88,4 +91,32 @@ public void testSwaggerMocking() {
...
+   @Test
+   public void testFallbackMatchingWithRegex() {
+      // Upload modified pastry spec
+      uploadArtifactFile("target/test-classes/io/github/microcks/util/openapi/pastry-with-details-openapi.yaml", true);
+
+      // Check operation with a defined mock (name: 'Millefeuille')
+      ResponseEntity<String> response = restTemplate.getForEntity("/rest/pastry-details/1.0.0/pastry/Millefeuille/details", String.class);
+      assertEquals(200, response.getStatusCode().value());
+      try {
+    	  JSONObject details = new JSONObject(response.getBody());
```
lbroudoux, line 104: LGTM but code style has extra indent on 2nd and 3rd lines

webapp/src/main/java/io/github/microcks/web/RestController.java
```diff
@@ -143,8 +143,8 @@ public ResponseEntity<?> execute(
                if (operation.getResourcePaths() != null) {
                   // Produce a matching regexp removing {part} and :part from pattern.
                   String operationPattern = getURIPattern(operation.getName());
-                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "(.)+");
-                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "\\/(.+)");
+                  operationPattern = operationPattern.replaceAll("\\{.+\\}", "([^/])+");
+                  operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "([^/])+");
```
lbroudoux, line 147: I think this one is not OK as it removes first /.
`operationPattern = operationPattern.replaceAll("(/:[^:^/]+)", "\\/([^/]+)");` is the way to go. COuld you please modify this one ?

lbroudoux (changes requested):
Need a fix on postman notation and remove extra indents and we're good to merge!

webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
```diff
@@ -88,4 +91,32 @@ public void testSwaggerMocking() {
...
+
+      // Check operation with a defined mock (name: 'Millefeuille')
+      ResponseEntity<String> response = restTemplate.getForEntity("/rest/pastry-details/1.0.0/pastry/Millefeuille/details", String.class);
+      assertEquals(200, response.getStatusCode().value());
+      try {
+         JSONObject details = new JSONObject(response.getBody());
+         String description = details.getString("description");
+         assertTrue(description.startsWith("Detail -"));   
+      } catch (Exception e) {
+    	 fail("No Exception should be thrown here");
```
imod, line 108: I don't know whats happening with these indents - copy paste just does not work on my site
Do you have a code formatter rule or anything I can use to apply?

imod:
sorry, this took a lot more commits then it should - some how the formatting was not displayed the same way on GH as in my IDE...

lbroudoux (approved):
Thanks!

PR #616, opened 2022-06-17
Added README note for installing Microcks on arm64/M1 machine

HassanAzam:
image: jboss/keycloak:14.0.0 does not work when installing locally on arm64/M1 machine. Changing it to image which supports arm64 works fine. (image: wizzn/keycloak:14 supports arm64/M1 chips)


lbroudoux (approved):
Thanks for this note!

Issue #619, opened 2022-06-24
Dispatcher URI_PARTS does not ignore url-parts not mentioned in rule


ChristianLangmann:
When having an API with multiple URL-Path-parameters and a dispatchers  not including all of those parameters, the missing ones are not ignored when selecting the response.

For example a path .../{a}/{b} and a dispatcher URL_PARTS with rule "a" and provided examples for parameter a does not find the correct response ignoring parameter b. This actually works when using query-parameters (URL_PARAMS).

Attached is a simple openAPI-Spec describing this behaviour.
[mock-test - issue.yaml.zip](https://github.com/microcks/microcks/files/8977656/mock-test.-.issue.yaml.zip)

> curl http://localhost:8080/rest/mock-test/1.0/base/a1/undefined 
should return {"res":"A-1"}
> curl http://localhost:8080/rest/mock-test/1.0/base/a2/undefined 
should return {"res":"A-2"}

but they return the same result.

The log shows:
i.github.microcks.web.RestController     : Request body: null
i.github.microcks.web.RestController     : Found resourcePath: /base/a2/undefined
i.github.microcks.web.RestController     : Found a valid operation GET /base/{a}/{b} with rules: a
i.github.microcks.web.RestController     : Dispatch criteria for finding response is /a=a2/b=undefined
i.github.microcks.web.RestController     : No responses found so far, tempting with just bare operationId...

lbroudoux:
Hi @ChristianLangmann, thanks for creating this issue. Could you please specify the version of Microcks you're using ? I think we made some changes in `1.6.x` so that I'm not should it would event match the correct operation...

ChristianLangmann:
Hi Laurent,

I’m using Microcks 1.5.3. 

Best regards
Christian
---
Christian Langmann | Solution Architect

codecentric AG | Große Elbstraße 14 | 22767 Hamburg | Deutschland
mobil: +49 (0) 173. 6447742
www.codecentric.de | blog.codecentric.de | www.meettheexperts.de

Sitz der Gesellschaft: Solingen | HRB 25917| Amtsgericht Wuppertal
Vorstand: Rainer Vehns (Vorsitzender) . Ulrich Kühn . Lars Rückemann
Aufsichtsrat: Patric Fedlmeier (Vorsitzender) . Klaus Jäger . Jürgen Schütz . Michael Hochgürtel

[...]

PR #625, opened 2022-07-12
Issue619 url parts use dispatch rules

ChristianLangmann:
Idea of fix of issue 619 is to compare the parts of the request params with the param-rules and include the into the dispatch-criteria only if they are part of the rules. So params not part of the rules are wildcards, comparable to the query-params implementation.


lbroudoux:
Fix for #619

Issue #528, opened 2021-12-15
Microcks Ui doesnt show APIs & Services


jennerEVRFLL:
Hello,

It seems I have a problem with my k8s v1.20.11 installation of microcks.  I´ve  deployed microcks on two company clusters and local minikube instance. I have tested it with different versions locally and in the cluster. The minikube instance works fine but the cluster installations have a strange behaviour.

I can create APIs:
![image](https://user-images.githubusercontent.com/91196017/146151831-b39828ea-102b-4a09-9ffa-f9fa0d5ae03b.png)
But only get a blank page when try to list them.
![image](https://user-images.githubusercontent.com/91196017/146151587-ad3e8fc7-b34e-4d0e-8fe5-cc35d36d1687.png)

The pods do not return any error messages except this warning:
<details>
<summary>Podlog</summary> 

```
08:51:54.822 [DEBUG] io.github.microcks.web.InvocationController - Getting invocations stats for last 20 days
08:51:54.822 [DEBUG] io.github.microcks.web.InvocationController - Getting top 20 invocations stats for day 20211215
08:51:54.822 [WARN] org.keycloak.adapters.RequestAuthenticator - SSL is required to authenticate. Remote address 100.x.x.34 is secure: false, SSL required for: EXTERNAL .
```

</details>

The network analysis of Chrome and Firefox shows the following errors:

![image](https://user-images.githubusercontent.com/91196017/146156003-7b6ff329-c0a2-4a7d-8098-3c51915808a1.png)

![image](https://user-images.githubusercontent.com/91196017/146156018-31797827-8792-4ab0-b221-90967da88676.png)

jennerEVRFLL:
Hi @lbroudoux 

thanks for the reply. Yes we use a external Keycloak (Keycloak Operator) and a https endpoint because we also use a apicurio instance.

Here are some information about the configuration.

<details>
<summary>Helm Values</summary> 

```
keycloak:
  install: false
  realm: apicurio
  url: apicurio.xxx.xxx.one 
  privateUrl: https://apicurio.xxx.xxx.one/auth
```
</details>

 Configured Keycloak Clients: 

![image](https://user-images.githubusercontent.com/91196017/146205344-7712e2d6-b89a-43ae-9dad-47208ee310c2.png)

Mocking feature works but also displays a blank page on mickrocks:
![image](https://user-images.githubusercontent.com/91196017/146205686-24ab1375-ce75-4185-a3c1-062aa541f8c9.png)

[...]

PR #671, opened 2022-09-27
Add extraProperties to values and configmap so file with extraProperties can be specified in helm install.

carolgschwend:


lbroudoux:
This provides a way to fix #528. Thanks a lot @carolgschwend 🙏

lbroudoux (approved):
LGTM! Will just add a condition to switch profile from `prod` to `prod,extra` based on the `extra.enabled` property.

PR #678, opened 2022-10-27
WIP: feat: add conditional to create ingresses

marsou77:
I don't need to expose the service outside. I make Ingresses optional but the default value still create them.


marsou77:
Cancelled because I wasn't on the right branch when starting this PR

PR #679, opened 2022-10-27
Allow to disable ingress creation

marsou77:
I don't need an ingress to run Microcks since I use Istio on my cluster. To use this Helm chart, I need this change.


marsou77:
Hello @lbroudoux, when can I expect this to be merged ? Thank you :)

lbroudoux:
Hi there! Yes this is of great interest: thanks a lot! I currently have some days off but will have a look at it next week.

install/kubernetes/microcks/values.yaml
```diff
@@ -1,4 +1,5 @@
 appName: microcks
+ingress: true
```
lbroudoux, line 2: I think it would be better to name it `ingresses` as it controls the generation of all the `Ingress` objects.

lbroudoux (changes requested):
Thnaks!
Looks good to me apart the naming of param that'd better be `ingresses` IMO. Also could you complete the 2 `README.md` files (basically they're the same) to include documentation on this setting?

marsou77:
> Thnaks! Looks good to me apart the naming of param that'd better be `ingresses` IMO. Also could you complete the 2 `README.md` files (basically they're the same) to include documentation on this setting?

Done ! Agree with Ingresses :)
For the documentation I added `appName` which was missing. Not sure about the name I put on `section`, you tell me if it sounds right

lbroudoux:
That was super fast! 😊
Time to sleep for me now but will finish it tomorrow. Thanks again,

lbroudoux (approved):
LGTM 👍

Issue #722, opened 2022-12-14
Include RANDOM dispatch type for SoapUI project


sfraigneau:
Hello,

Is it possible to support the RANDOM dispatch type for the SoapUI projects ?

Thanks

lbroudoux:
Hey @sfraigneau 
Happy new year! I've merged this one 2 weeks ago. Did you have the ability to check if everything works as expected on `1.7.x`branch? The corresponding container images tag is `next`.

sfraigneau:
Hi Laurent,

Bonne année :)

I was on vacation last week and haven't had no time to check since then. Sorry.

I'll give you my feedback this week-end for sure

sfraigneau:
Hi Laurent,

When I start microcks:next, an error occurs : 
`Fatal glibc error: CPU does not support x86-64-v2`

I run it through a vm from a windows station and this issue could be related to the activation of 'Microsoft Hypervisor Platform' on my machine. Unfortunately, this is a computer from my organization and I'm not allowed to administrate it.

I'll try to find a solution next week

lbroudoux:
Looks like we have a platform architecture mismatch for your machine... Does it happen only with new `next` container image tag or also with previous `latest` tag?

lbroudoux:
Working on it with #734

sfraigneau:
Hi Laurent,

I deleted my previous `latest `docker image and pulling it again. The application starts as usual, no glibc error.

lbroudoux:
Hum ... may also be some difference related to Java 11 to Java 17 upgrade we have in `1.7.x` branch (`next` tag). Could you try pulling again the `next` tag as I have started pushing `arm64` images ?

sfraigneau:
Same error with the `next` tag

I'm using virtual box 6.1 and I saw that version 7 has an option to use host CPU directly. 
I'll give it a try

sfraigneau:
Hi @lbroudoux 

With virtual box 7, I didn't have any issue running the `next` tag. No need to change the VM settings nor to deactivate 'Microsoft Hypervisor Platform'. Just updating virtual box is enough.

By the way : 
I have tested a SoapUI project using a context in the dispatcher : work like a charm
I also tested the random dispatcher : work like a charm

This is perfect. Thank you.

lbroudoux:
Awesome thanks for your feedback and thanks for great contribution! Looking forward to hearing from you on all the cool stuff you're doing with Microcks 😉

PR #724, opened 2022-12-15
feat #722: [SOAP] include random dispatch for SoapUI project

sfraigneau:
My first one.

If this is ok, I'll do the same for the REST protocol


lbroudoux (changes requested):
Loog very good to me! Thanks for this PR!
One minor thing to change though is the indentation on Java files (looks like you've used 2 spaces instead of 3). It would be awesome to fix that.

sfraigneau:
Hi Laurent,

I rebase and change the indentation as requested

lbroudoux (approved):
Thanks a lot!

PR #780, opened 2023-04-06
Code refactoring

CapnSK:
### Description

- Extracted complex conditions into a method for increased readability: DispatchCriteriaHelper.java:374
- Created a method out of an operations finding task for reusability  & improved readability: RestController.java:132
- Renamed method as per semantics for better understanding its use: DocumentationController.java: 56
- Moved method computeMaxPossibleConformanceScore logic from MetricService to MessageService because of lack of cohesion
- Created SchemaValidator base abstract class with abstract validate method and OpenAPISchemaValidator and SwaggerSchemaValidator to extend the base class and implement validate method.
- Response, Request and EventMessage classes use id variable which can be kept in Message class for better maintainability
- DynamicMockRestController has certain utilities which have more to do with Json handling, extracted these methods into a nested class for better abstraction

### DIFF
--- a/commons/util/src/main/java/io/github/microcks/util/openapi/SchemaValidator.java
+++ b/commons/util/src/main/java/io/github/microcks/util/openapi/SchemaValidator.java
@@ -1,0 +1,9 @@
+package io.github.microcks.util.openapi;
+
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public abstract class SchemaValidator {
+    public abstract List<String> validate(JsonNode schemaNode, JsonNode jsonNode);
+}
--- a/webapp/src/main/java/io/github/microcks/service/MessageService.java
+++ b/webapp/src/main/java/io/github/microcks/service/MessageService.java
@@ -22,6 +22,8 @@ import io.github.microcks.domain.*;
 import io.github.microcks.repository.EventMessageRepository;
 import io.github.microcks.repository.RequestRepository;
 import io.github.microcks.repository.ResponseRepository;
+import io.github.microcks.util.IdBuilder;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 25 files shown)


lbroudoux:
Thanks a lot for these refactoring 👍 
Will rebase it on the `1.7.x` branch as this is where next developments are. 
It will probably take a bit of time to review everything 😉 
Do not hesitate to join GH discussions or zulip to discuss refactoring beforehand so that we can also agree on a strategy to split them and accelerate review and integration.
Thanks again!

CapnSK:
Apologies for not discussing this beforehand; thanks for considering merging this pull request.

commons/util/src/main/java/io/github/microcks/util/openapi/SchemaValidator.java
lbroudoux, line 1: It could be nice to discuss the purpose of introducing this interface... Would it be possible to use it again for some other specs like AsyncAPI and GraphQL for example? Does it ease schema validator client usage as well? I don't see the point at the moment and I'd like to see further before applying the change.

webapp/src/main/java/io/github/microcks/service/MessageService.java
```diff
@@ -49,6 +51,19 @@ public class MessageService {
    private EventMessageRepository eventMessageRepository;
 
 
+   public List<? extends Exchange> computeMaxPossibleConformanceScore(Service service, Operation operation){
```
lbroudoux, line 54: I think the name of this method is not appropriate here... Something like `getExchangesByOperation()` would be more appropriate.

webapp/src/main/java/io/github/microcks/service/MetricsService.java
```diff
@@ -185,16 +186,8 @@ private double computeMaxPossibleConformanceScore(Service service) {
...
-         if (ServiceType.EVENT.equals(service.getType()) || ServiceType.GENERIC_EVENT.equals(service.getType())) {
-            // If an event, we should explicitly retrieve event messages.
-            exchanges = messageService.getEventByOperation(
-                  IdBuilder.buildOperationId(service, operation));
-         } else {
-            // Otherwise we have traditional request / response pairs.
-            exchanges = messageService.getRequestResponseByOperation(
-                  IdBuilder.buildOperationId(service, operation));
-         }
+         List<? extends Exchange> exchanges = messageService.computeMaxPossibleConformanceScore(service, operation);
```
lbroudoux, line 189: See comment on MessageService extracted method name...

lbroudoux (changes requested):
Please let me know what you think of above comments.

lbroudoux:
HI @CapnSK! Would you still want to pursue on this one? I think some renaming is required if I remember correctly but nothing that looks too complicated. Let me know.

lbroudoux:
Shifting this one to next version as we'll not be able to handle it for 1.7.1

lbroudoux:
Hi @CapnSK!
Are you still interested in contributing this one? If you could have a look at it this week, I could integrate those into the next `1.8.0` release. Otherwise, we'll have to shift this one...
Many thanks!

lbroudoux:
It's been too long since this PR receive attention. Also, most of the code refactoring has been done apart from successive smallest  issues. Closing it as undone.

PR #786, opened 2023-04-17
chore: use import realm startup argument instead of using a custom script to import realm

skirankumars31:
### Description

- used import-realm that can import the json realm file automatically from the /opt/keycloak/data/import folder while starting up
- The earlier solution also worked but the new one is straightforward

### DIFF
--- a/install/docker-compose/docker-compose.yml
+++ b/install/docker-compose/docker-compose.yml
@@ -9,20 +9,19 @@ services:
 
   keycloak:
     #image: jboss/keycloak:14.0.0
-    image: quay.io/microcks/microcks-keycloak:20.0.2
+    image: quay.io/keycloak/keycloak:21.0.2
     container_name: microcks-sso
     ports:
       - "18080:8080"
     environment:
       KEYCLOAK_ADMIN: "admin"
       KEYCLOAK_ADMIN_PASSWORD: "admin"
-      KEYCLOAK_IMPORT: "/tmp/microcks-realm.json"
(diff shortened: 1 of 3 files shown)


lbroudoux (approved):
Awesome! This was on my todo list!

Some months ago, the latest version of Keycloak did not allow importing realms def on startup and that's why I came with `microcks-keycloak` custom image. I realize this feature was present in latest Keycloak just before releasing `1.7.0` so did the update in Helm chart and operator but forgot docker-compose.

If you're interested in some other contribution, Keycloak is also used and references in some other docker-compose files in the same repo 😉

install/docker-compose/docker-compose.yml
```diff
@@ -9,20 +9,19 @@ services:
 
   keycloak:
     #image: jboss/keycloak:14.0.0
-    image: quay.io/microcks/microcks-keycloak:20.0.2
+    image: quay.io/keycloak/keycloak:21.0.2
```
lbroudoux, line 12: I didn't see it changed at first... It would be better to stick to 20.0.2 so that we'll use the same version everywhere.
Upgrading to `21.0.2` would be another topic and require more work.

lbroudoux (changes requested):
Can you please stick with `20.0.2` version?

install/docker-compose/docker-compose.yml
skirankumars31: I have now changed the version back to 20.0.2. Made changes to the other docker compose files as well

lbroudoux (approved):
LGTM. Awesome! Thanks for the PR.

PR #880, opened 2023-07-26
fix: OpenAPI v1.7 schema

JulienBreux:
### Description

Some fields of OpenAPI v1.7 schema are wrong

Fixes:
- Additional properties of `CounterMap` (it's a `number`)
- `public-client` of `KeycloakConfig` component returns a boolean, not a string


lbroudoux (approved):
LGTM! Thanks!

Issue #884, opened 2023-08-01
Replace SoapUI jar dependencies with our own implementation


lbroudoux:
### Reason/Context

SoapUI jar and dependencies have been embedded into Microcks since day one as the support of Soap WebServices was (and is still) on the list of features. At that time, directly embedding SoapUI jars was a nice idea and was a way of getting faster to our goal. However, today, SoapUI brings a lot of challenges:

* Latest version `5.7.0` is 2 years old and cannot be integrated into Microcks because of conflicting libraries,
* Project is in maintenance mode and quite abandoned in favor or new priorities for SmartBear teams,
* The number of CVEs in dependency is huge,
* The number and volume of dependency - as well as their diversity licences to manage - is huge,
* The runtime behavior of SoapUI is not very pleasant - slowed by UI considerations that should not bother us in headless mode.

That's why we decide to start this initiative of replacing SoapUI implementations with our own. The goal is not to get rid of/discontinue support of SoapUI but rather to get even better control of it.

### Description

We are then considering replacing SoapUI jar and libraries with our own implementation as our use-cases are well-known and very well scoped. We are using SoapUI java APIs in 2 cases today:
* During the import of a SoapUI project to transform SoapUI model elements into Microcks one. The initial decision of using it to avoid Xml parsing "at hand" can be revised and this will just require a bunch of pure Xml parsing,
* During the test of an endpoint when running the `SOAPUI` strategy. This is actually using SoapUI classes to execute the different requests and attached assertions. This can also be easily replaced by our own `HttpTestRunner` base class but we have to port some SoapUI assertions into our own code-base.

For 1st use-case, we have a pretty descent unit tests coverage and it's just the matter of switching implementation while ensuring everything still running as is with the test suite.

The 2nd use-case is a bit trickier as we don't have an exhaustive view of the SoapUI test assertions that are actually used on the field... We propose starting with the implementation of the list below:
* Valid HTTP Status Codes
* SOAP Response
* SOAP Fault Assertion
* Simple Contains
* Not Contains
* Schema Compliance
* XPath Match
* JsonPath Match
* JsonPath Existence Match
* SLA

Please do not hesitate to comment and add new requirements regarding assertions!

cc @sfraigneau as I know you're a heavy user of this. The new implementation will stay in the repo next to the existing one in order to be able to easily switch them. I encourage you to test as early as possible and giving feedback!

lbroudoux:
[...]

PR #891, opened 2023-08-22
Handle namespace in XPATH as soapui xmlHolder does

sfraigneau:
### Description

I added three tests to query the XmlHolder class. They pass when using the SOAPUI implementation.
I fixed the microcks implementation to handle that cases.

### Related issue(s)

 #884


lbroudoux (approved):
Awesome! LGTM!

lbroudoux:
Hey @sfraigneau! That's awesome! Thanks a lot! 

Could you please just sign your commit to pass the DCO check above? This is a Best Practice we have now enforced since we joined the CNCF. 

As you just have one commit, it would just be a matter of running the 2 following commands on your local branch:

```
git rebase HEAD~1 --signoff
git push --force-with-lease origin 1.8.x
```

Thanks again!

sfraigneau:
Done,

PR #912, opened 2023-09-08
Add ingressClassName property

JParisR:
### Description
This PR adds the 'ingressClassName' field to the spec property of the ingress yaml. It will allow to specify a custom ingress directly in the property rather than using the annotation.

The chart was tested with and without the ingressClassesName and it works as expected.

image
*Example shows microcks-grpc ingress without ingressClassName set and microcks-keycloak with ingressClassName added*


lbroudoux (approved):
LGTM!

lbroudoux:
Thanks a lot for the contribution!

PR #924, opened 2023-10-03
Replaced Javafaker by Datafaker

bodiam:
Hi there! Thanks for creating this project. I'm planning on using it to test our APIs, and I saw you're using Javafaker, a library which is no longer maintained. We now maintain a fork of Javafaker called Datafaker, and I was hoping you'd be open to use that. I changed the implementation (creating an instance of Datafaker is cheap since all data is lazily loaded, while Javafaker is very expensive). 

I've also addressed a few other things:

* removed commons.lang3, which is no longer maintained. 
* Cleaned up some code
* added Maven wrapper
* and fixed a few small bugs.


bodiam:
PS: I had a few ideas on making the random parts more predictable, but not sure if you're open to that, so I left most of this code untouched. Cheers, Erik

bodiam:
Hi @lbroudoux, I'll see what I can do.

PR #940, opened 2023-10-15
Replaced Javafaker by Datafaker.

bodiam:
### Description

- Replaces Javafaker (no longer maintained, couple of CVE's, etc) by Datafaker (actively maintained, no CVE's afaik)


webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -201,7 +202,7 @@ public ResponseEntity<?> execute(
          }
 
          if (!responses.isEmpty()) {
-            int idx = DispatchStyles.RANDOM.equals(dispatcher) ? RandomUtils.nextInt(0, responses.size()) : 0;
+            int idx = DispatchStyles.RANDOM.equals(dispatcher) ? new Faker().random().nextInt(0, responses.size()) : 0;
```
lbroudoux, line 205: This change doesn't look equivalent, as it makes a test fail (see https://github.com/microcks/microcks/actions/runs/6522873384/job/17732084315#step:6:300). I propose removing this change and addressing `commons.lang3` replacement in another issue.

lbroudoux: The max from Faker random is not exclusive like the one from RandomUtils.

bodiam: The reason why I removed it is because you never really depended on it. It was a transitive dependency introduced by Javafaker, and it has a few CVEs, so I copied the logic into a util class, though I think it would be better to get rid of the whole logic, cause I doubt it's really needed.

bodiam: Regarding the max, did you change it or do you want me to change it?

PR #947, opened 2023-10-22
Reduce some sonar cloud issue and add dev sh

robertonav20:
### Description
I've seen sonar cloud issue and i want to contribute to close them some

This is my first PR, so correct me if i'm wrong something


lbroudoux:
Hi Roberto!

Thanks a lot for this PR! Awesome work! Could you please sign your commits to fix the DCO check before going further?
You can review the [instructions here](https://github.com/microcks/microcks/pull/947/checks?check_run_id=17937089943), usually it's just a matter of a few Git commands.

I'll start reviewing the PR just after that.

> Note: also I've changed the target of your PR to the `1.8.x` branch, as this is where development is occurring for next `1.8.1` release. Not a big deal as it is in sync with the `master` at the moment but beware for next PRs 😉

robertonav20:
Hi @lbroudoux,
i think now is good for the merge

commons/util-el/src/main/java/io/github/microcks/util/el/function/RandomBooleanELFunction.java
```diff
@@ -25,8 +25,10 @@
  */
 public class RandomBooleanELFunction implements ELFunction {
 
+   private Random random = new Random();
```
lbroudoux, line 28: Just a side note: I think it would be even better to make this Random static or shared at an upper level like a `AbstractRandomELFunction`. This would allow us to circumvent this bug of not random generation mentioned by Sonar cloud but also to save more resources and share random objects between more instances.

lbroudoux (approved):
LGTM! Thanks a lot!

lbroudoux:
Now merged! Awesome!

commons/util-el/src/main/java/io/github/microcks/util/el/function/RandomBooleanELFunction.java
robertonav20: Yes of course I'm agree with you

robertonav20: If you want I can add this on a new pr

PR #960, opened 2023-10-27
refactor: Rename Constant in Enum

apoorva256:
Ref: https://sonarcloud.io/project/issues?open=AYmZth2KBCcdL_hhQMcl&id=microcks_microcks

### Description

- ...
- ...
- ...


lbroudoux:
Unfortunately, this will introduce a breaking change in the Microcks API as this enumeration is used as is. I think we'll have to postpone this refactoring for a v2.0

Issue #966, opened 2023-10-29
Add kafka login callback handler support


Afanas10101111:
### Reason/Context

Hi!
I want to add support for a custom Kafka login callback handler to the project.
This will allow to integrate Microcks with the Kafka broker, which uses authorization.

### Description

To do this, I need to add a few lines of code. There will be no breaking changes.
After the improvement all we need is to build microcks-async-minion as legacy-jar, add the property with fully qualified name of our custom handler class, add its library to the classpath.

### Implementation ideas

The idea is to add a property setting here:
https://github.com/microcks/microcks/blob/3088050a14e990cdc55df1215e1351bfcde3ad2d/minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java#L290

lbroudoux:
Is this related to the support of `OAUTHBEARER` sasl mechanism?

Afanas10101111:
In my case, yes - I needed to produce mock messages to kafka topic under OAUTHBEARER. As I know, there are few auth mechanisms supported by kafka (https://docs.confluent.io/platform/current/kafka/overview-authentication-methods.html), and I assume that sasl.login.callback.handler.class may be used for any of them. But, truely, I tested OAUTHBEARER only.

lbroudoux:
@ludovic-pourrat: I know you had the same request for feature enhancement and may have done some things on your side. Could you check this one, please?

lbroudoux:
Hey @Afanas10101111!
Sorry for the delay, but I was at KubeCon last week and had to manage priorities.
I'm still interested in this one and will probably have a look at it tomorrow to go forward on merging. One thing I was wondering is the need to change Quarkus packaging to `legacy-jar`: don't we need to apply changes in the `pom.xml` of the module too?

Afanas10101111:
No, no additional changes are required, all we need to do is run `mvn package -Dquarkus.package.type=legacy-jar`. Alternatively we can add `quarkus.package.type` as a property in `minions/async/src/main/resources/application.properties`, but I think it is more suitable for build time, so I added it as part of the instruction in `minions/async/src/main/docker/Dockerfile.legacy-jar`

Afanas10101111:
Hey @lbroudoux 
I want to ask for advice. I want to add basic `OAUTHBEARER` support for the consumer too. Is it correct to make a new pull request and link it to this issue?

lbroudoux:
[...]

Issue #897, opened 2023-08-25
Accept external lib in groovy script


sfraigneau:
### Reason/Context

We want to use external lib in our groovy script

### Description

We coulp provide to microcks a path where our libs belong (jar files)

lbroudoux:
Done some tests on this, and it's already possible 😉 Just a matter of writing a nice documentation or blog post...

To give you quickly the raw directions: there are 2 approaches depending if you use docker-compose or pure Kubernetes installation.

#### docker-compose method

1. Put your Jar files into a `./lib` folder
2. Add the following lines into your compose file for the Microcks container:
```yaml
   volumes:
      - ./lib:/deployments/lib
    environment:
      - JAVA_OPTIONS=-Dloader.path=/deployments/lib
      - JAVA_MAIN_CLASS=org.springframework.boot.loader.PropertiesLauncher
      - JAVA_APP_JAR=app.jar
```
3. Restart and see the Jar files appended to the application classpath.
4. You can directly use the Java or Groovy classes from your Jar in a `SCRIPT`

#### Pure Kubernetes method

1. Put your Jar files into a local `./lib` folder
2. Create a new Docker file that extends the Microcks provided image and embeds your libs
```yaml
FROM quay.io/microcks/microcks-uber:nightly

# Copy libraries jar files
COPY lib /deployments/lib

ENV JAVA_OPTIONS=-Dloader.path=/deployments/lib
ENV JAVA_MAIN_CLASS=org.springframework.boot.loader.PropertiesLauncher
ENV JAVA_APP_JAR=app.jar
```
3. Restart and see the Jar files appended to the application classpath.
4. You can directly use the Java or Groovy classes from your Jar in a `SCRIPT`

Let me know if it helps!

sfraigneau:
Hi Laurent,

This is exactly what I need. I also do perform tests on my side and it works like a charm !

Thanks :)

lbroudoux:
Awesome! Thanks for the feedback Sébastien.
As you reviewed and replayed the process, would you like to help with the official documentation write-up?

sfraigneau:
Yes, of course.

lbroudoux:
What's your thought on most suitable form for documentation?
* A formal reference documentation that would live in https://github.com/microcks/microcks.io/tree/master/content/documentation/using/advanced?
* Or a more practical guide that would live in https://github.com/microcks/microcks.io/tree/master/content/documentation/guides?

[...]

PR #967, opened 2023-10-29
feat: kafka login callback handler support

Afanas10101111:
### Description

- Add kafka sasl.login.callback.handler.class property support
- Add Dockerfile for quarkus legacy-jar build

### Related issue(s)

- <a href="https://github.com/microcks/microcks/issues/966">#966</a>

### DIFF
--- a/minions/async/src/main/docker/Dockerfile.legacy-jar
+++ b/minions/async/src/main/docker/Dockerfile.legacy-jar
@@ -1,0 +1,48 @@
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
+#
+# Before building the docker image run:
+#
+# mvn package -Dquarkus.package.type=legacy-jar
+#
+# Then, build the image with:
+#
+# docker build -f src/main/docker/Dockerfile.legacy-jar -t microcks/microcks-async-minion-legacy-jar .
+#
+# Then run the container using:
+#
+# docker run -i --rm -p 8080:8080 microcks/microcks-async-minion-legacy-jar
+#
+###
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2-691
+
+ARG JAVA_PACKAGE=java-17-openjdk-headless
+ARG RUN_JAVA_VERSION=1.3.8
+
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en'
+
+# Install java and the run-java script
+# Also set up permissions for user `1001`
+RUN microdnf install curl-minimal ca-certificates ${JAVA_PACKAGE} -y \
+    && microdnf update -y \
+    && microdnf clean all \
+    && mkdir /deployments \
+    && chown 1001 /deployments \
+    && chmod "g+rwX" /deployments \
+    && chown 1001:root /deployments \
+    && curl https://repo1.maven.org/maven2/io/fabric8/run-java-sh/${RUN_JAVA_VERSION}/run-java-sh-${RUN_JAVA_VERSION}-sh.sh -o /deployments/run-java.sh \
+    && chown 1001 /deployments/run-java.sh \
+    && chmod 550 /deployments/run-java.sh \
+    && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
+
+# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
+ENV JAVA_MAIN_CLASS="io.quarkus.runner.GeneratedMain"
+ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
+
+COPY target/lib/ /deployments/lib/
+COPY target/*-runner.jar /deployments/
+
+EXPOSE 8080
+USER 1001
+
+ENTRYPOINT [ "/deployments/run-java.sh" ]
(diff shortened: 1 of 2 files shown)


lbroudoux:
Thanks for the PR! I am having a few days off at the moment but will probably have a look at it on Thursday. Cheers,

lbroudoux:
Hi @Afanas10101111 

Is the `legacy-jar` packaging type required to load your customized callback handler implementation class? Am I correct? Is it the only way of doing this stuff for Quarkus app? I know there are different types of packaging and a re-augmentation process available, but I am not fully aware of the differences and pros vs cons.... Are there any helpful resources to better understand this?

We have the same concern that is enabling the loading of custom classes for the main component/container (see https://github.com/microcks/microcks/issues/897) and are in the process of describing the complete way of doing things. Could you elaborate on how you'll also project to use this for the minion component? That way we may try to make things consistent for both components.

Thanks,

Afanas10101111:
Yes, You are correct! The standard package type for Quarkus apps is fast-jar and it indexes the contents of all dependencies at build time to provide a quick start. But I want to avoid rebuild each time I change handler implementation. To do this I need an app to use standard classloader, which in Quarkus apps available for legacy-jar package type (https://quarkus.io/guides/maven-tooling#fast-jar).

To use my custom handler class after building as legacy-jar I do next steps:

- add property kafka.sasl.login-callback-handler-class=my.custom.kafka.oauth.OAuthCallbackHandler

- add env property JAVA_CLASSPATH=/deployments/*:/deployments/custom/* (to use logic of /deployments/run-java.sh)

- mount my custom library to /deployments/custom/

- restart minion

- look at our messages in kafka :)

Afanas10101111:
I have added the path to the app dir in `Dockerfile.legacy-jar`. Now instead of adding `JAVA_CLASSPATH` env with the path to the app dir and the custom lib dir, we need to add only `JAVA_LIB_DIR=/path/to/custom/lib/*`. And now without adding `kafka.sasl.login.callback.handler.class` and `JAVA_LIB_DIR` the minion runs as it was. I think this is a more clear way to configure this feature.

lbroudoux:
Hey @Afanas10101111,
I got it working and checked that I can now load custom classes using the abovementioned mechanism! That's great!
However, I think it still needs extra configuration to get it fully working to support OAuth bearer. I'd like to be able to have a fully working example and grasp what other configuration variables (`kafka.sasl.login.callback.handler.class` but maybe also others...) we'd need to add as configuration properties in Microcks Helm Chart or Operator. Would you have a guide or procedure on how to test that with a Keycloak for example?

Thanks,

Afanas10101111:
I assume that all extra configuration should be on the custom classes side. The idea is that when you use Kafka with oauth, you provide a client library, and its implementation may be different.

For example, <a href="https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java">here is one of the possible implementations</a>. To use it we, need
- add in the minion `application.properties`
```
# Already existing fields
kafka.security.protocol=SASL_SSL
kafka.ssl.truststore.location=/deployments/config/truststore.jks
kafka.ssl.truststore.password=truststore_secret
kafka.sasl.mechanism=OAUTHBEARER
kafka.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required;

# New field
kafka.sasl.login-callback-handler-class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
```
- set env `JAVA_LIB_DIR=/path/to/strimzi/kafka-oauth-client/*:/path/to/strimzi/kafka-oauth-client/lib/*`
- mount directories with kafka-oauth-client.jar and its libs
- configure this handler according to <a href="https://github.com/strimzi/strimzi-kafka-oauth/tree/main#client-config-example">its documentation</a> by setting env `OAUTH_CLIENT_ID`, `OAUTH_CLIENT_SECRET` and `OAUTH_TOKEN_ENDPOINT_URI`

This last step may be completely different for different implementations. For example, the names of the parameters and their number may differ, or we may need to specify them in the properties file.

I hope this will help to understand my way of researching this feature and get a working sample :)

lbroudoux:
Thanks!

> This last step may be completely different for different implementations. For example, the names of the parameters and their number may differ, or we may need to specify them in the properties file.

This point was the thing I wanted to check. We need adding - thus - the ability to add arbitrary environment variables and properties to the async-minion configuration.

In the case of properties in the `application.properties` file, I imagine we had to also store them into the Kafka producer properties so that they can be retrieved and used at configuration time by the `AuthenticationCallbackHandler` in `public void configure(Map<String, ?> map, String saslMechanism, List<AppConfigurationEntry> jaasConfigEntries)`? Could you confirm this later point?

Afanas10101111:
I tested this feature with an implementation that depends on 'kafka-oauth.properties', and it was enough to add this file to the classpath (in the custom lib directory).
If the implementation directly depends on `application.properties`, it is enough to add the necessary properties to `application.properties` of the minion, I guess. I'll try to test such case tomorrow. But in this case, we have a risk of collision of property names, I'm afraid.

Afanas10101111:
I've checked - with the implementation depends on `application.properties` everything works as I expected.

About this question
> In the case of properties in the application.properties file, I imagine we had to also store them into the Kafka producer properties so that they can be retrieved and used at configuration time by the AuthenticationCallbackHandler in public void configure(Map<String, ?> map, String saslMechanism, List<AppConfigurationEntry> jaasConfigEntries)? Could you confirm this later point?

Not quite. You are right, this method works with the Kafka producer properties. But our additional custom handler configuration is required to connect to the oauth server and get a token and should be configurated separately (by a propertiy file or/and env variables)

lbroudoux:
Thanks again for this contribution!

lbroudoux:
@Afanas10101111 
I'm sorry, man, but I screwed this one and merged it into `master`... I had to revert this as the devs are going into branch `1.8.x` at the moment. Could you re-create your PR targeting this branch? I'd merge it super quickly now.
Sorry again.

PR #985, opened 2023-11-09
fix: add null check to getFileName

apoorva256:
### Description
When importing from gitlab the code fails when trying to get the filename as there is a null key in the Headers


lbroudoux (approved):
Thanks for the check and the PR!

PR #999, opened 2023-11-17
Skip test feature

### LINKED ISSUE #998: Allow to skip automated test with OpenAPI/AsyncAPI
### Reason/Context

Hello, while testing api with microcks I had some issue with some endpoints not returning a body therefore the test fails because there is no mapping between the request and the response as presented in this issue [#944](https://github.com/microcks/microcks/issues/944)

I know changing the way the test works by allowing empty response can take some time, and a easy temporary workaround would be to skip the test while such a feature is not there yet, and may be other feature that are not supported yet can make a skippable test a nice thing

### Description

Make a test skippable by adding some kind of property in the spec definition

This way test that are using Api feature that are not already handled by Microcks can pass without breaking any CD/CI pipeline that should assert that the tests are passing

I already did a POC and a PR should be pushed soon (I won't forget to sign off my commit this time I hope 😅) 

I also wanted to thank you for the fast feedback for the other issue/PR, agains thanks for all this work

### Implementation ideas

A nice thing could be to add a property in the openapi/asyncapi spec like
```yaml
x-microcks-skip-test: true
```
that would just skip the iteration of the test in the test runner while setting the success flag to true, and adding a new boolean flag like "skipped" 
and then display them as skipped in the frontend

Snorkell:
### Description

- Add a "skipped" flag to test results
- Add a new microcks property "x-microcks-skip-test: true" for asyn api and open api
- Let the skipped test marked as success to not make the test suite fails

### Related issue(s)
Resolves #998

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ To get involved with our community, please make sure you are familiar with the p
 
 ## Build Status
 
-The current development version is `1.8.0-SNAPSHOT`. [![GitHub Workflow Status](https://img.shields.io/github/actions/workflow/status/microcks/microcks/build-verify.yml?branch=1.8.x&logo=github&style=for-the-badge)](https://github.com/microcks/microcks/actions)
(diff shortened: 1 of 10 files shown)


lbroudoux:
I'd like to keep on going the discussion on this one. Do you need this in addition or operation selection ability? Do you think it's useful to have this marked at the spec level? Any comments are welcomed!

PR #1006, opened 2023-11-28
feat: kafka login callback handler support

Afanas10101111:
### Description

Add a simple SASL OAUTHBEARER support for Kafka consumers.

- Add a path variable that defines a custom handler for the Kafka consumer.
- Add a path variable that defines a consumer group.id property (there may be a restriction on the broker's side).

### Related issue(s)

- <a href="https://github.com/microcks/microcks/issues/966">#966</a>


Afanas10101111:
Hey @lbroudoux !
I want to disscuss this idea with you.
This is the simplest way to support OAUTHBEARER for consumers.
On the plus side
- we can use different handles on the fly.

On the minus side
- it only supports `OAUTHBEARER` (but we can also add `SASL_MECHANISM` and `SASL_JAAS_CONFIG` as path variables).
- users must know the fully qualified handler class name.

But we can take a different approach.
What if we use a path variable like `configSecurityProtocol=SASL_SSL` which tells the consumer that it should take the sasl options from the minion config (we'll have to inject the `Config config`)?
On the plus side
- easier setup of consumer tasks for users.

On the minus side
- only preconfigured handler class is supported.

And I have some concerns about path variable names, so I'm asking for your advice about the naming.

Afanas10101111:
Hi @lbroudoux !
I've added an implementation of another approach with a new commit. It looks to be more compatible with the producer's previous modification to support custom handler.

Afanas10101111:
Hello @lbroudoux 
Could you take a look at this PR please?

lbroudoux:
Yes sure. I was busy releasing the 1.8.1 last week. I'd like to tackle this one for the next iteration.

Issue #895, opened 2023-08-25
Add explicit error message in mock response


sfraigneau:
### Reason/Context

When we send a request to a corrupt mock, we receive an error 500 from microcks without detail about the error.
An explict error message would be appreciate to understand what's wrong

### Description

For instance, messages like :
 - exception in groovy script : xxx
 - unknown response name : xxx
 - ...

Apoorva64:
Groovy Errors currently only works for the SOAP mocks, would it be possible to add it the the REST mocks, or maybe refactor to extract this behaviour that is common to all types of mocks?

Apoorva64:
@lbroudoux i think we could close this one as this issue is solved by the live tracing feature https://microcks.io/blog/microcks-1.13.0-release/

lbroudoux:
I think you're right! Closing this one.

Issue #819, opened 2023-05-19
Microcks returns HTTP 500 when calling non-existent API


hguerrero:
### Describe the bug

When calling a non-existent service using GraphQL, the server returns HTTP 500

```
{
    "timestamp": 1684525088128,
    "status": 500,
    "error": "Internal Server Error",
    "path": "/graphql/Book+Graph+API/1.0"
}
```

### Expected behavior

It should return HTTP 404 or another code

### Microcks version or git rev

1.6.0

### Install method (`docker-compose`, `helm chart`, `operator`, `docker-desktop extension`,...)

docker-desktop extension

hguerrero:
I agree that 404 maybe not be the correct response, but we certainly need to avoid throwing the HTTP 500. Something in the 400 range makes sense. 406 or 415 could be good options.

lbroudoux:
Assigning this to the next release (`1.8`?) as we cannot change this behavior for the next patch version (`1.7.1`)

edeandrea:
I would argue that 404 is exactly the correct response. The client asked for something which doesn't exist, which is what 404 means.

lbroudoux:
We discussed this topic while reviewing a PR [here](https://github.com/microcks/microcks/pull/1007) with @sfraigneau.  

It looks like we have different cases to handle, and I've come to the proposition below. Could you please check it out and tell us what you think? Ideally, the different cases must be handled in the same way across the different supported protocols.

1. Service + version pair cannot be found. They are unknown from Microcks => returns a `404`
2. We're missing a resource (WSDL or something else) that is mandatory to validate input => returns a `412` (client placed a validation precondition on the request, but it cannot be fulfilled)
3. Validation is executed and fails => return a `400`
4. Dispatcher rules evaluation (Script or `JSON_BODY`) fails => returns a `500`
5. Response cannot be found from evaluated dispatch criteria => returns a `400` (client provides inadequate dispatch criteria)
6. Response can be found but clearly indicates that it is an unknown resource => returns a `404`

Thanks for your help in sorting this out.

lbroudoux:
Changes are now applied to both `RestController` and `GraphQLController`.

PR #1007, opened 2023-11-28
Add explicit error message (SOAP response)

sfraigneau:
### Description

Return an explicit message when an error occurs while mocking soap response

Note : The error message is returned as the response body like others. Using Spring exception handler may be more appropriate.

### Related issue(s)
#895

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/SoapController.java
+++ b/webapp/src/main/java/io/github/microcks/web/SoapController.java
@@ -45,6 +45,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.server.ResponseStatusException;
 import org.xml.sax.InputSource;
 
 import jakarta.servlet.http.HttpServletRequest;
@@ -115,6 +116,11 @@ public class SoapController {
       log.info("Service name: " + serviceName);
       // Retrieve service and correct operation.
       Service service = serviceRepository.findByNameAndVersion(serviceName, version);
+      if (service == null) {
+         return new ResponseEntity<String>(
+            String.format("The service %s with version %s does not exist!", serviceName, version),
+            HttpStatus.NOT_FOUND);
+      }
       Operation rOperation = null;
 
       // Enhancement : retrieve SOAPAction from request headers
@@ -134,6 +140,9 @@ public class SoapController {
       // Enhancement : if not found, try getting operation from soap:body directly!
       if (rOperation == null) {
          String operationName = extractOperationName(body);
+         if (action == null || action.length() > 0) {
+            action = operationName;
+         }
          log.debug("Extracted operation name from payload: {}", operationName);
 
          if (operationName != null) {
@@ -154,7 +163,13 @@ public class SoapController {
          if (validate != null && validate) {
             log.debug("Soap message validation is turned on, validating...");
 
-            Resource wsdlResource = resourceRepository.findByServiceIdAndType(service.getId(), ResourceType.WSDL).get(0);
+            List<Resource> wsdlResources = resourceRepository.findByServiceIdAndType(service.getId(), ResourceType.WSDL);
+            if (wsdlResources.isEmpty()) {
+               return new ResponseEntity<String>(
+                  String.format("The service %s with version %s does not have a wsdl!", serviceName, version),
+                  HttpStatus.NOT_FOUND);
+            }
+            Resource wsdlResource = wsdlResources.get(0);
             List<String> errors = SoapMessageValidator.validateSoapMessage(wsdlResource.getContent(), new QName(service.getXmlNS(), rOperation.getInputName()),
                      body, resourceUrl);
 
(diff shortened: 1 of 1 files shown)


webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -115,6 +116,11 @@ public ResponseEntity<?> execute(
       log.info("Service name: " + serviceName);
       // Retrieve service and correct operation.
       Service service = serviceRepository.findByNameAndVersion(serviceName, version);
+      if (service == null) {
+         return new ResponseEntity<String>(
+            String.format("The service %s with version %s does not exist!", serviceName, version),
+            HttpStatus.NOT_FOUND);
```
lbroudoux, line 122: We've got a specific discussion on this topic here: https://github.com/microcks/microcks/issues/819
Can you check and join the party?

webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -134,6 +140,9 @@ public ResponseEntity<?> execute(
       // Enhancement : if not found, try getting operation from soap:body directly!
       if (rOperation == null) {
          String operationName = extractOperationName(body);
+         if (action == null || action.length() > 0) {
+            action = operationName;
+         }
```
lbroudoux, line 145: I don't get this condition purpose... At this point of flow, we haven't found `rOperation` using the `action` so we're looking at operation extracted from the body payload. 
I think we could directly set `action = operationName` no?

webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -154,7 +163,13 @@ public ResponseEntity<?> execute(
          if (validate != null && validate) {
             log.debug("Soap message validation is turned on, validating...");
 
-            Resource wsdlResource = resourceRepository.findByServiceIdAndType(service.getId(), ResourceType.WSDL).get(0);
+            List<Resource> wsdlResources = resourceRepository.findByServiceIdAndType(service.getId(), ResourceType.WSDL);
+            if (wsdlResources.isEmpty()) {
+               return new ResponseEntity<String>(
+                  String.format("The service %s with version %s does not have a wsdl!", serviceName, version),
+                  HttpStatus.NOT_FOUND);
```
lbroudoux, line 170: Agree with the check but we need to have a decision on status code.

webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -200,6 +223,10 @@ public ResponseEntity<?> execute(
          if (!responses.isEmpty()) {
             int idx = DispatchStyles.RANDOM.equals(dispatcher) ? RandomUtils.nextInt(0, responses.size()) : 0;
             response = responses.get(idx);
+         } else {
+            return new ResponseEntity<String>(
+               String.format("The response %s does not exist!", dispatchContext.dispatchCriteria()),
+               HttpStatus.BAD_REQUEST);
```
lbroudoux, line 229: This is matching with my opinion in https://github.com/microcks/microcks/issues/819 😉 
We have to conclude as a community on this point, though.

webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -179,13 +194,21 @@ public ResponseEntity<?> execute(
...
+               dispatchContext = getDispatchCriteriaFromScriptEval(dispatcherRules, body, request);
+            } else if (DispatchStyles.RANDOM.equals(dispatcher)) {
+               dispatchContext = new DispatchContext(DispatchStyles.RANDOM, null);
+            } else {
+               return new ResponseEntity<String>(
+                  String.format("The dispatch %s is not supported!", dispatcher),
+                  HttpStatus.NOT_FOUND);
+            }
+         } catch (ResponseStatusException e) {
+            return new ResponseEntity<String>(e.getMessage(), e.getStatusCode());
```
lbroudoux, line 211: Shouldn't we force the `ResponseEntity` status code to be 500 - Internal Error as this happens when we got a dispatcher evaluation exception?

lbroudoux (commented):
Thanks a lot for the PR. I had a few comments and questions I have on status code... Looking forward to getting your opinions on this!

webapp/src/main/java/io/github/microcks/web/SoapController.java
sfraigneau: As the requested action in header takes precedence over the requested action in the body, I return the header one if the client asks for action in both header and body.
I set the action variable only in case the client asks for action in body only

webapp/src/main/java/io/github/microcks/web/SoapController.java
sfraigneau: Yes, we can force the status code. 
In fact I would have preferred to catch the ResponseStatusException at upper level (by using an exception handler for instance) instead of using ResponseEntity to return errors. What are you thoughts about it ?

webapp/src/main/java/io/github/microcks/web/SoapController.java
lbroudoux: Yes, I agree with this one but it's to get a global view on which status code to use depending on the failure case.

webapp/src/main/java/io/github/microcks/web/SoapController.java
lbroudoux: But `action == null || action.length() > 0` is only true if action is null or not empty, right? 
So we have 2 cases here:
* client didn't ask for action in header (action is null because not found in header)
* client did ask for action in header (action is not empty as found in header).

Are you sure it's a `||`? I don't see why we couldn't just have `action = operationName`... sorry.

webapp/src/main/java/io/github/microcks/web/SoapController.java
lbroudoux: I wonder if `412` wouldn't be more appropriate: a validation precondition set by the client cannot be fulfilled...

lbroudoux:
> > Also I have to figure out how to remove those 3 commits that were already approved into another PR now merged. Do you have any idea?
> 
> I forgot to sign-ff my commit and after force push, these commits appear in the PR. No idea why I will redo the PR starting from the microcks branch. I think it's the best option ;)

Maybe rebasing your branch on `1.8.x` and force-pushing the PR could do the trick?

webapp/src/main/java/io/github/microcks/web/SoapController.java
sfraigneau: Yes you're right, the condition was wrong. It should be :
`action == null || action.length() == 0`

I changed the way to check by using hasText from Spring utils. It is more readable.

webapp/src/main/java/io/github/microcks/web/SoapController.java
sfraigneau: By default, spring will handle exceptions. If you raise an exception in a controller, spring will return an 500 error automatically.

To control what spring do with our exceptions, there are differents alternatives.

The easiest way is to use ResponseStatusException to control the http status in the exception itself.
If we want to add the exception message to the error message returned by Srping, we just have to enable it in properties : 
`server.error.include-message=ALWAYS`
The drawback is that any exception message will be returned to the user. Sometimes, for security reason, we do not what that. By the way, I didn't think that it is a problem for microcks.
We could also want to log the exceptions when Spring catch them for us. This is possible again in property file : 
`spring.mvc.logResolvedException=true`

If we want to have full control of our exceptions, the best alternative is to define our own exception handler in a ControllerAdvice bean : https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-exceptionhandler.html

webapp/src/main/java/io/github/microcks/web/SoapController.java
sfraigneau: 412 is good for me

lbroudoux:
Thanks a lot @sfraigneau !

PR #1012, opened 2023-11-29
Add customization for labels/annotations

### LINKED ISSUE #1005: Customization of labels and annotations
### Reason/Context

To improve standardisation of kubernetes based application, companies are using tools like https://github.com/open-policy-agent/gatekeeper that ensure such standardisation.
If a resource is not valid, the application cannot be installed and must comply with the rules in policy.
The first level of standardisation is often the labelling and annotating part of kubernetes resources such as pod/deployments/services ... 

### Description

Microck chart should allow such labels and annotation customization for all resources defined.

### Implementation ideas

There are many possibilities : 
- provide ad-hoc templating for annotation/label customization
- or the most portable one would be to migrate the chart to [bitnami's common template](https://github.com/bitnami/charts/blob/4683a83b8e8ce9c19e944b59c3f17dfe4a1a43a6/template/README.md)

Note using kustomize would be a valid solution, but would have an impact of helm releases deployment.

vietk:
### Description
This commit mainly allow the customization of common labels and annotations for all resources of this helm chart. Also, in some case, specify memory limits to avoid pod to be evicted first.
Also, in some case, specify memory request to avoid host computation of the available memory and avoid OOM killed.
Apply resource request == limit for java containers Provide documentation for new/missing helm values.

I opt for the simple solution ad-hoc customization; Using bitnami charts is a bit of a big challenge :D 
Hope this solution is also fine for you

### Related issue(s)
Fixes #1005


install/kubernetes/microcks/values.yaml
```diff
@@ -54,6 +54,12 @@ microcks:
 postman:
   image: quay.io/microcks/microcks-postman-runtime:latest
   replicas: 1
+  resources:
+    requests:
+      cpu: 30m
```
lbroudoux, line 59: `30m` seems a bit low for this container. Do you know if this value is based on measured experiments? Some users in the community are using `postman-runtime` intensively, and I think this low value will impact them... Could you please stick to the previous version? (actually not specifying `request.cpu`). You'll still be able to include this value in your own.

install/kubernetes/microcks/values.yaml
```diff
@@ -23,10 +23,10 @@ microcks:
   resources:
     requests:
       cpu: 200m
-      #memory: 400Mi
+      memory: 400Mi
     limits:
       #cpu: 500m
-      memory: 512Mi
+      memory: 400Mi
```
lbroudoux, line 29: I haven't run new benchmarks recently (with Java 17 and new SpringBoot 3.1) and I remember `512Mi` was previously chosen as a compromise between resource consumption and startup time (according to probes settings as well). Have you performed new measurements that could make us change to `400Mi`?

install/kubernetes/microcks/values.yaml
```diff
@@ -88,7 +94,7 @@ keycloak:
       memory: 400Mi
     limits:
       #cpu: 500m
-      #memory: 512Mi
+      memory: 400Mi
```
lbroudoux, line 97: `512Mi` was inherited from previous benchmarks with older versions of Keycloak. I don't have new benchmarks with KeycloakX that (typically) would make this value lower... Can you please stick with the previous value? That should not impact you as you're not deploying Keycloak and thus it'll not impact existing users as well.

We have to perform new benchmarks in the future but are waiting for some work done regarding Open Telemetry support addition and Grafana dashboard availability.

lbroudoux (changes requested):
Thanks a lot @vietk for the PR.

I suggested a few changes regarding the default value of resources - in order to stick with current default values waiting for a future benchmark session that would allow us to get measurements for change decisions.

Could you review them?

install/kubernetes/microcks/values.yaml
vietk: I haven't run any benchmark and I based the new request/limit value on the the one that was commented out: 400Mi.

> I haven't run new benchmarks recently (with Java 17 and new SpringBoot 3.1) and I remember 512Mi was previously chosen

I will modify the PR accordingly to your benchmark

install/kubernetes/microcks/values.yaml
vietk: good catch, I totally made up this one.
So I'll remove it from the default value

vietk:
Hello, 
I modified the PR accordingly to your comments

lbroudoux (approved):
LGTM! Thanks a lot!

PR #1013, opened 2023-11-29
feat: #1010 Skip generation of Helm chart secrets

### LINKED ISSUE #1010: Helm chart compatibility with ArgoCD (gitops pull model)
### Reason/Context

When using ArgoCD to deploy microcks, I'm getting warnings due to kind:Secret manifests generated by the chart.
Resource kind:Secret are typically not synchronized by ArgoCD because:

- If the secret value is in clear in the git repo, it's not really secured.
- If the secret value is randomly generated by helm, it will create an infinite loop, as 'helm template' must always generate the same output for a given git commit (https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#random-data)

### Description

There is already `microcks.ingressSecretRef` which allows to re-use an existing secret (for instance one generated by cert-manager) and skips the generation of microcks-microcks-ingress-secret

I'd like to have the same for:
* microcks-mongodb-connection
* microcks-microcks-grpc-secret
* microcks-keycloak-admin

### Implementation ideas

In https://github.com/microcks/microcks/blob/master/install/kubernetes/microcks/templates/secret.yaml:

- For microcks-mongodb-connection, we can add an if on the presence of `.Values.mongodb.secretRef.secret` and that should be backward compatible (it's actually redudant to generate it)

- For microcks-microcks-grpc-secret, I think we need to create a microcks.grpcIngressSecretRef or maybe reuse ingressSecretRef ?
I'm not super familiar with TLS validation in gRPC, but I assume the HTTP certificate could be re-used, as long as it has the right list of dnsNames in it ?

- For microcks-keycloak-admin, probably a new keycloak.secretRef (similar to mongodb.secretRef) ?

I could contribute a PR for these changes.

rquinio1A:
Fixes #1010

Changes:

- Add keycloak.secretRef to re-use an existing secret
- Take ingresses/microcks.ingressSecretRef/microcks.generateCert into account also for generating gRPC ingress secret
- Take mongodb.secretRef into account for generating mongodb secret

Validation: no secret gets generated with

helm template . --set microcks.ingressSecretRef.secret=mysecret --set keycloak.ingressSecretRef.secret=mysecret --set keycloak.secretRef.secret=mysecret --set mongodb.secretRef.secret=mysecret | grep -A 3  "kind: Secret"


lbroudoux:
Hey @rquinio1A,

Thanks a lot for this PR. I should have warned you before that this one should have been created (and then contributed back) to the `1.8.x` branch where the developments are happening at the moment. I tried to change the base branch in the PR, but it then pulled some polluting commits that will make our life harder when merging.

Could you please re-issue this PR from `1.8.x`? Also, we've got another PR in review from one of your colleagues (see #1005) that impacts all the Kubernetes resources. Then it may be easier to wait for its merge before applying yours (still in the goal of making our life easier during review and merge). Would you mind waiting a bit? (I think it should be ok for beginning of next week)

Thanks again!

install/kubernetes/microcks/templates/secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if and (.Values.ingresses) (not .Values.microcks.ingressSecretRef) (.Values.microcks.generateCert) }}
```
lbroudoux, line 1: Secret for GRPC is needed even if we do not expose the ingresses because it is used directly by the GRPC server within the Spring Boot app. I think it would be better to introduce a new variable `microcks.grpcSecretRef` and just have:

```
{{- if hasKey  .Values.microcks.grpcSecretRef) }}
```

Also, this new secret ref must be used in `deployment.yaml` where we mount the GRPC secret into the main pod.

lbroudoux (changes requested):
Thanks a lot. Can you please check the comment related to GRPC secret? This one is a bit different as it is not related to ingress.

install/kubernetes/microcks/templates/secret.yaml
rquinio1A: I've added microcks.grpcSecretRef

lbroudoux: Awesome! Will have a look at it in a few hours.

lbroudoux (approved):
LGTM! Thanks a lot!
BTW: would you be open to having a private chat about your Microcks usages? I'd really like to learn more 😉  I've looked for you on LinkedIn but looks like we cannot connect...

PR #1015, opened 2023-12-05
feat: microcks#895 Add explicit error message (SOAP response)

sfraigneau:
### Description
Return an explicit message when an error occurs while mocking soap response

Note : The error message is returned as the response body like others. Using Spring exception handler may be more appropriate.

### Related issue(s)

#895


sfraigneau:
Arf, my previous PR has been closed. I'll try to reopen it : https://github.com/microcks/microcks/pull/1007

PR #1025, opened 2023-12-15
Optimize: Extract Tabset in service view into another component and render only the selected tab

### LINKED ISSUE #979: UI hangs when loading / rendering large payload dataset
### Describe the bug

When you have a request with a lot of responses the UI frezzes upon opening large data.

### Expected behavior

The UI should not freeze but show a loading icon or somthing

### Actual behavior

The UI freezes and becomes unusable

### How to Reproduce?

1. load this yaml file 
https://file.io/pQ4qP3zQLNhN

2. go to the service and open the operation. The UI will freeze for 4~sec. Of cause adding examples or making them bigger worsen the problem.

1.8.x

docker-compose

### Additional information

_No response_

Apoorva64:
Extracted <tabset/> section from the component (webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page) into the exchanges-tabset component (webapp/src/main/webapp/src/app/pages/services/{serviceId}/_components/exchanges-tabset)

Resolves #979

### DIFF
--- a/webapp/src/main/webapp/src/app/app.module.ts
+++ b/webapp/src/main/webapp/src/app/app.module.ts
@@ -82,7 +82,7 @@ import { ArtifactUploaderDialogComponent } from './pages/importers/_components/u
 import { HubPageComponent } from './pages/hub/hub.page';
 import { HubPackagePageComponent } from './pages/hub/package/package.page';
 import { HubAPIVersionPageComponent } from './pages/hub/package/apiVersion/apiVersion.page';
-
+import { ExchangesTabsetComponent } from './pages/services/{serviceId}/_components/exchanges-tabset/exchanges-tabset.component';
 import json from 'highlight.js/lib/languages/json';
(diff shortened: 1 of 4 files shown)


lbroudoux:
Can you explain the optimizations you've made to make things better regarding rendering speed? I saw that you extracted stuff into a component and that already make things a lot clearer (Thank you!) but could you be more specific regarding the rendering performance as stated in #979?

Apoorva64:
The should render function is called in the template to decide if the content of a response should be processed(highlighting, formatting of code). 

The function checks if the tab index its given is equal to the active tab. If it is it returns true so we render the response of this tab else if it is not equal it returns false so we do not render the response.

In all cases all the headers of the tabset are still rendered which gives us the same output visually and with this optimisation the ui feels way snappier with large examples

lbroudoux (approved):
LGTM! Thank you very much for this optimization!

PR #1049, opened 2024-01-22
fix(#1016): support for query params array

### LINKED ISSUE #1016: [OpenAPI] Test & Mock an endpoint with a query param array
### Describe the bug

Hi!
In an OpenAPI contract, when you have an endpoint that defines an array of query params, mocking or testing doesn't seem to work properly:

```yaml
parameters:
  - in: query
    name: ids
    schema:
      type: array
      items:
        type: string
```

**EDIT**: The OpenAPI Spec has multiple ways to handle query parameter arrays, the default one is to repeat the parameter key multiple times. (style: form, explode: true)
see: [https://swagger.io/docs/specification/serialization/](https://swagger.io/docs/specification/serialization)

### Expected behavior

1. In microcks GUI, you should get a correct mocking url:
image
2. Calling this url while changing the ids passed in the array should match correctly against the examples
3. When running a test, it should send those parameters correctly

### Actual behavior

There are multiple weird behaviours at play here:

1. The mocking URL displayed in microcks GUI shows the array as a JSON Array
image

2. When calling that malformed mock URL, you get an error
image

3. Mock answers with an example when you manually create a correct URL but
image

4. The mock always answers with the same case regardless of which values you pass
image

5. When testing, Microcks sends the query parameters as a JSON Array too

### How to Reproduce?

To reproduce this, you can just import the following OpenAPI definition: 

```yaml
openapi: 3.0.3
info:
  title: Someone API
  description: Someone API
  version: 1.0.0
servers:
  - url: "{protocol}://{hostname}:{port}/{basePath}/"
    variables:
      protocol:
        enum:
          - https
          - http
        default: http
      hostname:
        default: localhost
      port:
        default: "8080"
      basePath:
        default: api
tags:
  - name: Someone
paths:
  '/someone/something':
    get:
      security:
        - bearerAuth: [ ]
      description: Search someone based on a list of ids
      operationId: searchStuff
      tags:
        - StuffiStuffo
      parameters:
        - in: query
          name: ids
          required: true
          description: |-
            Get stuff using a list of ids
          schema:
            type: array
            items:
              type: string
          examples:
            SINGLE_SUCCESS:
              value: 
                - "123"
                - "1234"
            SECOND_CASE:
              value: 
                - "666"

        - in: query
          name: dataReturned
          schema:
            type: string
            default: 'SUMMARY'
        - $ref: "#/components/parameters/pageParam"
        - $ref: "#/components/parameters/sizeParam"
      responses:
        '200':
          description: SomeoneInfo
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/SomeonePage"
              examples:
                SINGLE_SUCCESS:
                  value:
                    pageSize: 10
                    pageNumber: 0
                    totalPage: 1
                    totalElements: 1
                    isFirst: true
                    isLast: true
                    content:
                      - id: "123"
                        firstname: "Avril"
                        lastname: "Lavigne"
                SECOND_CASE:
                  value:
                    pageSize: 10
                    pageNumber: 0
                    totalPage: 1
                    totalElements: 1
                    isFirst: true
                    isLast: true
                    content:
                      - id: "666"
                        firstname: "Jack"
                        lastname: "Black"
components:
  securitySchemes:
    bearerAuth: # arbitrary name for the security scheme
      type: http
      scheme: bearer
      bearerFormat: JWT
  schemas:
    Page:
      type: object
      properties:
        content:
          type: array
          items:
            type: object
        pageNumber:
          type: integer
        pageSize:
          type: integer
        totalPage:
          type: integer
        totalElements:
          type: integer
        isFirst:
          type: boolean
        isLast:
          type: boolean

    Someone:
      type: object
      description: Someone information
      properties:
        id:
          type: string
        firstname:
          type: string
        lastname:
          type: string
          
    SomeonePage:
      allOf:
        - $ref: "#/components/schemas/Page"
        - type: object
          properties:
            content:
              default: []
              items:
                $ref: '#/components/schemas/Someone'
  parameters:
    pageParam:
      in: query
      name: pageNumber
      schema:
        type: integer
        minimum: 0
        default: 0
    sizeParam:
      in: query
      name: pageSize
      schema:
        type: integer
        default: 10
```

9c54fcef9668c749ceb63d1a50f12597d2df5c8a

Docker run latest nightly version of microcks-uber

### Additional information

After looking at the code, the issue seems to lie in the OpenApiImporter class, in extractParametersByExample>getExampleValue.
The example value is considered to be a JsonArray, which it should not be.
I'm down to do a PR for this

aeoncl:
### Description
Add support for array of query params by repeating the same key multiple times.
I tried a very simple fix in order to change the least amount of code, changing the Map<String. String> of params to a Multimap, allowing us to have the same key multiple times.
The OpenApiSpec parser, when dealing with queryParams, will add all the examples under the same key in this multimap.

This doesn't fully implement all the ways that the OpenAPI spec can deal with array of query params but this works with the default configuration and is a really simple fix. In order to support the full spec, i think a big rewrite will be needed

### Related issue(s)

Resolves  #1016


webapp/src/main/java/io/github/microcks/util/DispatchCriteriaHelper.java
```diff
@@ -15,17 +15,18 @@
...
+import com.google.common.collect.TreeMultimap;
+
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
+import java.util.*;
```
lbroudoux, line 24: Can you please let the full explicit list of imports?

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
```diff
@@ -15,44 +15,22 @@
...
-import io.github.microcks.util.MockRepositoryImportException;
-import io.github.microcks.util.MockRepositoryImporter;
-import io.github.microcks.util.ReferenceResolver;
-import io.github.microcks.util.URIBuilder;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import io.github.microcks.domain.*;
+import io.github.microcks.util.*;
```
lbroudoux, line 23: Can you please avoid stars imports and let the full explicit list of needed classes?

webapp/src/test/java/io/github/microcks/util/DispatchCriteriaHelperTest.java
```diff
@@ -218,6 +220,24 @@ public void testBuildFromParamsMap() {
...
+    paramsMap.put("status", "available");
+    paramsMap.put("status", "busy");
+
+    // 2 parameters should be taken into account, one with two values according to rules.
+    String dispatchCriteria = DispatchCriteriaHelper.buildFromParamsMap("page && status", paramsMap);
+    assertEquals("?page=1?status=available?status=busy", dispatchCriteria);
+
+    // 1 parameter with two values should be taken into account according to rules
+    dispatchCriteria = DispatchCriteriaHelper.buildFromParamsMap("status", paramsMap);
+    assertEquals("?status=available?status=busy", dispatchCriteria);
```
lbroudoux, line 238: Are the params values sorted in natural order? We need a way to be predictable regarding how a dispatch criteria is built either from the imported spec or from the incoming Http request.

webapp/src/test/java/io/github/microcks/util/URIBuilderTest.java
```diff
@@ -127,4 +129,27 @@ public void testBuildURIFromPatternWithMap() {
...
+    parameters.put("year", "2018");
+    parameters.put("month", "05");
+    parameters.put("status", "published");
+    parameters.put("status", "proofred");
+    parameters.put("page", "0");
+
+    // Test with old wadl like template format.
+    String pattern = "http://localhost:8080/blog/{year}/{month}";
+    String uri = URIBuilder.buildURIFromPattern(pattern, parameters);
+    assertTrue("http://localhost:8080/blog/2018/05?page=0&status=published".equals(uri)
```
lbroudoux, line 146: Shouldn't we have both status (`proofred` and `published`) present in the URI here?

lbroudoux (changes requested):
Thanks a lot for this PR. This approach looks nice and - as you said - requires minimal effort to get implemented for at least default behavior. I have some concerns regarding how imports are formatted and how/if param values are sorted when build a dispatch criteria. Can you check this out please?

webapp/src/test/java/io/github/microcks/util/DispatchCriteriaHelperTest.java
aeoncl: They should be in natural order, because i kept the use of a TreeMap to sort the keys, replacing the TreeMap by a MultiTreemap

webapp/src/test/java/io/github/microcks/util/URIBuilderTest.java
aeoncl: They should :D will fix that

webapp/src/test/java/io/github/microcks/util/DispatchCriteriaHelperTest.java
aeoncl: Oh right, the keys are sorted by natural order, but not the values, i'll check what i can do

aeoncl: After checking the Multimap doc, it seems that both the values and the keys are ordered naturally by default 
image

lbroudoux: Excellent! Thank you.

lbroudoux (approved):

aeoncl:
Hi, is there some issue with the sonarqube config??

lbroudoux:
No. That's ok. There are just too many things to do that prevent me from finishing this... 😉

aeoncl:
ahahah good luck man :D lmk if i need to do something

PR #1051, opened 2024-01-23
feature(helm-Chart): Added the possibility of using a secret with a keystore

romain66:
### Description

- Added the possibility of using a secret with a keystore
- By default we keep the cacerts file and its password in order to avoid any modification of the JAVA options.

### Related issue(s)
Since the lib change for SSO connection with Keycloak, the certificate must be trusted by Microcks.
In order to be able to use personalized certificates you must be able to modify the Java Keystore.


lbroudoux:
Hey @romain66!

Thanks a lot for the contribution, as it addresses different use cases that have no solution at the moment. 
I had a few suggestions to make your proposed solution more generic, though ...

* Regarding the name of the Helm parameter, I think that `customSecret` would be better as some users would also have this need for MongoDB certificates...

* I don't think that it needs to be `enabled` as we used `enabled` for components or features. I would have said that we can just rely on absence/presence of the attribute. So the default value in `values.yaml` would be commented and people who wanted to add a custom secret mount would just specify this block - like we did for the Kafka secret:

```yaml
customSecretRef:
   secret: my-custom-secret
   key: cacert-keystore
```

* Regarding the `mountPath` I would prefer something that is not internal to the image (Java version may change) and much more spotted as `custom`. We could use something like `/deployments/config/custom/secret`

* Regarding this change of location and the password, I would prefer not to rely on internal defaults but force people to change custom `JAVA_OPTIONS` to adapt stuff to their own settings. In my opinion, the solution described in https://github.com/orgs/microcks/discussions/893 makes a lot of sense.

Sorry if it makes you change stuff, but I think it's for the best.

Thanks again for having taken the time to put everything together for these issues!

install/kubernetes/microcks/values.yaml
```diff
@@ -52,6 +52,16 @@ microcks:
   # Enabling/disabling mock invocation stats.
   mockInvocationStats: true
 
+  # Enabling/disabling Custom keystore for JAVA
+  # Use an existing secret with the provided keystore
+  # You must concerve cacerts to avoid modifications on JAVA configuration
+  customKeystore:
```
lbroudoux, line 58: Can we change to simple:

```yaml
customSecretRef:
   secret:
   key:
```

We don't have to enable it, just put it in the values. By default, this block will be commented out in the `values.yaml` file.

install/kubernetes/microcks/templates/deployment.yaml
```diff
@@ -160,6 +160,11 @@ spec:
             mountPath: "/deployments/config/kafka/keystore"
             {{- end }}
           {{- end }}
+          {{- if .Values.microcks.customKeystore.enabled }}
+          - name: "{{ .Values.microcks.customKeystore.secretRef.secret }}"
+            mountPath: "/etc/java/java-17-openjdk/java-17-openjdk-17.0.9.0.9-2.el9.x86_64/lib/security/{{ .Values.microcks.customKeystore.secretRef.key }}"
```
lbroudoux, line 165: Can we change to something less tied to the imnage internals? Something more spotted as custom would make more sense IMO. Something like `/deployments/config/custom/secret` for example.

lbroudoux (changes requested):
Thanks @romain66 for contribution. Please review the enhancements suggestions.

romain66:
Which do you think is better:
Force JAVA options on the helm chart or add a comment to explain how to configure it if a custom keystore is used?

lbroudoux:
I think that documenting the configuration is better as it allows us to cover many more use cases. I would not make the Helm chart change or set `JAVA_OPTIONS` values... it's up to the user to do so.

install/kubernetes/microcks/values.yaml
lbroudoux: Thanks for the change @romain66 !

install/kubernetes/microcks/templates/deployment.yaml
lbroudoux: Thanks for the new commit @romain66 !

lbroudoux (approved):

lbroudoux (approved):
LGTM. Thank you!

Issue #1069, opened 2024-02-06
Response Delays per exemple


Apoorva64:
### Reason/Context

I need to be able to simulate a timeout in a specific exemple.

### Description

When adding response delays per exemple, we can have an exemple that will simulate a timeout.

### Implementation ideas

- [ ] Add microcks custom annotations per exemple
- [ ] Add delay data to Response Object 
- [ ] Use the Response Object in RestController

PR #1081, opened 2024-02-12
Delays per exemple

Apoorva64:
### Description

Reason/Context
I need to be able to simulate a timeout in a specific exemple.

Description
When adding response delays per exemple, we can have an exemple that will simulate a timeout.

This PR includes
- [x] Add microcks custom annotations per exemple
- [x] Add delay data to Response Object
- [x] Use the Response Object in RestController to make a delay

### Related issue(s)
#1069

### DIFF
--- a/commons/model/src/main/java/io/github/microcks/domain/Response.java
+++ b/commons/model/src/main/java/io/github/microcks/domain/Response.java
@@ -33,7 +33,7 @@ public class Response extends Message {
    private String mediaType;
    private String dispatchCriteria;
    private boolean isFault = false;
-   private Long delay = 0L;
+   private long delay = 0L;
 
    public String getId() {
       return id;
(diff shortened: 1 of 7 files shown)


Apoorva64:
@lbroudoux The "delay" variable isn't saving to the database do you have an idea why it dosen't do so?

Apoorva64:
OpenApi file can be found [here](https://github.com/Orange-OpenSource/microcks/blob/39c0d3e1e2704b02d0d878b60d9206e71205219c/webapp/src/test/resources/io/github/microcks/util/openapi/openapi-extensions-delay.yaml)

Example with 1 second delay
image
Example with 5 second delay
image
Example with no delay
image

Apoorva64:
> Please hold on this as I'd like to have a discussion on this use-case and how it would fit the whole picture.

Where are we going to discuss this > github, discord or zulip>?

PR #1085, opened 2024-02-13
Add missing headers for DynamicCorsFilter

Apoorva64:
### Description
When doing CORS request, we can send a header("Access-Control-Request-Headers") to request specific headers.
This PR implements this additional generation in the DynamicCorsFilder

### Related issue(s)
#

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/filter/DynamicCorsFilter.java
+++ b/webapp/src/main/java/io/github/microcks/web/filter/DynamicCorsFilter.java
@@ -24,8 +24,11 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.List;
 
 /**
  * Servlet filter that set the response CORS headers accordingly the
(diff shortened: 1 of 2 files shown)


lbroudoux:
Hi @Apoorva64 
Can you point to some reference documentation that explains this behavior? I am no Cors expert and was not aware of this mechanism... Could you share some reference link so that I'll understand if this behavior is standard? Thanks,

Apoorva64:
CORS requests are explained here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

The specifics of Access-Control-Request-Headers are explained here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Headers

lbroudoux:
Do you have a small HTML+JS page that would allow testing the different combination? In the case of https://github.com/microcks/microcks-quarkus/issues/22#issuecomment-1941485212 it seems that `content-type` is actually part of the request headers but not detected present into `Access-Control-Allow-Headers` in the preflight response.

Apoorva64:
where should i execute thoses commands>? in webapp ?

lbroudoux:
> @lbroudoux where should i execute thoses commands>? in webapp ?

Sorry 😉  No, from `distro/uber` folder.

Apoorva64:
image
Works !

lbroudoux (approved):
LGTM and you've got a working test case on Quarkus Super Heroes! Woot!

PR #1090, opened 2024-02-15
Fix path to mount secrets on microcks deployments

romain66:
### Description
HELM CHART Update:
Avoid double directory when adding secrets in Microcks deployment.


lbroudoux (approved):
Thanks for the fix.

Issue #1100, opened 2024-02-21
Add proxy support for the rest controller


Afanas10101111:
### Reason/Context

Sometimes it may be useful to forward Microcks' client to real endpoint, so we can implement some logic when we need mock and when a real call. With such feature we don't have to change the url on the client's side if we want both (mock or real) behaviors.

### Description

To do this, we can add Spring proxy support and use some header to say that we want proxy behavior
```
      responses:
        "200":
          headers:
            x-microcks-proxy-to:
              schema:
                type: string
              examples:
                proxy:
                  value: 'http://realhost:8080/some-service/v1/real-pastry'
```

### Implementation ideas

<a href="https://github.com/microcks/microcks/pull/1101">Idea</a>

lbroudoux:
Hello. We had some discussions in the past about this feature, but we weren't sure how it fit our target use cases... You give some details above but I'd like to learn more ;-)
Also, on the implementation side, we were thinking about something lighter than this. We try to avoid new dependency when possible ... I'll have a look at this.

lbroudoux:
For the implementation, I found this article a while ago (https://blogs.ashrithgn.com/writing-a-reverse-proxy-service-in-spring-boot/) and I was thinking if it couldn't be a good starting point ... It seems easy, lightweight and would avoid pulling out new dependencies. What do you think?

Afanas10101111:
I've <a href="https://github.com/microcks/microcks/pull/1101/commits/0a2342ab56c05fc670fbf19ce6ff5b0c8f09f4f0">tried</a> this approach and it works just fine. I'll test it a bit more next week. But it's still demo :) Integration into the dispatching mechanism looks more interesting.

Afanas10101111:
> I wonder if we could expand this to their protocol than rest? At least Soap and maybe GraphQL...

I think it's worth at least trying :) If we can make an elegant implementation with the dispatching mechanism, I hope that expanding to other protocols will be a matter of technique.

> Got another question: are you using Microcks at your company? Is that an enhancement suggestion coming from your real-lige experience?

Yes and yes :) Features of Microcks look very interesting, and now I'm working on integrating it into the development process. So all my issues are inspired by integration or users problems.

lbroudoux:
Have you ever considered taking the time to chat with us and joining our adopters list?? 😉

Afanas10101111:
[...]

PR #1101, opened 2024-02-21
feat: Add proxy support for the rest controller

Afanas10101111:
### Description

Add proxy support for the rest controller

- new dependency `spring-cloud-gateway-mvc`
- new property to pass an auth header thowgh proxy
- proxy logic in `RestController.java`
- test

### Related issue(s)

- <a href="https://github.com/microcks/microcks/issues/1100">#1100</a>


Afanas10101111:
Hello @lbroudoux !
I've done with the first version of the `PROXY` dispatcher implementation. I need flexibility in deciding to handle mocks or to proxy, so for now I haven't come up with anything better than reusing the `FALLBACK` dispather approach.
Could you take a look at this please?

lbroudoux:
Targeting the `1.9.x` branch as this is the development branch for the next `1.9.1` release

webapp/src/main/java/io/github/microcks/util/dispatcher/ProxySpecification.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Represent the specification of a Proxy evaluation. <code>dispatcher</code> should be the original dispatcher to apply
+ * with its <code>dispatcherRules</code> companion. If no response actually found by mock controller, the original
+ * request will be forwarded to the <code>proxyUrl</code>.
+ */
+@JsonPropertyOrder({ "dispatcher", "dispatcherRules", "proxyUrl" })
```
lbroudoux, line 26: I'm not sure why this class is needed as, basically, I think the dispatcher specification needs a backend or proxy URL that could fit into the `dispatcherRules` property living at the operation level. We introduced this complex structure for `FALLBACK` as it acts like a try-catch:
* 1st try the original dispatcher specified in `dispatcher` and using `dispatcherRules`
* 2nd if 1st is not found, we finally use the `fallback` to get the default response name.

So that for fallback, we ended up with:

```yaml
  /pastry/{name}:
    summary: Global operations on pastries
    get:
      x-microcks-operation:
        dispatcher: FALLBACK
        dispatcherRules: |
          {
            "dispatcher": "URI_PARTS",
            "dispatcherRules": "name",
            "fallback": "Our default (not found - 404) response"
          }
```

I don't think this complexity is needed for proxy stuff but I may miss something.

lbroudoux:
Hi @Afanas10101111 

Thank you very much for taking the time to put this to life. I made a bunch of comments in the review as I think it doesn't need to be that complicated. Please let me know if I am missing something though.

webapp/src/test/resources/io/github/microcks/util/openapi/pastry-with-proxy-openapi.yaml
```diff
@@ -0,0 +1,89 @@
...
+    summary: Global operations on pastries
+    get:
+      x-microcks-operation:
+        dispatcher: PROXY
+        dispatcherRules: |
+          {
+            "dispatcher": "URI_PARTS",
+            "dispatcherRules": "name",
+            "proxyUrl": "http://localhost/rest/pastry-proxy/1.0.0/real-pastry/{{ request.path[2] }}"
+          }
```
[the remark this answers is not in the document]
Afanas10101111: I need more control over the choice when we need to proxy or mock. In the test example, we get a stub for the name defined in the schema, and proxy the request for any other one. I can even come up with more complex (and real-life) requirements - to proxy a request only during certain hours. That's why I played around the SCRIPT dispatcher in one of the [previous commits](https://github.com/microcks/microcks/pull/1101/commits/95ef5068704c53a7120defe725b35eea757b3247#diff-9221377ec824b9a79247881ce41786353687133945988037369bd39c48ed99c2R21).

webapp/src/main/java/io/github/microcks/util/dispatcher/ProxySpecification.java
Afanas10101111: Please check out my comment below

webapp/src/main/java/io/github/microcks/util/DispatchCriteriaHelper.java
```diff
@@ -457,6 +467,36 @@ public static String extractFromParamMap(String paramsRule, Map<String, String>
       return result.toString();
    }
 
+   public static DispatcherDetails extractDispatcherWithRules(Operation operation) {
```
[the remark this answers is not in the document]
Afanas10101111: Please check out my comment below

Afanas10101111:
Hello @lbroudoux !
All this complexity is needed only for control. Unfortunately, I couldn't think of any better idea to implement. If you have any suggestions, please share them with me and I'll try to implement them.

lbroudoux:
Thanks for the explanation! I got it!
I wonder if we couldn't have 2 dispatchers in fact... 

One simple PROXY that proxies everything to the backend and one PROXY_FALLBACK that is the one you wrote and that fallbacks to backend if no match on existing mock has been found. 

What do you think?

Afanas10101111:
Yes, I even see the simple PROXY way as another way to control behavior using an uploaded schema. More clear way for someone who needs manual control.
I'll be back with the implementation in the next few days.

lbroudoux:
Let me know; otherwise, I can try to spend some time on this by the beginning of next week.

Afanas10101111:
The first version of the simple PROXY dispatcher is ready.
@lbroudoux , could you take a look at this please?

webapp/src/test/resources/io/github/microcks/util/openapi/pastry-with-proxy-openapi.yaml
```diff
@@ -0,0 +1,84 @@
...
+  license:
+    name: MIT License
+    url: https://opensource.org/licenses/MIT
+paths:
+  /pastry/{name}:
+    summary: Global operations on pastries
+    get:
+      x-microcks-operation:
+        dispatcher: PROXY
+        dispatcherRules: http://localhost/rest/pastry-proxy/1.0.0/real-pastry/{{ request.path[2] }}
```
lbroudoux, line 20: That's not exactly what I had in mind... I was thinking of something even simpler where the `dispatcherRules` is just the base URL of a backend service that implements the same API. If reusing another mock (here called `real-backend`) served by Microcks, it would have done something like this: 

```yaml
 x-microcks-operation:
   dispatcher: PROXY
   dispatcherRules: http://localhost/rest/real-backend/1.0.0
```

Afanas10101111: Thanks for this example! I've refactored both dispatcher according to this logic.

webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
```diff
@@ -137,4 +152,92 @@ public void testHeadersTemplating() {
...
+
+      // Set original URL to the dispatcher
+      Service service = serviceRepository.findByNameAndVersion("pastry-proxy", "1.0.0");
+      Operation op = service.getOperations().stream().filter(o -> o.getName().endsWith("GET /pastry/{name}"))
+            .findFirst().orElseThrow();
+      op.setDispatcherRules(op.getDispatcherRules().replaceFirst("http://localhost", getServerUrl())
+            .replaceFirst("pastry-real", "pastry-proxy"));
+      serviceRepository.save(service);
+
+      ResponseEntity<String> response = restTemplate.getForEntity("/rest/pastry-proxy/1.0.0/pastry/realDonut",
```
lbroudoux, line 195: I think I understand how `/pastry/realDonut` can be resolved by the proxy...
As `pastry-for-proxy-openapi.yaml` only defines a `/pastry/donut` request path, it actually not match and rely on the fact that Microcks will still answer with the only existing question...

It would have been better IMHO to have something like:
`/pastry/donut` -> not found then redirect to proxy -> `/pastry/donut` -> found on proxy

I will probably adapt a bit this sample to that test will not rely on Microcks mechanism for the backend and will be easier understandable

lbroudoux (approved):
It looks excellent! Thank you so much for taking the time to put this together. I will probably adjust a bit one of the test but that's really cool!

webapp/src/test/java/io/github/microcks/web/RestControllerIT.java
Afanas10101111: Yes, I agree that this example is not clear enough. Its main task is to verify that if someone defines the base URL of mitchrocks, it will not result in a circular query. To simulate this, I change the path in [line 192](https://github.com/microcks/microcks/pull/1101/files/c20afb0eb16e37d548b75561f073cefaa7ec2475#diff-d7164af73de0fcfd4aba7238940c3655bb45e4b702f76316e3118f7428448cd9R192). But, of course, it's not the best solution.

Afanas10101111:
Thank you for your valuable comments! I'm glad we got this to the finish line :)
Please let me know if I can help expand this to other protocols, as we discussed in parent issue.

lbroudoux:
For sure! Reusing this and making it available for GraphQL and Soap would be awesome.

PR #1104, opened 2024-02-25
Reduce amount of tokens needed in prompt for creation of rest API examples using openai

### LINKED ISSUE #1103: Reduce the amount of tokens in AI prompt for API Rest example generation whit openAI
## Reason/Context

When the API specification has a large number of operations, the prompt sent to openAI may cause an error by exceeding the number of tokens supported by the model selected for generating the example.

## Description

I think the prompt to create the example doesn't need the full specification; instead, use only the specification for the operation to create examples.

## Implementation ideas

Extend the functionality of methods such as removeExamplesInNode and removeExamplesFromOpenAPISpec (AICopilotHelper.java) to resolve schema references and return only the JsonNode with the specification of the operation for creating the prompt

LG-Lopez:
### Description

Add prompt generation optimization for generate rest API examples with openai

### Related issue(s)

Resolves #1103

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
+++ b/webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
@@ -93,7 +93,7 @@ public class AICopilotHelper {
          - example: %1$d
            message:
              headers:
-               header_1: <value 1>
+               <header_name>: <value 1>
              payload: <message payload>
          """;
 
@@ -133,45 +133,6 @@ public class AICopilotHelper {
          builder.append(String.format(UNIDIRECTIONAL_EVENT_EXAMPLE_YAML_FORMATTING_TEMPLATE, i+1));
       }
       return builder.toString();
-   }
-
-   /** Some AI will reuse already present examples in the spec. This method cleans a specification from its examples. */
-   protected static String removeExamplesFromOpenAPISpec(String specification) throws Exception {
-      JsonNode specNode;
-      boolean isJson = specification.trim().startsWith("{");
-
-      if (isJson) {
-         specNode = JSON_MAPPER.readTree(specification);
-      } else {
-         specNode = YAML_MAPPER.readTree(specification);
-      }
-
-      // Remove examples recursively from the root.
-      removeExamplesInNode(specNode, specNode);
-
-      if (isJson) {
-         return JSON_MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(specNode);
-      }
-      return YAML_MAPPER.writeValueAsString(specNode);
-   }
-
-   private static void removeExamplesInNode(JsonNode specNode, JsonNode node) {
-      JsonNode target = followRefIfAny(specNode, node);
-      if (target.getNodeType() == JsonNodeType.OBJECT) {
-         if (target.has("examples")) {
-            ((ObjectNode) node).remove("examples");
-         }
-         Iterator<Map.Entry<String, JsonNode>> fields = target.fields();
-         while (fields.hasNext()) {
-            removeExamplesInNode(specNode, fields.next().getValue());
-         }
-      }
-      if (target.getNodeType() == JsonNodeType.ARRAY) {
-         Iterator<JsonNode> elements = target.elements();
-         while (elements.hasNext()){
-            removeExamplesInNode(specNode, elements.next());
-         }
-      }
    }
 
    /** Transform the output respecting the {@code REQUEST_RESPONSE_EXAMPLE_YAML_FORMATTING_TEMPLATE} into Microcks domain exchanges. */
@@ -393,7 +354,7 @@ public class AICopilotHelper {
    * Some AI will reuse already present examples and x-microcks-operation in the
    * spec. This method cleans a specification from its examples.
    */
-  protected static String removeTagsFromOpenAPISpec(String specification, String operationName) throws Exception {
+  protected static String removeTokensFromSpec(String specification, String operationName) throws Exception {
     JsonNode specNode;
     boolean isJson = specification.trim().startsWith("{");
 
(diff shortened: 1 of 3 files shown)


webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
```diff
@@ -388,4 +388,100 @@ private static JsonNode followRefIfAny(JsonNode spec, JsonNode referencableNode)
    private static JsonNode getNodeForRef(JsonNode spec, String reference) {
       return spec.at(reference.substring(1));
    }
+
+  /**
+   * Some AI will reuse already present examples and x-microcks-operation in the
+   * spec. This method cleans a specification from its examples.
+   */
+  protected static String removeTagsFromOpenAPISpec(String specification, String operationName) throws Exception {
```
lbroudoux, line 396: I thought at first that this method was dealing with OpenAPI `tags` within a spec. Maybe `removeTokensFromOpenAPISpec()` would be a better name... Also, as I understood it, it will replace the `removeExamplesFromOpenAPISpec()` method so we can delete it. Is that correct?

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
```diff
@@ -388,4 +388,100 @@ private static JsonNode followRefIfAny(JsonNode spec, JsonNode referencableNode)
...
+    }
+    if (target.getNodeType() == JsonNodeType.ARRAY) {
+      Iterator<JsonNode> elements = target.elements();
+      while (elements.hasNext()) {
+        resolveReferenceAndRemoveTagsInNode(specNode, elements.next());
+      }
+    }
+  }
+
+  protected static void reduceSpecSize(JsonNode specNode, String operationName) throws Exception {
```
lbroudoux, line 445: Wouldn't `filterOpenAPIOperation()` be a more appropriate name?

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
```diff
@@ -388,4 +388,100 @@ private static JsonNode followRefIfAny(JsonNode spec, JsonNode referencableNode)
...
+
+    if (isJson) {
+      specNode = JSON_MAPPER.readTree(specification);
+    } else {
+      specNode = YAML_MAPPER.readTree(specification);
+    }
+
+    // Resolve schemas and Remove examples recursively from the root.
+    resolveReferenceAndRemoveTagsInNode(specNode, specNode);
+    reduceSpecSize(specNode, operationName);
```
lbroudoux, line 408: Wouldn't be more efficient to switch both method invocations? Something like:
```java
reduceSpecSize(specNode, operationName);
resolveReferenceAndRemoveTagsInNode(specNode, specNode);
```
It may avoid resolving references and removing spec elements in operations that would have been dropped later on.

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
```diff
@@ -388,4 +388,100 @@ private static JsonNode followRefIfAny(JsonNode spec, JsonNode referencableNode)
...
+    String operationPathName[] = operationName.split(" ");
+    String verb = operationPathName[0].toLowerCase();
+    String path = operationPathName[1];
+
+    JsonNode pathsSpec = ((ObjectNode) specNode).get("paths");
+    JsonNode pathSpec = ((ObjectNode) pathsSpec).get(path);
+
+    List<String> keysToKeepInRoot = List.of("openapi", "paths", "info");
+    List<String> keysToKeepInPaths = List.of(path);
+    List<String> keysToKeepInPath = List.of(verb);
```
lbroudoux, line 455: Do you think this could be adapted/generalized so that it could be reused also for cleaning up an AsyncAPI specificatino?

webapp/src/test/java/io/github/microcks/util/ai/AICopilotHelperTest.java
```diff
@@ -422,4 +427,441 @@ public void testParseRequestResponseOutputSplitWithJson() {
...
+    assertTrue(specNode.get("paths").has("/pastry/{name}"));
+    assertFalse(specNode.get("paths").has("/pastry"));
+    // Ensure security property is removed
+    assertFalse(specNode.get("paths").get("/pastry/{name}").has("security"));
+  }
+
+  @Test
+  public void testResolveReferenceAndRemoveTagsInNode() throws Exception {
+
+    // Cadena JSON de ejemplo
```
lbroudoux, line 527: Is it possible to put this in english? Thanks.

webapp/src/test/java/io/github/microcks/util/ai/AICopilotHelperTest.java
```diff
@@ -422,4 +427,441 @@ public void testParseRequestResponseOutputSplitWithJson() {
...
+    ObjectMapper mapper = new ObjectMapper();
+    // Create a test JsonNode with references and tags
+    JsonNode specNode = mapper.readTree(jsonSpecNode);
+    JsonNode expectedSpecNode = mapper.readTree(jsonExpectedSpecNodeResult);
+
+    // Call the resolveReferenceAndRemoveTagsInNode method
+    AICopilotHelper.resolveReferenceAndRemoveTagsInNode(specNode, specNode);
+
+    // Verify that the references and tags are removed correctly
+    assertEquals(expectedSpecNode, specNode);
```
lbroudoux, line 559: Is it robust to compare raw json like this? What about using something JSONAssert? I think we already got it as a test dependency.

lbroudoux (changes requested):
Thanks a lot for this contribution! That's awesome and a very very good one! I've put some comment out there to be sure to fully understand how things are supposed to work and suggest some enhancements.

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
LG-Lopez: Thanks for the quick response, I chose to rename the method as removeTokensFromSpec that way it remains agnostic to the type of specification, whether openapi or AsyncAPI. On the other hand, this method effectively replaces removeExamplesFromOpenAPISpec() so I follow your advice to eliminate it

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
LG-Lopez: Of course I changed the method name by filterOpenAPISpec()

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
LG-Lopez: I understand, but it is necessary that the resolveReferenceAndRemoveTagsInNode method (now renamed to resolveReferenceAndRemoveTokensInNode) be executed first because in this method the schema references are resolved, this is necessary before removing the "components" token.

webapp/src/main/java/io/github/microcks/util/ai/AICopilotHelper.java
LG-Lopez: This is a great idea, I added a method (filterAsyncAPISpec) to generalize and clean up AsyncAPI specifications

webapp/src/test/java/io/github/microcks/util/ai/AICopilotHelperTest.java
LG-Lopez: Offcourse!

webapp/src/test/java/io/github/microcks/util/ai/AICopilotHelperTest.java
LG-Lopez: I understand that, but since the resolveReferenceAndRemoveTagsInNode method (renamed resolveReferenceAndRemoveTokensInNode) mutates the specNode object and does not return an object to compare, it is necessary to create a JsonNode in the test method. Now, I modified the validation to be done with String objects instead of JsonNode objects.

lbroudoux (approved):
LGTM. Thank you very much!

PR #1141, opened 2024-04-18
fix: add a null check before asserting on mediatype of the test http …

### LINKED ISSUE #1140: OpenAPITestRunner doesn't extract http call return code correctly
### Describe the bug

Hi, 
while playing with Microcks by testing the new feature with the `x-microcks-refs`, I found an issue with the OpenAPITestRunner.

It seems that after the test the runner will try to extract the HTTP return code from the call done during the test, but when I use the `x-microcks-refs` for an endpoint that doesn't return any value, the OpenAPITestRunner tries to extract the return code by looking at the return type of the response. but in this particular case there is no return type since there is nothing to return.

as this snippet shows from the `OpenAPITestRunner.java` at line 134
```java
if (!expectedResponse.getMediaType().equalsIgnoreCase(contentType)) {
    log.debug("Response Content-Type does not match expected one, returning failure");
    lastValidationErrors = List
        .of(String.format("Response Content-Type does not match expected one. Expecting %s but got %s",
        expectedResponse.getMediaType(), contentType));
   return TestReturn.FAILURE_CODE;
}
``` 

In this particular case `expectedResponse.getMediaType()` will always result to null and thus cause a NullPointerException

### Expected behavior

It should do a null check of the media type before asserting on it

### Actual behavior

Null is not checked causing a NullPointerException when using `x-microcks-refs`

### How to Reproduce?

Create a small spring boot app with this openAPI spec

```yaml
openapi: 3.0.3
info:
  title: 'Sample REST API'
  description: "Sample API"
  version: 1.0.0
tags:
  - name: A sample API

paths:
  /resource/{id}:
    delete:
      tags:
        - Delete-resources
      operationId: deleteResourceByID
      parameters:
        - in: path
          required: true
          schema:
            type: number
          name: id
          examples:
            ex:
              value: 1
      responses:
        '204':
          description: Delete something
          x-microcks-refs:
            - ex

```

With a simple controller that doesn't do much 
```java
package com.example.demo;

import backend.api.DeleteResourcesApi;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.http.ResponseEntity;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;

import java.math.BigDecimal;

import static org.springframework.http.ResponseEntity.noContent;

@RestController
@RequestMapping("/api")
public class Controller implements DeleteResourcesApi {
    private static final Logger LOGGER = LoggerFactory.getLogger(Controller.class);

    @Override public ResponseEntity<Void> deleteResourceByID(BigDecimal id) {
        LOGGER.info("DELETING ID {}", id.toString());
        return noContent().build();
    }
}
```
And start a microcks standalone or with a test container and just run the test against the small backend

1.9.0

test-container, local docker setup

### Additional information

Awesome project, with awesome feature, thanks for all this work

Snorkell:
### Description

- Add a null check before asserting on media type in case of not having a return value from http call

### Related issue(s)
fixes #1140

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPITestRunner.java
+++ b/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPITestRunner.java
@@ -131,7 +131,8 @@ public class OpenAPITestRunner extends HttpTestRunner {
                return TestReturn.FAILURE_CODE;
             }
 
-            if (expectedResponse.getMediaType() != null && !expectedResponse.getMediaType().equalsIgnoreCase(contentType)) {
+            if (expectedResponse.getMediaType() != null
+                  && !expectedResponse.getMediaType().equalsIgnoreCase(contentType)) {
                log.debug("Response Content-Type does not match expected one, returning failure");
                lastValidationErrors = List
                      .of(String.format("Response Content-Type does not match expected one. Expecting %s but got %s",
--- a/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPITestRunner.java
+++ b/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPITestRunner.java
@@ -131,7 +131,7 @@ public class OpenAPITestRunner extends HttpTestRunner {
                return TestReturn.FAILURE_CODE;
             }
 
-            if (!expectedResponse.getMediaType().equalsIgnoreCase(contentType)) {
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
Thanks!

PR #1144, opened 2024-04-19
feat: Add support for object query parameters

### LINKED ISSUE #1143: Support for object query parameters
### Reason/Context

Hello, as always thanks for this awesome project.

I was playing around with the OpenAPI specs when I tried to implement a very specific feature fo the spec -> Object in query parametes

https://swagger.io/docs/specification/serialization/

So by default the specs says that we can use `type: object` in the schema of a query param but it seems that the OpenAPIImporter class is not aware of this feature.

I know it's very specific but it's a default feature of the spec so I thought it could be nice to be able to test and mocks api with this feature in it. 

### Description

So the specs (with the default config) says that any non nested object defined as query parameter will be serialized as the following example
```yaml
User:
  name: bob
  age: 45
```
to something like this un an url
`http://myawesomeAPI/someendpoint?name=bob&age=45`

### Implementation ideas

The idea would be in the case of having a object as query params the dispatch rule won't be tied to the param name but the object fields name as the specs describe.

With that the importer should parse the example as independant parameter and not as a whole json object as it's done right now

A pull request will follow soon with some tests

Agains, thanks for all of this 😄

Snorkell:
### Description

- Add support for object query parameters as the openAPI specs describe https://swagger.io/docs/specification/serialization/
this only support the default configuration for handling query parameters with `style: form` and `explode: true`

### Related issue(s)
Resolves #1143

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
+++ b/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
@@ -303,9 +303,9 @@ public class OpenAPIImporter extends AbstractJsonRepositoryImporter implements M
                   for (JsonNode current : (ArrayNode) exampleValue) {
                      exampleParams.put(parameterName, getValueString(current));
                   }
-               } else if(PARAMETERS_QUERY_VALUE.equals(parameterType) && exampleValue.isObject()){
+               } else if (PARAMETERS_QUERY_VALUE.equals(parameterType) && exampleValue.isObject()) {
                   final var fieldsIterator = ((ObjectNode) exampleValue).fields();
-                  while(fieldsIterator.hasNext()){
+                  while (fieldsIterator.hasNext()) {
                      var current = fieldsIterator.next();
                      exampleParams.put(current.getKey(), getValueString(current.getValue()));
                   }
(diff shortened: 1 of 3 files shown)


webapp/src/test/resources/io/github/microcks/util/openapi/object-query-params.yaml
```diff
@@ -0,0 +1,48 @@
...
+      parameters:
+        - in: query
+          name: irrelevantHere
+          schema:
+            $ref: '#/components/schemas/Person'
+          examples:
+            dune:
+              value:
+                lastName: Atreides
+                firstName: Paul
```
lbroudoux, line 25: Nice example ;-)

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIImporter.java
```diff
@@ -45,16 +46,10 @@
...
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
lbroudoux, line 49: I'd prefer explicit import style but I'll fix that just after.

lbroudoux (approved):
LGTM! Thanks!

Issue #1128, opened 2024-04-03
Add more tests


mcruzdev:
### Reason/Context

This proposal aims to increase the quality and maintainability of the project, adding more unit and integration tests.

### Description

Add tests for JobService

lbroudoux:
Thanks for the suggestion! Do you plan to have an attached PR on this one or have you faced a particular situation on `JobService` that would require some specific cautions?

mcruzdev:
I would like to work on baby steps, one class per time, because I normally I contribute in my free times hehe, what do you think?

mcruzdev:
It is a good way to know in deep dive the project too, I want to grow the coverage of the project testing some scenario, this is my idea.

mcruzdev:
Seeing the SonarQube map, I started with `AMQPMessageConsumptionTaskITest` class.

lbroudoux:
Hey @mcruzdev! Thanks again for the PR regarding `AMQPMessageConsumptionTask`!
Do you think you'll push some other ones and we will keep this issue open, or is it better to close this one, and you may reopen some other PR in the future? WDYT?

mcruzdev:
Hi @lbroudoux do not close, I want to add more tests :)

lbroudoux:
Awesome!

lbroudoux:
Just updated the issue name to scope it to `1.9.1` release (for now 😉)

lbroudoux:
Hi @mcruzdev !
I'm gonna close this issue as `1.9.1` is now ready for release. New tests are more than welcome but it will be on another branch to prepare the next `1.10` iteration.
See you around!

PR #1150, opened 2024-04-23
Add integration test to UploadArtifactController 

mcruzdev:
### Considerations

> [!IMPORTANT]
> We need to be careful changing the specification on  samples, because can break those tests.

### Description

- Rename `AMQPMessageConsumptionTaskITest` to `AMQPMessageConsumptionTaskIT`
- Add integration test to ÙploadArtifactController`

### Related issue(s)

#1128

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
+++ b/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
@@ -51,8 +51,11 @@ public class UploadArtifactController {
    /** A simple logger for diagnostic messages. */
    private static Logger log = LoggerFactory.getLogger(UploadArtifactController.class);
 
-   @Autowired
    private ServiceService serviceService;
+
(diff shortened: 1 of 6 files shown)


webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
```diff
@@ -51,11 +51,14 @@ public class UploadArtifactController {
    /** A simple logger for diagnostic messages. */
    private static Logger log = LoggerFactory.getLogger(UploadArtifactController.class);
 
-   @Autowired
    private ServiceService serviceService;
 
+   public UploadArtifactController(ServiceService serviceService) {
```
lbroudoux, line 56: `Autowired` import should be removed as well on L27. I'll do that after merge.

lbroudoux (approved):
Great! Thanks!

Issue #1138, opened 2024-04-15
Hide User informations and Logout buttons in unauthenticated mode


lbroudoux:
### Reason/Context

When running Microcks without authentication/Keycloak support, we still have the User's information and Logout buttons that are available. This make things confusing such as in #1086 

### Description

We should hide those informations and buttons if not authentication support.

### Implementation ideas

This is a UI only hack ;-)

lbroudoux:
Fixed in #1179. Thanks to @Apoorva64

PR #1179, opened 2024-05-24
fix: disable dropdown and info when in unauthenticated mode

Apoorva64:
### Description

- Checks when Microcks is using Auth by checking the config service
- If Microcks is not using Auth it removes the user Info with the dropdown
- else if Microcks is using auth it show the user Info

Not using Auth:
image
Using Auth:
image

### Related issue(s)
#1138 

### Issues Found (Not fixed in this pr)
- Preferences points to empty url "http://localhost:8080/#0"
image
- When Keycloak user has no "first Name" or "last Name". Microcks doesn't show a user. 
image
We could use "prefered_username" in this case

### DIFF
--- a/webapp/src/main/webapp/src/app/components/vertical-nav/vertical-nav.component.html
+++ b/webapp/src/main/webapp/src/app/components/vertical-nav/vertical-nav.component.html
@@ -29,7 +29,7 @@
           </li>
         </ul>
       </li>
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
LGTM.

lbroudoux:
Thanks @Apoorva64 !

PR #1194, opened 2024-06-07
Code smell in ressource utils

Apoorva64:
Remove code smells in RessourceUtils.java

There were no tests for this class so i've created one but i don't know if is aligned with what we want.


lbroudoux:
That looks nice! This class is used for generating OpenAPI and AsyncAPI specs from the templates we have in `/webapp/src/main/resources/templates`. It could be nice to integrate those templates in the unit tests to be sure we get correct generation with no carriage return issues. Thanks!

Apoorva64:
@lbroudoux i've added tests for async and openapi3.0 but not for 2.0.
Do we support openapi 2.0?

PR #1294, opened 2024-08-29
Add missing guard on jsonNode schema path access

### LINKED ISSUE #1293: OpenApiTestRunner don't break the test if the response schema is empty
### Describe the bug

Hello, 
a colleague of mine was working with microcks using testconainers by doing tests on his API while we found a small issue when validation the API spec.

if for some reason the response part of an endpoint does not have a shcema by mistake, the tests are running in a loop without any result.

After searching in the source coude I found that int the `OpenAPISchemaValidator` class in the method `validateJsonMessage`
at line `213` there is no guard on the `schema` field of the json node.

In a perfect world every objet shoudl have a schema definition but in our case it was missing which is an error from our side. 

### Expected behavior

We expected that microcks would raise an exception of some sort to validate our spec and tells us what's wrong

### Actual behavior

The tests stay in some sort of pending state without any result 

### How to Reproduce?

Create an api with a response of the mimetype you want but omit the schema part

1.10.x

test-container

### Additional information

Again, thanks for all the work, a small PR with a guard for this will follow soon

Snorkell:
### Description

add a guard on the jsonNode schema path access to prevent testing staying in pending state.

### Related issue(s)

Fixes #1293

### DIFF
--- a/commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
+++ b/commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
@@ -210,6 +210,10 @@ public class OpenAPISchemaValidator {
       // Build a schema object with responseNode schema as root and by importing
       // all the common parts that may be referenced by references.
       JsonNode schemaNode = messageNode.path("schema").deepCopy();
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
LGTM. Merci!

PR #1304, opened 2024-09-08
Add support for Postman Collection structured versioning

### LINKED ISSUE #1297: Add support for Postman Collection structured versioning
### Reason/Context

As of today, we retrieve a Postman Collection version using 2 different ways: via `info.version` attribute or via the `info.description` text field. We do things as follow:
* If `info.version` is defined, use it as a raw string for finding the matching API version in Microcks,
* If `info.version` is not provided, check the `info.description` text field and check that it starts with `version=<something>`. In that case, `<something>` is considered as the API version.
* If nothing found, raise an exception.

It looks like the Postman Collection Format v2.1.0 (and probably the v2.0.0 as well) has another way of specifying the version using a structured object in the `info.version` path (see https://schema.postman.com/collection/json/v2.1.0/draft-07/docs/index.html). We should also support this way of doing things in order to ease integration with other tools.

### Description

As defined per https://schema.postman.com/collection/json/v2.1.0/draft-07/docs/index.html, you can also define a Postman Collection version like this:

```json
{
  "info": {
    "version": {
      "major": 3,
      "minor": 4,
      "patch": 5,
      "identifier": "beta-4"
    }
  }
}
```

The `major`, `minor` and `path` attributes are mandatory integers according to the schema. `identifier` is an optional human-friendly identifier that should make more sense than the version numbers.

Hence, we propose to adapt the way of getting version that way:
* First, use `info.version.identifier` if provided
* If not, then check `info.version.major` and others to extract a version like `3.4.5` in the above example,
* If not provided, then consider `info.version` as raw text,
* Then, finally, consider the `info.description` text and `version=<something>` marker as the latest option.

What do you think? Comments and suggestions are welcome!

### Implementation ideas

_No response_

Harsh4902:
### Description
- Added support for postman collection structured versioning.
### Related issue(s)
- Fixes #1297

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/postman/PostmanCollectionImporter.java
+++ b/webapp/src/main/java/io/github/microcks/util/postman/PostmanCollectionImporter.java
@@ -145,7 +145,16 @@ public class PostmanCollectionImporter implements MockRepositoryImporter {
       // On v2.1 collection format, we may have a version attribute under info.
       // See https://schema.getpostman.com/json/collection/v2.1.0/docs/index.html
       if (collection.path("info").has(SERVICE_VERSION_PROPERTY)) {
-         version = collection.path("info").path(SERVICE_VERSION_PROPERTY).asText();
+         if (collection.path("info").path(SERVICE_VERSION_PROPERTY).has("identifier"))
+            version = collection.path("info").path(SERVICE_VERSION_PROPERTY).path("identifier").asText();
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
Thanks for the contribution! This looks good! Just have been a bit better with the unit test ;-) I'm gonna write it as it's fairly trivial. Thanks again!

PR #1317, opened 2024-09-12
Issue#1298

### LINKED ISSUE #1298: Filter Test Runners depending on available artifacts - in the UI
### Reason/Context

When launching a new Test from the UI, we don't filter the list of Test Runners. So typically, it's possible to launch a `Postman` test even if a Postman Collection has not been correctly imported and attached to our API | Service.

This may lead to errors for new users or people who haven't noticed that a previous import has gone wrong.  

### Description

We could enhance the "Test creation form" page by requesting the available resources attached to the current API | Service and then removing from the default list - initialized based on the Service type - the Test Runners that have no matching resources (eg. no Postman Collection attached).

Obviously, this would work only when creating a test from the UI, we could not provide this help when directly launch from the Microcks REST API. What we could check, though, is that each Test Runner implementation actually raises a meaningful error message when trying to launch a test with missing resources. 

### Implementation ideas

_No response_

Harsh4902:
### Description

This PR adds `isRunnerAvailable` filter into `test-create.page` to check whether a runner is available for a given service or not.

### Related issue(s)
Fixes #1298

### DIFF
--- a/webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
+++ b/webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
@@ -24,9 +24,10 @@ import { Notification, NotificationEvent, NotificationService, NotificationType 
 import { ServicesService } from '../../../services/services.service';
 import { TestsService } from '../../../services/tests.service';
 import { SecretsService } from '../../../services/secrets.service';
-import { Operation, Service } from '../../../models/service.model';
+import {Contract, Operation, Service} from '../../../models/service.model';
 import { TestRunnerType, OAuth2ClientContext } from '../../../models/test.model';
 import { Secret } from '../../../models/secret.model';
+import {ContractsService} from "../../../services/contracts.service";
 
 @Component({
   selector: 'app-test-create-page',
@@ -40,6 +41,7 @@ export class TestCreatePageComponent implements OnInit {
   serviceId: string;
   testEndpoint: string;
   runnerType: TestRunnerType;
+  contracts: string[]
   showAdvanced = false;
   submitEnabled = false;
   notifications: Notification[];
@@ -55,7 +57,7 @@ export class TestCreatePageComponent implements OnInit {
   filteredOperation: string;
   removedOperationsNames: string[] = [];
 
-  constructor(private servicesSvc: ServicesService, public testsSvc: TestsService, private secretsSvc: SecretsService,
+  constructor(private servicesSvc: ServicesService, private contractsSvc: ContractsService, public testsSvc: TestsService, private secretsSvc: SecretsService,
               private notificationService: NotificationService, private route: ActivatedRoute, private router: Router) {
   }
 
@@ -78,6 +80,14 @@ export class TestCreatePageComponent implements OnInit {
         this.initializeFromPreviousTestResult(fromTestId);
       }
     });
+    this.route.paramMap.pipe(
+      switchMap((params: ParamMap) =>
+        this.contractsSvc.listByServiceId(params.get('serviceId'))
+      )
+    ).subscribe((contracts: Contract[]) => {
+      this.contracts = contracts.map((contract : Contract) => contract.type.toString());
+    });
+
   }
 
   getSecrets(page: number = 1): void {
@@ -245,5 +255,9 @@ export class TestCreatePageComponent implements OnInit {
         complete: () => console.log('Observer got a complete notification')
       }
     );
+  }
+
+  public isRunnerAvailable(runner: string) : boolean{
+    return this.contracts.includes(runner)
   }
 }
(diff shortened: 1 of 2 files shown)


Harsh4902:
@lbroudoux I am not sure with the exact name of `contract.type` for given runner, so please review it and suggest required changes.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
```diff
@@ -24,9 +24,10 @@ import { Notification, NotificationEvent, NotificationService, NotificationType
 import { ServicesService } from '../../../services/services.service';
 import { TestsService } from '../../../services/tests.service';
 import { SecretsService } from '../../../services/secrets.service';
-import { Operation, Service } from '../../../models/service.model';
+import {Contract, Operation, Service} from '../../../models/service.model';
```
lbroudoux, line 27: Could you please follow the same format? => `{ Contract, Operation, Service }`

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
```diff
@@ -24,9 +24,10 @@ import { Notification, NotificationEvent, NotificationService, NotificationType
...
 import { TestRunnerType, OAuth2ClientContext } from '../../../models/test.model';
 import { Secret } from '../../../models/secret.model';
+import {ContractsService} from "../../../services/contracts.service";
```
lbroudoux, line 30: Could you please follow the same format? => `{ ContractsService }`
and put this import within the "services" block? => insert it at line 24

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
```diff
@@ -246,4 +256,8 @@ export class TestCreatePageComponent implements OnInit {
       }
     );
   }
+
+  public isRunnerAvailable(runner: string) : boolean{
+    return this.contracts.includes(runner)
```
lbroudoux, line 261: I think this cannot work as contract types (see the list on https://github.com/microcks/microcks/blob/1.10.x/webapp/src/main/webapp/src/app/models/service.model.ts#L106) are different than runner types (see the list on https://github.com/microcks/microcks/blob/1.10.x/webapp/src/main/webapp/src/app/models/test.model.ts#L83).

Typically, a contract type is `POSTMAN_COLLECTION` but the test runner type is `POSTMAN`. You'll need to initialize a translation map between the two.

lbroudoux (changes requested):
I highlight some changes. I think it cannot work OOTB.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
Harsh4902: Made this change.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
Harsh4902: Made this change.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
Harsh4902: Here I am passing value of contract type based on the runner option. As we have fixed runner options, I think we can put string value of contract type for the checking directly. You can see this at `line 61` of `test-create.page.html`

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
```diff
@@ -246,4 +256,8 @@ export class TestCreatePageComponent implements OnInit {
       }
     );
   }
+
+  public isRunnerAvailable(runner: string) : boolean{
```
lbroudoux, line 260: I saw you latest comment and seems all good. However, I think this function should be renamed to something like `isContractAvailable(contractType: string)` as we check the contract availability from a runner in the select option.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
```diff
@@ -40,6 +41,7 @@ export class TestCreatePageComponent implements OnInit {
   serviceId: string;
   testEndpoint: string;
   runnerType: TestRunnerType;
+  contracts: string[]
```
lbroudoux, line 44: Renaming to `contractTypes` would be more precise.

lbroudoux (changes requested):
LGTM but renaming property and function would make things clearer I think.

webapp/src/main/webapp/src/app/pages/tests/create/test-create.page.ts
Harsh4902: Yup, made this change.

lbroudoux (approved):
LGTM! Thank you!

Issue #1350, opened 2024-09-30
Documentation pages error 404


ffischer-mdf:
### Describe the bug

Documentation links are broken in Microcks application ("Help") or in https://github.com/microcks (readme.md), we get a 404 error page.
Examples : 
https://microcks.io/documentation/using/postman/
https://microcks.io/documentation/using/soapui/

lbroudoux:
I think everything is now fixed.

PR #1352, opened 2024-09-30
Fix 404 on conformance metrics 'Learn More'

jjatria:
### Description

The Conformance Metrics "Learn More" link is broken and points to a 404. This PR points it to what appears to be the new version of that page.

### Related issue(s)

This is not one of the links mentioned in #1350 but it seems to be in the same larger group.


lbroudoux:
Thanks a lot for the PR! I just changed the base as we're now working on the `1.11.x` branch for next iteration.

lbroudoux (approved):
LGTM. Thank you!

PR #1358, opened 2024-10-01
feat: Adding affinity and tolerations support to the helm chart

mvtavares:
### Description

Hello. At Bitso, we're using microcks a lot and recently we've found an issue installing microcks on some clusters because of the lack of affinity/tolerations support. 

This simple PR is adding a common affinity/toleration field to group the deployments.


install/kubernetes/microcks/README.md
```diff
@@ -166,7 +166,9 @@ The table below describes all the fields of the `values.yaml`, providing informa
...
+| `<root>`   | `commonAnnotations`         | **Optional**. Defines the list of annotations that will be added to all kubernetes resources (Default to `empty`)                                                                                                          
+| `<root>`   | `commonAffinities`              | **Optional**. Defines the affinities that will be added to all kubernetes deployments (Default to `empty`)                                                                                                                                                                                                          |
+| `<root>`   | `commonTolerations`         | **Optional**. Defines the list of tolerations that will be added to all kubernetes deployments (Default to `empty`)                                                                                                                                                                                                        |
```
lbroudoux, line 171: `common` actually means that it applies to all components - and that's what you did below: fine! However, I wonder if the same affinities and tolerations  could fit with every components needs. Don't you need toset specific constraints for Webapp, Keycloak or for MongoDB?

mvtavares: well, maybe it would make sense to have separate affinities/tolerations, but since we don't have any currently, i think defining at least the default one makes sense for all.

Latter, IMO we could improve to make possible to have different ones for each component.

mvtavares: Right now, all we need to do is to move all the components out from the default nodegroup

lbroudoux: Makes sense. Thanks for the answer so that we're tracking the context.

lbroudoux (approved):
LGTM.

lbroudoux:
Thanks again for this nice PR!

PR #1387, opened 2024-10-22
Add escape character on dynamic response

### LINKED ISSUE #1353: Add escape character on dynamic response
### Reason/Context

We have fields in API responses based on UUIDs between { }. I write the following line in the example in my reply: “myfield”: “{{{uuid()}}}” and I get “myfield”: “}” back. On the other hand, if I write: “myfield”: “{ {{uuid()}} }”, I get a UUID, but there's too much space between it and the braces. I try to put \ to escape if I paste to get the right format but it doesn't change much : “ myfield”: “\}”

### Description

Possibility to add escape \ for special caracters like { or } on response template.

Ex : “myfield”: “\{{{uuid()}}\}” will give “myfield”: "{82777902-E163-4DA3-A27C-EED07500DC92}" on response mock call.

### Implementation ideas

_No response_

Harsh4902:
### Description

- This PR adds support for escape character, which are provided in response example of specification file.
- This PR modifies `skipToCorrectEndSuffix()` method of `ExpressionParser` which now will now return index of last '}}'.
- I will enable us to wrap values inside the wrapper like `{}`, `()`, `##`, `[]` in response body.

### Behavior After Changes 

- Now Microcks will be able to parse escape characters which are provided in response example of specification file like given below:

```
{
   "id": "{{{uuid()}}}",
    "name": "{{ #request.body/name# }}"
}
```
- This will generate result like given below:
```
{
  "id": "{570e8c11-cac6-42e1-ab3a-6f4e6d9a10a2}",
  "name": "#Rusty#"
}
```

**it will support wrappers like '{}', '()', '[]' '##' only.

### Related issue(s)
Fixes #1353

### DIFF
--- a/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
+++ b/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
@@ -89,7 +89,12 @@ public class ExpressionParser {
                throw new ParseException(template, prefixIndex, "No expression defined within delimiter '"
                      + expressionPrefix + expressionSuffix + "' at character " + prefixIndex);
             }
-            expressions.add(doParseExpression(expr, context));
+            if (expr.charAt(0) == '{' || expr.charAt(0) == '(' || expr.charAt(0) == '[' || expr.charAt(0) == '#') {
+               expressions.add(new LiteralExpression(expr.substring(0, 1)));
+               expressions.add(doParseExpression(expr.substring(1, expr.length() - 1), context));
+               expressions.add(new LiteralExpression(expr.substring(expr.length() - 1)));
+            } else
+               expressions.add(doParseExpression(expr, context));
             startIdx = suffixIndex + expressionSuffix.length();
             log.debug("Expression accumulated. Pursuing with index {} on {}", startIdx, template.length());
          } else {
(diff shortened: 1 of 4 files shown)


lbroudoux:
Hey @Harsh4902, thanks for tackling this one! 
Are there any incompatibilities that would make this difficult to support with other kind of expression? Or is it just because it's a work in progress?

Harsh4902:
> Thinking about it, I wonder if there isn't an easier way to handle this case...
> 
> Let's say I want to introduce any extra character around the result of an expression - `#` for example. I would use something like: `"id": "#{{ uuid() }}#"` that would generate something like: `"id": "#570e8c11-cac6-42e1-ab3a-6f4e6d9a10a2#"`. That works well!
> 
> Now let's say I want the same thing with curly braces `{}`. The natural thing to do would be to use something like: `"id": "{{{ uuid() }}}"` that would generate something like: `"id": "{570e8c11-cac6-42e1-ab3a-6f4e6d9a10a2}"`.
> 
> So instead of dealing with escape characters like `\`, why not just considering the 2 opening braces on the right and the 2 closing braces on the left?
> 
> As of today, the algorithm is something like:
> 
> * Search for `{{` and `}}`, extract content at these exact positions to get expression
> * Evaluate expression and get result
> * Replace `{{ expression }}` by its value in the string => `value`
> 
> It would evolve to:
> 
> * Check for `{{` and `}}`, extract content between the 2 most right opening braces (it could be `{{` out of `{{{`) and the 3 most left closing braces (it could be `}}` out of `}}}`)
> * Evaluate expression and get result
> * Replace `{{{ expression }}}` by its value in the string => `{value}`
> 
> I think the approach is much more natural to use and generic and could be handled at the `ExpressionParser` level only. What do you think?

Yes, I was not sure about this solution. Actually, I want to discuss with you after implementation and that's why I just made the changes for FunctionExpression and raised draft PR.

Harsh4902:
@lbroudoux Sorry for the delay. I am on it, and will complete it ASAP.

Harsh4902:
@lbroudoux I am closing this PR and raising a new one.

PR #1404, opened 2024-11-06
fix: Respect operation `delay` when using the `PROXY` dispatcher.

kaos:
### Description

When using the `PROXY` (or `PROXY_FALLBACK`) dispatcher, any `delay` for the operation was ignored.

### Related issue(s)

None filed.

### Test results

Pending...

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/RestController.java
+++ b/webapp/src/main/java/io/github/microcks/web/RestController.java
@@ -309,9 +309,17 @@ public class RestController {
          response = getResponseByMediaType(responses, request);
       }
 
+      // Setting delay to default one if not set.
+      if (delay == null && ic.operation.getDefaultDelay() != null) {
+         delay = ic.operation.getDefaultDelay();
+      }
+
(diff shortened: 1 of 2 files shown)


kaos:
I did look for any tests regarding the delay option, but couldn't find any. Let me know if we want to add any tests, as I may need some initial guidance in that case.

lbroudoux:
The integration tests for that are implemented into `RestControllerIT` class. You can check the `testOpenAPIMockingWithValidation()` method I added recently for request payload validation and do similar things with delay.  There's also a `testProxyFallback()` method and you could add a `testProxyFallbackWithDelay()`?

lbroudoux (commented):
Looks nice! I just added a comment regarding a test case. Let me know if doable.

kaos:
OK, I've added a test case that introduces a 100ms delay, and verifies that the request time taken for the delayed version takes at least 50ms longer to process than those without delay.

This makes the test fail for me, if I revert the fix introduced in this PR, while it pass with the fix present.

lbroudoux (approved):
LGTM. The integration test could benefit from some refactoring to avoid duplication of assertions with other test or just to focus on the delay stuff. I'll take care of that! Thanks a lot for contributing this!

PR #1406, opened 2024-11-07
Add escape character on dynamic response

### LINKED ISSUE #1353: Add escape character on dynamic response

Harsh4902:
### Description

- This PR adds support for escape character, which are provided in response example of specification file.
- This PR modifies `skipToCorrectEndSuffix()` method of `ExpressionParser` which now will now return index of last '}}'.
- I will enable us to wrap values inside the wrapper like `{}`, `()`, `##`, `[]` in response body.

### Behavior After Changes 

- Now Microcks will be able to parse escape characters which are provided in response example of specification file like given below:

```
{
   "id": "{{{uuid()}}}",
    "name": "#{{ request.body/name }}#"
}
```
- This will generate result like given below:
```
{
  "id": "{570e8c11-cac6-42e1-ab3a-6f4e6d9a10a2}",
  "name": "#Rusty#"
}
```

**it will support wrappers like '{}', '()', '[]' '##' only.

### Related issue(s)
Fixes #1353

### DIFF
--- a/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
+++ b/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
@@ -88,7 +88,7 @@ public class ExpressionParser {
                throw new ParseException(template, prefixIndex, "No expression defined within delimiter '"
                      + expressionPrefix + expressionSuffix + "' at character " + prefixIndex);
             }
-            if (expr.charAt(0) == '{' || expr.charAt(0) == '(' || expr.charAt(0) == '[' || expr.charAt(0) == '#') {
+            if (expr.charAt(0) == '{') {
                expressions.add(new LiteralExpression(expr.substring(0, 1)));
                expressions.add(doParseExpression(expr.substring(1, expr.length() - 1), context));
                expressions.add(new LiteralExpression(expr.substring(expr.length() - 1)));
--- a/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
+++ b/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
@@ -88,7 +88,12 @@ public class ExpressionParser {
                throw new ParseException(template, prefixIndex, "No expression defined within delimiter '"
                      + expressionPrefix + expressionSuffix + "' at character " + prefixIndex);
             }
-            expressions.add(doParseExpression(expr, context));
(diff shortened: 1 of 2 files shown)


lbroudoux:
Thank you @Harsh4902!

The `"{{ #request.body/name# }}"` to `"#Rusty#" seems counter-intuitive to me...
I think that `"#{{ request.body/name }}#" already produces the same result. Maybe we don't have to explicitely manage those cases?

commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
```diff
@@ -88,7 +88,12 @@ public static Expression[] parseExpressions(String template, EvaluationContext c
                throw new ParseException(template, prefixIndex, "No expression defined within delimiter '"
                      + expressionPrefix + expressionSuffix + "' at character " + prefixIndex);
             }
-            expressions.add(doParseExpression(expr, context));
+            if (expr.charAt(0) == '{' || expr.charAt(0) == '(' || expr.charAt(0) == '[' || expr.charAt(0) == '#') {
```
lbroudoux, line 91: Looks good but I think we don't have to manage the `(`, `[` or `#` cases. These are already possible using standard notation. See my above comment.

Harsh4902: Oh! Yeah, valid point. Actually, I was thinking about everything inside `{{}}`. I removed explicit check now.

lbroudoux (changes requested):
Awesome! Looks nice. Wouldn't be possible a add a unit test for this? We already have a `ExpressionParseTest` class...

lbroudoux (approved):
Looks great! Thnak you very much for the PR!

PR #1425, opened 2024-11-28
feat: 1418 access gRPC metadata (headers) in SCRIPT dispatcher

### LINKED ISSUE #1418: Support headers in gRPC APIs
We recently started using Microcks for our gRPC APIs and really love the functionality and extensibility. 
One feature we are currently missing for gRPC APIs are headers. Since headers (or in gRPC lingo: gRPC Metadata) are used for authentication, custom headers, tracing or other internal usages, it would be great to support headers in mocks and especially in tests. 

I currently see the following points that would need to be addressed to fully support headers for gRPC APIs: 
- API Testing: Pass headers specified in gRPC Test Runs through to the gRPC call
- API Mocking:  Make headers available in `SCRIPT_DISPATCHER` (currently, it is always an empty list)

It would be great to see this feature in microcks! I am also happy to support here and work on a PR 🙂

anika-apel:
### Description

- Make gRPC Metadata available in SCRIPT dispatcher via `mockRequest.getRequestHeaders()` by passing headers explicitly into the `ScriptEngineBinder`
- Use a ServerInterceptor to pass Metadata to Server (`GrpcServerCallHandler`) via Context (see [grpc-java Example](https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/header/HeaderServerInterceptor.java))
- During the implementation I encountered a bug in `HttpHeadersStringToStringMap`. Adding a single value with `add(String key, String value)` was throwing an error since `get()` returns null if the value is an empty list (to correctly mock the behaviour of HttpHeaders)

### Related issue(s)

Partially resolves #1418


lbroudoux:
Wow! Awesome! I'll check it out and do a review asap. Thanks very much!

webapp/src/main/java/io/github/microcks/web/SoapController.java
```diff
@@ -393,7 +394,8 @@ private DispatchContext getDispatchCriteriaFromScriptEval(Service service, Strin
          // Evaluating request with script coming from operation dispatcher rules.
          ScriptEngine se = sem.getEngineByExtension("groovy");
          ScriptEngineBinder.bindEnvironment(se, body, requestContext,
-               new ServiceStateStore(serviceStateRepository, service.getId()), request);
+               new ServiceStateStore(serviceStateRepository, service.getId()),
+               HttpHeadersUtil.extractFromHttpServletRequest(request), request);
```
lbroudoux, line 398: See my above comment on `ScriptEngineBinder`. As we're still passing `request`, the binder is still knowledgeable about HTTP stuff, so I would let regular HTTP headers parsing its responsibility.

webapp/src/test/java/io/github/microcks/util/script/ScriptEngineBinderTest.java
```diff
@@ -47,7 +47,7 @@ void testRequestContentIsBound() {
       try {
          // Evaluating request with script coming from operation dispatcher rules.
          ScriptEngine se = sem.getEngineByExtension("groovy");
-         ScriptEngineBinder.bindEnvironment(se, body, null, null);
+         ScriptEngineBinder.bindEnvironment(se, body, null, null, null);
```
lbroudoux, line 50: See my above comment on `ScriptEngineBinder` to avoid changes here.

webapp/src/test/java/io/github/microcks/util/script/ScriptEngineBinderTest.java
```diff
@@ -95,7 +95,7 @@ void testRequestContextIsModified() {
       try {
          // Evaluating request with script coming from operation dispatcher rules.
          ScriptEngine se = sem.getEngineByExtension("groovy");
-         ScriptEngineBinder.bindEnvironment(se, body, context, null);
+         ScriptEngineBinder.bindEnvironment(se, body, context, null, null);
```
lbroudoux, line 98: See my above comment on `ScriptEngineBinder` to avoid changes here.

webapp/src/test/java/io/github/microcks/util/script/ScriptEngineBinderTest.java
```diff
@@ -192,7 +192,7 @@ void testMicrocksXmlHolder() {
...
```
lbroudoux, line 195: See my above comment on `ScriptEngineBinder` to avoid changes here.

webapp/src/test/java/io/github/microcks/util/script/ScriptEngineBinderTest.java
```diff
@@ -234,7 +234,7 @@ void testEviwareXmlHolder() {
...
```
lbroudoux, line 237: See my above comment on `ScriptEngineBinder` to avoid changes here.

webapp/src/main/java/io/github/microcks/util/grpc/GrpcMetadataUtil.java
```diff
@@ -0,0 +1,74 @@
...
+package io.github.microcks.util.grpc;
+
+import io.github.microcks.util.script.HttpHeadersStringToStringsMap;
+import io.github.microcks.util.script.StringToStringsMap;
+import io.grpc.CallOptions;
+import io.grpc.Context;
+import io.grpc.Metadata;
+
+import java.io.UnsupportedEncodingException;
+import java.util.*;
```
lbroudoux, line 25: Please avoid wildcard imports and replace with explicit ones.

webapp/src/main/java/io/github/microcks/util/grpc/GrpcMetadataUtil.java
```diff
@@ -0,0 +1,74 @@
...
+      for (String key : keys) {
+         // Depending on the suffix of the key we either extract binary values or
+         // ASCII string values.
+         if (key.endsWith("-bin")) {
+            // binary metadata entry
+            Metadata.Key<byte[]> metadataKey = Metadata.Key.of(key, Metadata.BINARY_BYTE_MARSHALLER);
+            Iterable<byte[]> values = metadata.getAll(metadataKey);
+
+            for (byte[] value : values) {
+               String stringValue = new String(value, "UTF-8");
```
lbroudoux, line 60: Please use StandardCharset.UTF_8

webapp/src/main/java/io/github/microcks/util/grpc/GrpcMetadataUtil.java
```diff
@@ -0,0 +1,74 @@
...
+    *                                      string.
+    */
+   public static StringToStringsMap convertToMap(Metadata metadata) throws UnsupportedEncodingException {
+      StringToStringsMap result = new HttpHeadersStringToStringsMap();
+
+      Set<String> keys = metadata.keys();
+      for (String key : keys) {
+         // Depending on the suffix of the key we either extract binary values or
+         // ASCII string values.
+         if (key.endsWith("-bin")) {
```
lbroudoux, line 54: Is it a common convention? Don't we have another way of telling how a key should be marshalled/unmarshalled? 
Plus: shouldn't we have at least a constant definition like `BINARY_VALUE_SUFFIX`?

webapp/src/test/java/io/github/microcks/util/grpc/GrpcMetadataUtilTest.java
```diff
@@ -0,0 +1,112 @@
...
+      }
+      StringToStringsMap expectedHeaders = new StringToStringsMap();
+      expectedHeaders.put("foo1", "bar");
+      assertEquals(expectedHeaders, headers);
+   }
+
+   @Test
+   void testMetadataWithBinaryMetadata() {
+      Metadata metadata = new Metadata();
+      metadata.put(Metadata.Key.of("foo1-bin", Metadata.BINARY_BYTE_MARSHALLER), "bar".getBytes());
```
lbroudoux, line 100: See my above comment on `GrpcMetadataUtil`. As we provide the marshaller when registering the key/value pair, wouldn't we be able to retrieve it later on when converting in a map and not having to rely on `*-bin` naming convention?

lbroudoux (changes requested):
Thanks a lot! Added some questions and change suggestions.

webapp/src/main/java/io/github/microcks/util/grpc/GrpcMetadataUtil.java
anika-apel: Yes, this is a common convention and actually required by grpc.
Here a snippet from the [grpc-specification](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md):

> Applications define binary headers by having their names end with "-bin". Runtime libraries use this suffix to detect binary headers and properly apply base64 encoding & decoding as headers are sent and received.


The constant was a great suggestion. Actually,`io.grpc.Metadata` defines already the constant BINARY_HEADER_SUFFIX which we can use.

When iterating over all keys in the metadata, we only get the name of the key and not the type of the key . There is one alternative, though. We could replace it by checking explicitly with which marshaller a key exists:
```

if (metadata.containsKey(Metadata.Key.of(key, Metadata.BINARY_BYTE_MARSHALLER) )) {
    // binary metadata
    ....
} else { 
    // ASCII string metadata
    ....
}
```

lbroudoux: Awesome!

lbroudoux (approved):

lbroudoux:
Looks all good! We are triggering the build and then will merge it if everything is okay. Great work!

lbroudoux:
👍 🚀  Thanks a lot for putting this together! That's a great addition!

PR #1438, opened 2024-12-05
#feat: 1418 support headers in grpc tests

### LINKED ISSUE #1418: Support headers in gRPC APIs

anika-apel:
### Description

- Add headers specified in Requests and Test Run to test invocations

### Related issue(s)

Resolves #1418 

### Notes
During the implementation, I noticed that all test runners share the same code to collect the headers and refactored it into one method in `TestRunnerCommons`. If that is not wanted, let me know and I can revert this change.

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
+++ b/webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
@@ -33,9 +33,6 @@ import com.google.protobuf.Descriptors;
 import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.util.JsonFormat;
-import com.google.rpc.Code;
-import com.google.rpc.Status;
-
 import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
@@ -48,7 +45,6 @@ import io.grpc.InsecureChannelCredentials;
 import io.grpc.ManagedChannel;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
-import io.grpc.StatusRuntimeException;
 import io.grpc.TlsChannelCredentials;
 import io.grpc.stub.ClientCalls;
 import org.slf4j.Logger;
@@ -63,7 +59,6 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
```diff
@@ -60,6 +70,36 @@
  */
 public class GrpcTestRunner extends AbstractTestRunner<HttpMethod> {
 
+   /* Call Option used to pass gRPC Metadata from client invocation to header client interceptor */
+   public static final CallOptions.Key<Metadata> METADATA_CUSTOM_CALL_OPTION = CallOptions.Key
+         .createWithDefault("request-metadata", null);
```
lbroudoux, line 75: Is `request-metadata` a standard stuff? Don't we have a Java constant for that in the `grpc-java` library?

webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
```diff
@@ -220,6 +269,17 @@ public void checkServerTrusted(X509Certificate[] certs, String authType) {
       return results;
    }
 
+   private static Metadata convertHeadersToMetadata(Set<Header> headers) {
+      Metadata metadata = new Metadata();
+      for (Header header : headers) {
+         log.info("Adding header {} to request", header.getName());
```
lbroudoux, line 275: Maybe `INFO` level is a bit high here as these are details. Could you lower this to `DEBUG`?

lbroudoux (changes requested):
It looks good! (great initiative moving headers collection to `TestRunnerCommons` BTW).
I see in the CI that build is failing due to integration tests on GRPC mocks. Are your investigating if it's a side effect?

webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
anika-apel: Absolutely. Additionally, it was an unnecessary logging message only meant for debugging purposes for me during development. I removed it completely

webapp/src/main/java/io/github/microcks/util/grpc/GrpcTestRunner.java
anika-apel: No there is no standard. Actually any string can be used as a name here. It should just describe the key. I moved it into a constant to highlight the intention of the name.

lbroudoux (approved):
Looks super good to me! Thanks!

PR #1479, opened 2025-01-29
fix: finish test run with details about failure if gRPC test suite fails

anika-apel:
### Description

- Running a gRPC test suite with failing tests currently leads to a `StatusRuntimeException` being thrown and the test run staying in progress forever
- PR adds logic to catch `StatusRuntimeException` for unsuccessful RPC calls (structured as in `HttpTestRunner`) and store failure results in Test Run
- Failing test run contains message with gRPC status code and error message

### Test Results

**Before**: 
Test run stays `in_progress` forever:

image

With error in logs: 

image

**After**: 

Test run finishes and is marked as `failed`. 
Detailed test results show error code and message. 

image


lbroudoux (approved):
It looks perfect! Thanks a lot for contributing this enhancement!

PR #1525, opened 2025-03-05
WIP: fix: #1505 Prevent user role from adding direct APIs

### LINKED ISSUE #1505: Read-only user role
Hi.

This is a question regarding the roles Microcks provides by default.

I have noticed users with the `user` role can create services via the **Add Direct API** button in the web UI.
At first, I thought this was a read-only role, but it seems I was wrong.

Is there any way to create a pure read-only role for UI users?

Thanks in advance.

Marckman:
### Description

- Hide Add Direct API button for users with the user role
- Forbid user access to POST: /services/generic in the backend

### Related issue(s)

Resolves #1505


lbroudoux (approved):
Looks good to me! Thanks for taking the time re-submitting the PR.

lbroudoux:
@Marckman  Thanks a lot for the contrib!

Issue #1480, opened 2025-01-29
Improving Microcks delivery and validation with GitHub Actions CI deployment tests


yada:
> [!NOTE]
> This issue reflects our community's decision to submit projects to the LFX mentorship program.
> See: https://github.com/orgs/microcks/discussions/1463
> [Project Idea 6](https://github.com/orgs/microcks/discussions/1463#discussioncomment-11891322) : Improving Microcks delivery and validation with GitHub Actions CI deployment tests

This project focuses on enhancing the reliability and quality of Microcks releases by introducing comprehensive CI deployment tests and validations using GitHub Actions. While the project already includes unit and integration tests, recent issues (ex: https://github.com/microcks/microcks/issues/1470 and https://bugs.openjdk.org/browse/JDK-8345296) with dependencies have underscored the need for end-to-end validation to ensure new integrations do not introduce bugs or regressions.

Participants will develop workflows for building and deploying Microcks and running automated tests to confirm its functionality under real-world scenarios. This will help prevent edge cases and dependency-related issues affecting the Microcks community and adopters.

Workflows encompass a bunch of deployment tests:

- Testing of container images (with docker, with podman, on different architectures, ...)
- Testing of common docker-compose configurations
- Testing of Helm chart with different setup options
- ...

This project improves the delivery process to ensure that Microcks releases meet the community's expectations for quality and reliability.

inosmeet:
Hi @yada!,

I'm interested in this project! I have experience working with GitHub Actions and CI/CD test workflows. I'd love to contribute to improving Microcks' delivery process through comprehensive deployment tests.

Are there any pre-tests or prerequisites I should be aware of before applying?

Looking forward to discussing this further!

lbroudoux:
Hey @inosmeet, that's awesome!

I think the first step would be to have a look at the main `build-verify.yml` GitHub actions workflow in the repository and share your thoughts about it. We have a lot of ideas about things we'd like to improve/do differently or add to it, but it's always nice to have a fresh and unbiased view first.

yada:
Hi @inosmeet ,

Thank you for your interest in this project; your help and contribution are greatly appreciated.

Please join as a mentee on the LFX mentorship program; this project is now online. For further details, see my latest comment in this thread: https://github.com/orgs/microcks/discussions/1463#discussioncomment-12067922

The direct link to this project is: https://mentorship.lfx.linuxfoundation.org/project/0c667baa-94bf-405c-ada6-c2bea3bf3e56
[...]

PR #1533, opened 2025-03-10
ci: add health checks for docker-compose installation

inosmeet:
I've added a bare-bone script that runs basic healthchecks for docker-compose installation method.
Since this might be a routine check, I've made this run once a week.

Tested using `act` locally.

cc @lbroudoux @yada


inosmeet:
I think we can use `docker compose` here, loosing the installation step for `docker-compose`. It's pre-installed in ubuntu image, just wasn't for act.

install/docker-compose/docker-compose.yml
```diff
@@ -18,10 +23,20 @@ services:
     volumes:
       - "./keycloak-realm/microcks-realm-sample.json:/opt/keycloak/data/import/microcks-realm.json"
     command: ["start-dev", "--hostname=http://localhost:18080", "--import-realm"]
+    healthcheck:
+      test: ["CMD", "sh", "-c", "echo > /dev/tcp/localhost/8080"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
```
lbroudoux, line 30: I think that Keycloak has its own HTTP-based helth check probes on a dedicated port. Must check the Helm chart for this.

install/docker-compose/docker-compose.yml
```diff
@@ -18,10 +23,20 @@ services:
...
 
   postman:
     image: quay.io/microcks/microcks-postman-runtime:0.6.0
     container_name: microcks-postman-runtime
+    healthcheck:
+      test: ["CMD", "sh", "-c", "echo > /dev/tcp/localhost/3000"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
```
lbroudoux, line 39: We have integrated probes in Microcks Postman Runtime that are already used on the Kubernetes deployment. Ideally, we must converge and use the same ones.

install/docker-compose/docker-compose.yml
inosmeet: cc @lbroudoux

inosmeet:
Seems like I just forgot to add the flag that enabled the health check, should work as expected now.

lbroudoux:
Though this looks like nice improvement from the docker-compose installation POV itself, I don't think checking things this way in the context of our test suite is the way to go. Those are interesting but too low levels to me and moreover we should trick stuff to match intervals and delays of the probe itself... Typically, I would drop stuff related to the `scripts` and rename the PR.

inosmeet:
Let's hold this off for now until I come up with a generalized way for doing this.

inosmeet:
with this, health-checks for each docker-compose has been added!

lbroudoux (approved):
LGTM with the removal of Keycloak health port in proxy configuration.

For better tracking, this is a Part of #1480

yada (approved):
LGTM

PR #1546, opened 2025-03-21
feat: add script for helm installation

inosmeet:
### Description
This script attempts streamline the installation process for helm with minikube.
If it's agreed that this approach is feasible, I'll add option for async feature.

cc @lbroudoux

### DIFF
--- a/install/docker-compose/docker-compose-async-addon.yml
+++ b/install/docker-compose/docker-compose-async-addon.yml
@@ -1,5 +1,3 @@
-version: '2'
-
 services:
   zookeeper:
     image: quay.io/strimzi/kafka:0.43.0-kafka-3.8.0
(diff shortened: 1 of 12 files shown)


lbroudoux (changes requested):
Looks nice but as it is not a general purpose script install, I think it would be better to put this in a folder dedicated to the test suites. I would this under a global `/tests` or `/testsuite` folder to be discuss. 

Also don't forget that running this in CI, also means managing download and install of minikube on GitHub runner too.

inosmeet:
It makes sense to maybe put this one into a `test` or some similar directory, but what about [docker|podman] compose? As discussed in the weekly meeting, we do aim to have installation script for each installation method, where do we put those?

inosmeet:
I wonder if we should just move this to `install/` directory? One script for each installation method or a single one for all perhaps?
lmk what you think @lbroudoux

lbroudoux:
> I wonder if we should just move this to `install/` directory? One script for each installation method or a single one for all perhaps? lmk what you think @lbroudoux

I think the same comment in https://github.com/microcks/microcks/pull/1552#issuecomment-2753578688 applies here. I would prefer to move this script into a test-related folder to reduce its scope. Also renaming it to something like `install-minikube-microcks.sh` would be more appropriate.

lbroudoux (approved):
LGTM

PR #1551, opened 2025-03-25
feat: add healthcheck for podman-compose installation

inosmeet:
### Description

Add healthchecks for all podman compose files


lbroudoux:
For better tracking, this is a Part of #1480

lbroudoux (approved):
Thank you! Looks all good to me!

PR #1552, opened 2025-03-25
feat: add installation script for [podman|docker]-compose

inosmeet:
### Description
This script allows us to install microcks using docker-compose method along with it's several addons and modes.
- We can specify modes using `--mode` flag [default | devmode | proxy]
- addons are specified in comma separated list using `--addons` flag.

I've made some questionable changes here. Namely renaming the `docker-compose-async-addon.yml` to `async-addon.yml` to for consistency. This allows us to make the script smart, however by doing this, we do need to update many documentation instances, so this may not be appropriate.

Lastly, if this one is approved then it can be incorporated to handle podman-compose installation as well. And again, file names will need to be updated.

As for the location of the script, it needs to be discussed further.
Part of #1480 
cc @lbroudoux


lbroudoux:
Hey, thanks for the proposals.

Problems with bash scripts are that they also need to be tested and - if publicly delivered - written in an OS-independent manner (there are sometimes subtle differences between Linux and Darwin bash and zsh and ...). That's why we prefer to use and document the usage of standard `docker compose` commands instead of custom scripts (it also allows users to better understand what they're actually executing). We made an exception by introducing `run-microcks.sh` in the Podman installation because that couldn't be avoided at a time when launching Podman was a mess on MacOS - that said, given yesterday's test this can certainly be removed in the future.

As a consequence, I would prefer to keep the `install-docker-compose.sh` script "private" (not really "private" as we're working in open 😉 ) or scoped to the `tests` or `testsuite` purpose only. That way it will be easier to maintain it on our target platform only (the GitHub runners we will select to run the CI). It could serve also as an inspiration to people who would like to automate stuff but it offloads us from the burden of making something industrial.

+1 for renaming the addons file for better consistency! We'll synchronize the documentation upgrade too.

So to sum up, I think the next step is to decide the folder where to put the scripts, configurations and functional tests you're describing. I see two obvious options at the moment:
* `/tests` - basic but can bring confusion with unit tests we store in `/src/test`
* `/testsuite`- IMHO sounds more elaborate and different from unit tests 

Personally, my preference goes to `/testsuite`. 

Any other suggestion is welcomed!

lbroudoux (approved):
Looks all good to me! Thanks,

lbroudoux:
Looks like the rebase to `1.12.x` involved some conflicts. Can you fix them so that I can merge the results?

lbroudoux (approved):
All the conflicts are now resolved. Go go go 🚀

PR #1564, opened 2025-04-06
feat: add workflow file for healthchecks

inosmeet:
### Description
Related: #1480 
Depends on: #1552 

This script mimics the timings specified in the healthchecks, but it results in some of the containers still being in the "starting" phase and the workflow run fails. Should we generalize timeouts?

This script will incorporate the remaining configurations of docker compose as well as the podman compose and helm installation incrementally.

cc @lbroudoux


lbroudoux (approved):
LGTM

lbroudoux:
Looks like changing the base to `1.12.x` raises some conflicts. Can you resolve them so that I can merge the results? Thanks,

PR #1567, opened 2025-04-08
feat: add installtion script for microcks

inosmeet:
### Description
Part of: #1480 

This script allows us to install _all_ microcks images along with different tags.
Testing this with `microcks-uber-async-minion` image fails. Seems like we hit regression just like before.
cc @lbroudoux


lbroudoux (approved):
LGTM

PR #1577, opened 2025-04-16
feat: add tests for sample APIs

inosmeet:
Part of #1480

### DIFF
--- a/testsuite/commons.js
+++ b/testsuite/commons.js
@@ -216,7 +216,7 @@ export function invokeREST_HelloAPIMocks() {
     const TEST_CASES = [
       { name: 'David', expStatus: 200, expGreeting: 'Hello David !' },
       { name: 'Gavin', expStatus: 200, expGreeting: 'Hello Gavin !' },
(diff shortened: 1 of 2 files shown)


testsuite/commons.js
```diff
@@ -173,3 +179,115 @@ export function invokeSOAPMocks() {
...
+  group('Hello API REST Mocks', () => {
+    const MOCK_NAME = 'Hello%20API%20Mock';
+    const VERSION   = '0.8';
+    const RESOURCE  = 'v1/hello';
+
+    const TEST_CASES = [
+      { name: 'David', expStatus: 200, expGreeting: 'Hello David !' },
+      { name: 'Gavin', expStatus: 200, expGreeting: 'Hello Gavin !' },
+      { name: 'Nobody', expStatus: 405, expGreeting: null          },
+    ];
```
inosmeet, line 220: In the last case `400` is being returned, but config file specifies `405` I'm not sure why?

inosmeet:
cc @lbroudoux

Issue #1584, opened 2025-04-22
UI Bug: Filter Input Field Breaks Alignment When Dropdown Opens


ChayanDass:
### Describe the bug

In the `pfng-filter-fields` component, when the filter field dropdown (`Search Field ▼`) is opened, the layout breaks — instead of keeping the input field inline, the input shifts to a new line below the dropdown.

### Actual behavior

![Image](https://github.com/user-attachments/assets/baa8eb1c-4ae0-4822-959f-7f9190cf5147)

### Microcks version or git rev

version : `1.12.x`

### Install method (`docker-compose`, `helm chart`, `operator`, `docker-desktop extension`,...)

docker compose

lbroudoux:
Yes, this appears during the Angular19 migration (see #1555). I don't know how to fix this at the moment.

ChayanDass:
I think ,I found the issue — when the dropdown opens, a .show class is added. This wasn't present in the previous version of Angular. The .show class doesn't serve any purpose in this context, but it's affecting the layout. So, the best solution is to simply override it.

![Image](https://github.com/user-attachments/assets/198f4cd7-ac44-40c5-8ca5-bef801528f4e)
![Image](https://github.com/user-attachments/assets/0d6a6640-7c38-463d-8283-ce6d0d499883)

lbroudoux:
Fix with #1586. Many thanks @ChayanDass!

PR #1586, opened 2025-04-23
fix: ui alignment for dropdown

ChayanDass:
This PR resolves the UI misalignment issue caused by the unnecessary addition of the `show` class when the dropdown is opened. The `show` class was overriding the default styling (`display: table-cell`) applied to the `.input-group-btn` component, resulting in a broken layout with child elements appearing on multiple lines.

Screenshot from 2025-04-23 21-02-56

Issue #1584

### DIFF
--- a/webapp/src/main/webapp/src/lib/patternfly-ng.css
+++ b/webapp/src/main/webapp/src/lib/patternfly-ng.css
@@ -787,9 +787,6 @@
   margin-right: 5px;
 }
 /* stylelint-disable */
-.input-group-btn{
-  display: table-cell !important;
-}
 .input-group .input-group-btn .dropdown-menu > .selected > a {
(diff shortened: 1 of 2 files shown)


ChayanDass:
#1854

lbroudoux (changes requested):
Awesome! Thanks for taking the time to investigate this!

As the `src/lib/patternfly-ng.css` is an integrated vendor file, it would be better not to touch it. Is it possible to move this directive to `src/style.css` instead, in one "patternfly override" sections (lines 229-303)? I think it will be easier to maintain and to remind ourselves that we changed this.

Thanks a lot!

ChayanDass:
done!

lbroudoux (approved):
LGTM. Thanks!

PR #1617, opened 2025-05-22
feat: enable health-checks for helm in workflow

inosmeet:
Part of: #1480 

cc @lbroudoux


lbroudoux (approved):
LGTM. Thanks a lot for this contrib!

PR #1649, opened 2025-07-04
Fix: AsyncAPI tests cannot handle contentType `any`

### LINKED ISSUE #1561: Tests cannot handle `string` message and throw `Message content cannot be parsed as JSON`
yeah, sorry, me again.

<details>
<summary>full asyncapi document</summary>

```yml
asyncapi: 3.0.0
info:
  title: Hoppscotch Echo WebSocket
  version: 1.0.0
  description: >
    Undestand how to use Hoppscotch Echo WebSocket as a client.
    Hoppscotch Echo WebSocket server echoes back any messages sent to it. You can use this to test WebSocket connections and message flows.

servers:
  echoServer:
    host: echo-websocket.hoppscotch.io
    protocol: wss

channels:
  echo:
    description: >
      The single channel where messages are sent to and echoed back. Echo server also regularly drops a timestampl to that channel.
    address: /
    messages:
      echo:
        $ref: '#/components/messages/echo'
      timestamp:
        $ref: '#/components/messages/timestamp'
    bindings:
      ws:
        method: POST

operations:
  handleTimeStampMessage:
    action: receive
    channel: 
      $ref: '#/channels/echo'
    summary: Receive the timestamp message sent from server every second.
    messages:
      - $ref: '#/channels/echo/messages/timestamp'
  sendEchoMessage:
    action: send
    channel: 
      $ref: '#/channels/echo'
    summary: Send a message to the echo server.
    messages:
      - $ref: '#/channels/echo/messages/echo'

components:
  messages:
    timestamp:
      summary: A message sent from server with current timestamp.
      payload: 
        $ref: '#/components/schemas/currentTime'
      examples:
        - name: someRandomDate
          payload: 11:13:24 GMT+0000 (Coordinated Universal Time)
    echo:
      summary: A message exchanged with the echo server.
      payload: {}
      examples:
        - name: string
          payload: test
        - name: object
          payload: 
            test: test text
  
  schemas:
    currentTime:
      type: string
      description: A timestamp with GMT timezone.
      example: 11:13:24 GMT+0000 (Coordinated Universal Time)
```

</details

### Troubleshooting

so my `sendEchoMessage` operation sends `echo` message that is basically `any`

```
    echo:
      summary: A message exchanged with the echo server.
      payload: {}
      examples:
        - name: string
          payload: test
        - name: object
          payload: 
            test: test text
```

in JSON Schema to specify that your type can be anything, you just do `{}`. To double check if Microcks respects that, I also did below:

```
    echo:
      summary: A message exchanged with the echo server.
      payload:
        type:
         - object
         - array
         - string
         - number
         - boolean
         - "null"
      examples:
        - name: string
          payload: test
        - name: object
          payload: 
            test: test text
```

So my examples cover `string` and `object`, and the `string` one always fails with `Message content cannot be parsed as JSON` 

even if I extend examples with:
```
      examples:
        - name: string
          payload: "test"
        - name: boolean
          payload: true
        - name: number
          payload: 123
        - name: object
          payload: 
            test: test text
```

all tests pass 🟢 except for the string one.

Image

Image

I think I tried all the possible options but none worked, so I have to bother you, sorry

### Doubts and workarounds

is it all happening because of what I see in the UI, that `application/json` content type is applied? 

but when on `echo` message I add `contentType: text/plain`:
```
echo:
      summary: A message exchanged with the echo server.
      contentType: text/plain
      payload: {}
      examples:
        - name: string
          payload: "test"
        - name: boolean
          payload: true
        - name: number
          payload: 123
        - name: object
          payload: 
            test: test text
```

it is not respected and it still tries to parse as json and `application/json` is expected.

the only working fix is when I add `defaultContentType: text/plain` to the root of the document.

so looks like Microcks does not respect `contentType` prop on a message level.

Harsh4902:
### Description
- This PR adds support for handling `contentType : <any>` at message level for AsyncAPI tests. 
- Previously, it was expecting `application/json`.

### What was the issue?
- `io.github.microcks.minion.async.AsyncAPITestManager.AsyncAPITestThread.getExpectedContentType()` was not able to determine message level content type because we were not following `$ref` recursively to identify the actual message node. 
- That's why it was returning `application/json` as content type because it's a default one. 

### Related issue(s)
Fixes #1561

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
@@ -296,11 +296,6 @@ public class AsyncAPITestManager {
          if (messageNode.has("contentType")) {
             contentType = messageNode.path("contentType").asText();
          }
-         if (messageNode.has("payload")) {
-            String payload = messageNode.path("payload").asText();
-            if (payload.isEmpty() || payload.equals("null"))
-               contentType = "any";
-         }
          return contentType;
       }
 
--- a/minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
@@ -281,14 +281,25 @@ public class AsyncAPITestManager {
          // Get message real content type if defined.
          String contentType = defaultContentType;
          JsonNode messageNode = specificationNode.at(messagePathPointer);
+
+         // messageNode will be an array of messages
+         if (messageNode.isArray() && messageNode.size() > 0) {
+            messageNode = messageNode.get(0);
+         }
+
          // If it's a $ref, then navigate to it.
-         if (messageNode.has("$ref")) {
+         while (messageNode.has("$ref")) {
             // $ref: '#/components/messages/lightMeasured'
             String ref = messageNode.path("$ref").asText();
-            messageNode = messageNode.at(ref.substring(1));
+            messageNode = specificationNode.at(ref.substring(1));
          }
          if (messageNode.has("contentType")) {
             contentType = messageNode.path("contentType").asText();
+         }
+         if (messageNode.has("payload")) {
+            String payload = messageNode.path("payload").asText();
+            if (payload.isEmpty() || payload.equals("null"))
+               contentType = "any";
          }
          return contentType;
       }
--- a/minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java
+++ b/minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java
@@ -32,6 +32,8 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.junit.jupiter.Container;
@@ -46,6 +48,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
 
 import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 
@@ -58,13 +61,16 @@ import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 class AsyncAPITestManagerIT {
 
    private static final String TOPIC_NAME = "test-topic";
-   private static final String TEXT_MESSAGE_TEMPLATE = """
-         {
-            "fullName": "Laurent Broudoux",
-            "email": "laurent@microcks.io",
-            "age": 4%s
-         }
-         """;
+
+   private static Stream<TestConfig> testConfigs() {
+      return Stream.of(new TestConfig("user-signedup-asyncapi-3.0.yaml", "SEND publishUserSignedUps", """
+            {
+               "fullName": "Laurent Broudoux",
+               "email": "laurent@microcks.io",
+               "age": 4%s
+            }
+            """), new TestConfig("send-message-asyncapi-3.0.yaml", "SEND sendEchoMessage", "test"));
+   }
 
    private static final Network NETWORK = Network.newNetwork();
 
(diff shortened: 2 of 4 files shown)


Harsh4902:
@lbroudoux Can you please suggest how can I introduce an Integration test for this?

lbroudoux:
> @lbroudoux Can you please suggest how can I introduce an Integration test for this?

Have you checked https://github.com/microcks/microcks/blob/master/minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java ?

minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
```diff
@@ -281,15 +281,26 @@ private String getExpectedContentType(JsonNode specificationNode, String message
...
+            messageNode = messageNode.get(0);
+         }
+
          // If it's a $ref, then navigate to it.
-         if (messageNode.has("$ref")) {
+         while (messageNode.has("$ref")) {
             // $ref: '#/components/messages/lightMeasured'
             String ref = messageNode.path("$ref").asText();
-            messageNode = messageNode.at(ref.substring(1));
+            messageNode = specificationNode.at(ref.substring(1));
```
lbroudoux, line 294: Oops... nice fix. I think it was a bug.

minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
```diff
@@ -281,15 +281,26 @@ private String getExpectedContentType(JsonNode specificationNode, String message
          // Get message real content type if defined.
          String contentType = defaultContentType;
          JsonNode messageNode = specificationNode.at(messagePathPointer);
+
+         // messageNode will be an array of messages
+         if (messageNode.isArray() && messageNode.size() > 0) {
+            messageNode = messageNode.get(0);
+         }
```
lbroudoux, line 288: I think that in AsyncAPI 2.x a message can be a single element and not an array. Can we check this and also add a unit tests for that part?

minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
```diff
@@ -281,15 +281,26 @@ private String getExpectedContentType(JsonNode specificationNode, String message
...
+         while (messageNode.has("$ref")) {
             // $ref: '#/components/messages/lightMeasured'
             String ref = messageNode.path("$ref").asText();
-            messageNode = messageNode.at(ref.substring(1));
+            messageNode = specificationNode.at(ref.substring(1));
          }
          if (messageNode.has("contentType")) {
             contentType = messageNode.path("contentType").asText();
          }
+         if (messageNode.has("payload")) {
```
lbroudoux: IIRC `payload` is mandatory in both AsyncAPI 2.x and 3.x. I think this is not necessary. Remember that spec has been correctly parsed before during the import, so basic stuff like this shouldn't happen.

minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
Harsh4902: Thanks :)

minions/async/src/main/java/io/github/microcks/minion/async/AsyncAPITestManager.java
Harsh4902: I have added unit test for this.

minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java
```diff
@@ -166,4 +166,94 @@ private static void sendTextMessagesOnTopic(int number) {
...
+            }
+            return Collections.emptyList();
+         }
+
+         @Override
+         public TestCaseResult reportTestCaseResult(String testResultId, TestCaseReturnDTO testCaseReturn) {
+            reportedTestCases.put(testResultId, testCaseReturn);
+            return null;
+         }
+      };
```
lbroudoux: Maybe we could extract the anonymous implementation present in different tests to a named class to avoid duplication?

minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java
```diff
@@ -166,4 +166,94 @@ private static void sendTextMessagesOnTopic(int number) {
...
+
+      KafkaProducer<String, String> producer = new KafkaProducer<>(props);
+
+      for (int i = 0; i < 5; i++) {
+         ProducerRecord<String, String> kafkaRecord = new ProducerRecord<>(TOPIC_NAME,
+               String.valueOf(System.currentTimeMillis()), "test");
+         producer.send(kafkaRecord);
+      }
+      producer.flush();
+      producer.close();
```
lbroudoux: Isn't it a duplication of the `sendTextMessagesOnTopic()` private method? Not really the same as you're just sending `test` string but... I don't get the point.

minions/async/src/test/java/io/github/microcks/minion/async/AsyncAPITestManagerIT.java
```diff
@@ -166,4 +166,94 @@ private static void sendTextMessagesOnTopic(int number) {
       producer.flush();
       producer.close();
    }
+
+   @Test
+   void testKafkaSendMessageTestSuccess() throws Exception {
```
lbroudoux: I don't get the naming and why it's a `KafkaSendMessage` test? It looks super similar to the `testKafkaAsyncAPITestSuccess()` above. Couldn't have we done this with a `ParametrizedTest`?

lbroudoux (commented):
Hey @Harsh4902 
LGTM but can you check if the test addition in `AsyncAPITestManagerIT` can be don using a `Parametrized JUnit Test`?

Harsh4902:
@lbroudoux I improved the code of `AsyncAPITestManagerIT`. Introduced a record `TestConfig` which will hold the spec name, operation name and message. Used `ParameterizedTest` and passed a stream of `TestConfig`.

Issue #1543, opened 2025-03-14
Improve `SCRIPT` dispatcher performance enabling caching


lbroudoux:
### Reason/Context

As of today, Groovy `SCRIPT` executions are not cached. Well, actually, invocations should not be cached but the script compilation (from a String to Java Bytecode) is a costly operation that can certainly benefit from caching.

### Description

Today, the `ScriptEngineManager` and the `ScriptEngine` that allows further execution are re-created for each mock invocation:

```java
ScriptEngineManager sem = new ScriptEngineManager();
ScriptEngine se = sem.getEngineByExtension("groovy");
```

Doing so makes the creation of a new `GroovyScriptEngineImpl` object that is expected to cache compiled scripts but cannot do that as it is a short-live object that is reclaimed by Garbage Collection after the mock invocation. 

The goal of this issue would be to make it possible to reuse the same instance of `GroovyScriptEngineImpl` so that script compilation will actually be cached.

### Implementation ideas

As a consequence of the above mutualization of `GroovyScriptEngineImpl`, we'll certainly have to review the way we create and manage bindings and context that are passed to the script. Today's implementation is thread-safe as those objects are also short-living objects but we have to be sure that the context and bindings that are passed to a shared ScriptEngine are actually dropped after a script execution to avoid collision or race conditions.

lbroudoux:
Here's the methodology I have applied to get baseline measures before going further.

The experiments are done on my development machine: a MacBook M2 Max, edition 2022 with 32GB RAM.
I will use the `Pastry API` as exposed in the tutorial: https://microcks.io/documentation/tutorials/getting-started/ and the [`siege`](https://github.com/JoeDog/siege) tool  during 30 seconds shots to get measures.

### Reference

First, I try to get the raw performance on the `GET /pastry` operation that uses no dispatcher at all.
I'm running `siege` in benchmark mode (`-b`) simulating 50 concurrent users (`-c 50`). Here's the command I issued and the results I got:

```sh
$ siege -b -c 50 http://localhost:8080/rest/API+Pastry+-+2.0/2.0.0/pastry
# 30 seconds later...
Lifting the server siege...
Transactions:          6150 hits
Availability:        100.00 %
Elapsed time:         30.42 secs
Data transferred:         2.22 MB
Response time:          0.25 secs
Transaction rate:       202.17 trans/sec
Throughput:          0.07 MB/sec
Concurrency:         49.81
Successful transactions:        6150
Failed transactions:            0
[...]

PR #1656, opened 2025-07-11
feat: Quickjs4J demo integratio

andreaTP:
### Description

This is a POC to demonstrate one possible approach to have a JavaScript engine to evaluate the script in dispatch.
The current implementation is for demo purposes, early questions:
- do we expect people to(only) upload little snippets? In Quickjs4J we invested in the DX to provide the script [as a "library"](https://github.com/roastedroot/quickjs4j/tree/main/core/src/test/resources/)
- I added a `DispatchStyle` but I'm unsure it's a good idea, any preference on how to roll this forward?
- is the proposed `JsApi` reasonable? any specific preference for accessing fields etc.? or high level guidance like: "should be as familiar as possible to JS/TS devs"
- the current script doesn't have access to FileSystem/Network and similar, so, it fall short on calling external APIs, do you want to make it possible dynamically in the script? Would it be ok to provide a bespoke HTTP invocation API?
- ... I have more questions, but let's iterate :slightly_smiling_face: 

Happy to hear feedback!

### Related issue(s)

See also #1559

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
+++ b/webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
@@ -1,0 +1,194 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.util.script;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import io.github.microcks.service.StateStore;
+import io.github.microcks.util.http.HttpHeadersUtil;
+import io.roastedroot.quickjs4j.annotations.Builtins;
+import io.roastedroot.quickjs4j.annotations.GuestFunction;
+import io.roastedroot.quickjs4j.annotations.HostFunction;
+import io.roastedroot.quickjs4j.annotations.Invokables;
+import io.roastedroot.quickjs4j.core.Engine;
+import jakarta.servlet.http.HttpServletRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+/**
+ * Utility class that holds methods for creating binding environments and evaluation context for a JSR 233 ScriptEngine.
+ * @author laurent
+ */
+public class JsScriptEngineBinder {
+
+   /** A simple logger for diagnostic messages. */
+   private static final Logger log = LoggerFactory.getLogger(JsScriptEngineBinder.class);
+
+   /** Private constructor to hide the implicit public one. */
+   private JsScriptEngineBinder() {
+   }
+
+   public static Engine buildEvaluationContext(String requestContent, Map<String, Object> requestContext,
+         StateStore stateStore, HttpServletRequest request) {
+      StringToStringsMap headers = HttpHeadersUtil.extractFromHttpServletRequest(request);
+      return buildEvaluationContext(requestContent, requestContext, stateStore, headers, request);
+   }
+
+   public static Engine buildEvaluationContext(String requestContent, Map<String, Object> requestContext,
+         StateStore stateStore, HttpServletRequest request, Map<String, String> uriParameters) {
+      StringToStringsMap headers = HttpHeadersUtil.extractFromHttpServletRequest(request);
+      return buildEvaluationContext(requestContent, requestContext, stateStore, headers, request, uriParameters);
+   }
+
+   public static Engine buildEvaluationContext(String requestContent, Map<String, Object> requestContext,
+         StateStore stateStore, StringToStringsMap headers, HttpServletRequest request) {
+      return buildEvaluationContext(requestContent, requestContext, stateStore, headers, request, null);
+   }
+
+   private static final ObjectMapper mapper = new ObjectMapper();
+
+   @Builtins("log")
+   public static final class LogContext {
+      @HostFunction
+      public void info(String str) {
+         log.info(str);
+      }
+
+      @HostFunction
+      public void debug(String str) {
+         log.debug(str);
+      }
+
+      @HostFunction
+      public void warn(String str) {
+         log.warn(str);
+      }
+
+      @HostFunction
+      public void error(String str) {
+         log.error(str);
+      }
+   }
+
+   @Builtins("store")
+   public final static class StoreApi {
+      private final StateStore delegate;
+
+      public StoreApi(StateStore store) {
+         this.delegate = store;
+      }
+
+      @HostFunction
+      public String get(String key) {
+         return this.delegate.get(key);
+      }
+
+      @HostFunction
+      public void put(String key, String value) {
+         this.delegate.put(key, value);
+      }
+
+      @HostFunction
+      public void delete(String key) {
+         this.delegate.delete(key);
+      }
+   }
+
+   @Builtins("requestContext")
+   public final static class RequestContextApi {
+      private final Map<String, Object> delegate;
+
+
+      public RequestContextApi(Map<String, Object> requestContext) {
+         this.delegate = requestContext;
+      }
+
+      @HostFunction
+      public JsonNode get(String key) {
+         return mapper.convertValue(delegate.get(key), JsonNode.class);
+      }
+
+      @HostFunction
+      public void set(String key, Object value) {
+         delegate.put(key, value);
+      }
+   }
+
+   @Builtins("mockRequest")
+   public final static class MockRequestApi {
+      private final FakeScriptMockRequest delegate;
+      private final ObjectMapper mapper = new ObjectMapper();
+
+      public MockRequestApi(FakeScriptMockRequest mockRequest) {
+         this.delegate = mockRequest;
+      }
+
+      @HostFunction
+      public String requestContent() {
+         return this.delegate.getRequestContent();
+      }
+
+      @HostFunction
+      public JsonNode getRequestHeader(String key) {
+         ArrayNode arr = mapper.createArrayNode();
+         StringToStringsMap reqHeaders = this.delegate.getRequestHeaders();
+         for (String str : reqHeaders.get(key)) {
+            arr.add(str);
+         }
+         return arr;
+      }
+
+      @HostFunction
+      public String getURIParameter(String key) {
+         if (this.delegate.getURIParameters() == null) {
+            return null;
+         }
+         return this.delegate.getURIParameters().get(key);
+      }
+   }
+
+   @Invokables("js")
+   public interface JsApi {
+      @GuestFunction
+      String process();
+   }
+
+   public static Engine buildEvaluationContext(String requestContent, Map<String, Object> requestContext,
+         StateStore stateStore, StringToStringsMap headers, HttpServletRequest request,
+         Map<String, String> uriParameters) {
+
+      // Build a fake request container.
+      FakeScriptMockRequest mockRequest = new FakeScriptMockRequest(requestContent, headers);
+      mockRequest.setRequest(request);
+      mockRequest.setURIParameters(uriParameters);
+
+      // Create bindings and put content according to SoapUI binding environment.
+      Engine engine = Engine.builder().addInvokables(JsApi_Invokables.toInvokables())
+            .addBuiltins(LogContext_Builtins.toBuiltins(new LogContext()))
+            .addBuiltins(StoreApi_Builtins.toBuiltins(new StoreApi(stateStore)))
+            .addBuiltins(RequestContextApi_Builtins.toBuiltins(new RequestContextApi(requestContext)))
+            .addBuiltins(MockRequestApi_Builtins.toBuiltins(new MockRequestApi(mockRequest))).build();
+
+      return engine;
+   }
+
+   public static String wrapIntoFunction(String script) {
+      return "function process() { " + script + "}";
+   }
+}
--- a/webapp/src/main/webapp/angular.json
+++ b/webapp/src/main/webapp/angular.json
@@ -108,5 +108,8 @@
         }
       }
     }
+  },
+  "cli": {
+    "analytics": false
   }
 }
--- a/webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
+++ b/webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.util.script;
+
+import io.github.microcks.service.StateStore;
+import io.roastedroot.quickjs4j.core.Engine;
+import io.roastedroot.quickjs4j.core.Runner;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * This is a test case for class JsScriptEngineBinder class.
+ */
+class JsScriptEngineBinderTest {
+
+   @Test
+   void testRequestContentIsBound() {
+      String script = JsScriptEngineBinder.wrapIntoFunction("""
+            return mockRequest.requestContent();
+            """);
+
+      String body = "content";
+
+      try {
+         // Evaluating request with script coming from operation dispatcher rules.
+         Engine engine = JsScriptEngineBinder.buildEvaluationContext(body, null, null, null);
+         Runner runner = Runner.builder().withEngine(engine).build();
+         JsScriptEngineBinder.JsApi jsApi = JsApi_Invokables.create(script, runner);
+
+         String result = jsApi.process();
+
+         assertEquals(body, result);
+      } catch (Exception e) {
+         fail("Exception should no be thrown");
+      }
+   }
+
+   @Test
+   void testRequestContentHeadersAreBound() {
+      String script = JsScriptEngineBinder.wrapIntoFunction("""
+            const fooHeader = mockRequest.getRequestHeader("foo");
+            log.info("header: " + fooHeader[0]);
+            return fooHeader[0];
+            """);
+
+      String body = "content";
+      MockHttpServletRequest request = new MockHttpServletRequest();
+      request.addHeader("foo", "bar");
+
+      try {
+         // Evaluating request with script coming from operation dispatcher rules.
+         Engine engine = JsScriptEngineBinder.buildEvaluationContext(body, null, null, request);
+         Runner runner = Runner.builder().withEngine(engine).build();
+         JsScriptEngineBinder.JsApi jsApi = JsApi_Invokables.create(script, runner);
+
+         String result = jsApi.process();
+
+         assertEquals("bar", result);
+      } catch (Exception e) {
+         fail("Exception should no be thrown");
+      }
+   }
+
+   @Test
+   void testUriParametersAreBound() {
+      String script = JsScriptEngineBinder.wrapIntoFunction("""
+            const fooParam = mockRequest.getURIParameter("foo");
+            log.info("uri parameter: " + fooParam);
+            return fooParam ?? "null";
+            """);
+
+      String body = "content";
+      MockHttpServletRequest request = new MockHttpServletRequest();
+      Map<String, String> uriParameters = new HashMap<>();
+      uriParameters.put("foo", "bar");
+
+      try {
+         Engine engine = JsScriptEngineBinder.buildEvaluationContext(body, null, null, request, uriParameters);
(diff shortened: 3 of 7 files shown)


webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -0,0 +1,194 @@
...
+
+      @HostFunction
+      public JsonNode get(String key) {
+         return mapper.convertValue(delegate.get(key), JsonNode.class);
+      }
+
+      @HostFunction
+      public void set(String key, Object value) {
+         delegate.put(key, value);
+      }
```
lbroudoux: I assume this is because of the JS representation of an object, but shouldn't we also convert on the `set()` method? Should we have a `mapper.readTree()` or something?

webapp/src/main/webapp/angular.json
```diff
@@ -108,5 +108,8 @@
         }
       }
     }
+  },
+  "cli": {
+    "analytics": false
```
lbroudoux: Is it a side change? I don't see how it relates to the topic.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -0,0 +1,194 @@
...
+         this.delegate = store;
+      }
+
+      @HostFunction
+      public String get(String key) {
+         return this.delegate.get(key);
+      }
+
+      @HostFunction
+      public void put(String key, String value) {
```
lbroudoux, line 121: Just to remember it lately: we're missing the `put()` with a TTL function here.

webapp/src/main/webapp/angular.json
andreaTP: good catch, leftover, removed.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: Correct, I just mocked a minimal API to show how it looks like.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: That's possible, I was unsure about what we need to expose to the JS environment from the `requestContext`.
When I have a few meaningful examples of `RequestContext` I'm happy to revisit and iterate on tests to figure out the best user experience.

lbroudoux:
Hey @andreaTP 

I've reviewed the code on this PR, and I like it! The approach is very elegant and clear. I think the "code part" of this proposition is really good, so I'll follow up here on more global questions and concerns.

Regarding your initial questions:

> do we expect people to(only) upload little snippets? In Quickjs4J we invested in the DX to provide the script [as a "library"](https://github.com/roastedroot/quickjs4j/tree/main/core/src/test/resources/)

IMHO, little snippets are the most common way of doing things. This is for mocking purposes so it shouldn't end in a full implementation 😉 That said, having the ability to reuse modules can be nice, and it is actually done (see https://microcks.io/blog/extend-microcks-with-custom-libs/). It requires additional operations, though, like mounting a file volume in the container image or rebuilding a custom image, embedding your modules. Do you confirm we could keep the same principles as those described in this blog post?

> I added a DispatchStyle but I'm unsure it's a good idea, any preference on how to roll this forward?

Ideally, I think it could be nice to keep `SCRIPT` and then dynamically detect if it's old Groovy or JS (or others?).

> is the proposed JsApi reasonable? any specific preference for accessing fields etc.? or high level guidance like: "should be as familiar as possible to JS/TS devs"

It looks simple and neat to me. I think "should be as familiar as possible to JS/TS devs" is very good guidance. I'd like to keep it as idiomatic as possible.

> the current script doesn't have access to FileSystem/Network and similar, so, it fall short on calling external APIs, do you want to make it possible dynamically in the script? Would it be ok to provide a bespoke HTTP invocation API?

Yes, that's a requirement. Regarding the above question, and as stated elsewhere, having the ability to use `fetch()` would be super convenient. It could be an adapted version of `fetch()` that is indeed a `HostFunction`. I assume that we would also have to adapt or get rid of the `async/await ` semantics, but this is one of my other questions 😉 

Some of my questions and concerns:

* How does QuickJS (or QuickJS4J) deal with `async/await` semantics? Would we have to restrict or forbid their usage? Is there a graceful fallback?  I can imagine that this is a common question, even more so if you allow reusing a complete module that may use this. 

* How QuickJS/QuickJS4J compares to GraalJS in terms of:
  * JS/TS/EcmaScript support?
  * Sandboxing model?
  * Weight/size of the library?

* QuickJS4J mentions being `Native-image friendly`. Is there any specific guidance other than traditional class registration for reflection? 

* Is it possible to cache script evaluation between different calls? Recently, we activated caching on Groovy evaluation using the JSR233 mechanisms with nice results (see #1543). Is there some way of doing this in QuickJS4J?

And yes: let's iterate!

andreaTP:
Thanks a lot for the in depth answer @lbroudoux , I hope this is a complete enough answer :crossed_fingers: :

> IMHO, little snippets are the most common way of doing things. 

Ok, agreed, let's keep focusing on code snippets :+1: 

> Do you confirm we could keep the same principles as those described in this blog post?

As far as I understand it there are no limitations here, the JS "library" file can be picked up at runtime and executed dynamically(even in native-mode as we are using a compiled js interpreter).

> dynamically detect

That's possible, but I don't have a great way to do it efficiently, do you have suggestions?
Maybe deprecating `SCRIPT` and adding `GROOVY_SCRIPT` and `JS_SCRIPT`? Or having an additional field for the language selection?

> I'd like to keep it as idiomatic as possible.

Cool :+1: 

> `fetch()`

Added a demo fetch implementation and usage [here](https://github.com/microcks/microcks/pull/1656/commits/dafbf0bf5cb9978a3b4171450d8a905d7e817e4f).

> `async/await`

QuickJs does support `async/await`, but it was not enabled in QuickJs4J, fixing it [here](https://github.com/roastedroot/quickjs4j/pull/52).
By doing it I noticed that things like `setTimeout` are not supported though, I don't think we should limit or forbid the usage, but we need to keep an eye at "how it is used" by the initial use-cases.

> - How QuickJS/QuickJS4J compares to GraalJS in terms of:

>   - JS/TS/EcmaScript support?

The support of EcmaScript is pretty complete [according to the docs](https://bellard.org/quickjs/), and we already stressed it with biggish chunks of JS.
There is no native TS support, but compilers and bundlers [are doing an excellent job here](https://github.com/roastedroot/quickjs4j/tree/main/core/src/test/resources/zod), I haven't attempted to compile TS "on the fly" as it would be required by a "barebone script".
GraalJS is a full fledged Node(V8) alternative, QuickJs is a pure JS runtime and you don't have any Node or browser specific API; the trade offs are very different.

>   - Sandboxing model?

QuickJs is executed in a Wasm sandbox:
- the memory is isolated and stored on the Java heap (there is **no** chance of data corruption caused by a module on the java host), you can also define bounds of it
- the execution honors Threads interruptions, and, by default, the `Runner` timeouts the plugin execution after a few seconds (configurable)
In GraalJS there is a concept of [isolates](https://www.graalvm.org/latest/reference-manual/embed-languages/#polyglot-isolates).

>  - Weight/size of the library?

QuickJs4J is, currently ~700Kb of distributable jar containing QuickJs plus the bindings to it.
This comes with additional transitive dependencies on e.g. Jackson and Chicory.
GraalJS is a very complete runtime and, if you want decent performance, you need to have GraalJIT available, last time I measured GraalJIT was ~7/8Mb.
That's the default for Graal JDK distros, but [might complicate the setup](https://github.com/shaunsmith/wasm-bench/blob/65ec99fbe65562d4c8fef9fee100645eeb3cb017/pom.xml#L179) for other Java distributions.

> Is there any specific guidance other than traditional class registration for reflection?

The expectation is that everything should work out of the box with `native-image` by design, anything else is going to be treated as a bug. (FYI we are testing this aspect in [the Quarkus integration](https://github.com/quarkiverse/quarkus-quickjs4j/blob/6095b064243e33f9b5a83853192a6ecf81d3520e/integration-tests/basic/pom.xml#L16) )

> Is it possible to cache script evaluation between different calls?

Yes, I have just [done it here](https://github.com/roastedroot/quickjs4j/pull/50/files#diff-4ef1f8606c46e5978acb2d20715d109393735dad7190b87f1a0e434627a2abf9R91), but I think we should wait a little to stabilize the design before optimizing this aspect (it might look slightly different depending on how things unroll).

webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
```diff
@@ -0,0 +1,212 @@
...
+      } catch (Exception e) {
+         fail("Exception should no be thrown");
+      }
+   }
+
+   @Test
+   void testFetchApiWorks() {
+      // TODO: make it self-contained and not accessing external resources
+      String script = JsScriptEngineBinder.wrapIntoFunction("""
+            const response = fetch('http://httpbin.org/get');
```
lbroudoux: 👍

lbroudoux:
Awesome! Thanks for the answers. It's super clear now!
The only remaining issue on which I have to form an opinion is the question of `SCRIPT` vs. `GROOVY_SCRIPT` and `JS_SCRIPT`. Groovy scripts could contain very specific constructs like `def` or `java.*` classes usage but those are far from being mandatory, so it wouldn't be enough to decide... Let me think about it.

lbroudoux:
Unfortunately, we got an issue on unit tests;

```
[INFO] Running io.github.microcks.util.script.JsScriptEngineBinderTest
14:00:48.396 [main] INFO  i.g.m.u.script.JsScriptEngineBinder - fetch status: 503
14:00:48.545 [main] INFO  i.g.m.u.script.JsScriptEngineBinder - header: bar
14:00:48.601 [main] INFO  i.g.m.u.script.JsScriptEngineBinder - uri parameter: bar
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.908 s <<< FAILURE! -- in io.github.microcks.util.script.JsScriptEngineBinderTest
Error:  io.github.microcks.util.script.JsScriptEngineBinderTest.testFetchApiWorks -- Time elapsed: 1.282 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at io.github.microcks.util.script.JsScriptEngineBinderTest.testFetchApiWorks(JsScriptEngineBinderTest.java:206)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

@andreaTP Can you check this one?

andreaTP:
> Unfortunately, we got an issue on unit tests;

On it :+1: , please don't consider this PR final just yet, with this I'm mostly gathering feedback and will tight it up later on :slightly_smiling_face:

andreaTP:
@lbroudoux I'll make soon another iteration or two on this PR to (start) bridge the gap on the subjects we discussed and try to finalize it :slightly_smiling_face:

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: done

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: done

andreaTP: went back to the original implementation, rationale:
- we set plain object from Java
- we get the values from JS

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
lbroudoux: Isn't it possible to call it also `put()` but with a third argument? Or the way JavaScript polymorphism is working prevents us to do so?

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
lbroudoux: Not really. The script inserts data or facts into the `requestContext` so that this data can be used later during the response rendering step. So typically, script will compute stuff, insert them in the context and the response will include it.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: > Or the way JavaScript polymorphism is working prevents us to do so?

That's correct:

```js
function add(op1, op2, op3) {
  console.log("3 args");
  return op1 + op2 + op3;
}

function add(op1, op2) {
  console.log("2 args");
  return op1 + op2;
}

add(1, 2);
add(1, 2, 3);
```

prints:

```
2 args
2 args
```

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: oh, ok, do you have an example so that I'm sure I'm testing the right use-case?

andreaTP:
Done an iteration on the code, and I think that most things are in place now.
Notes:

- I went down deprecating `SCRIPT` in favor of `GROOVY`: I though a bit at dynamically parsing, but it seems a bit too much of a guess, and it opens the door to questions like: what if we don't recognize the script?
- The Node `fetch` implementation is pretty dumb atm, but I think this can be improved with iterations and in response to issues, alternatively, I can cover a finite number of use cases :slightly_smiling_face: 

Not fussy on anything here though :slightly_smiling_face:

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: I used this test:
https://github.com/microcks/microcks/pull/1656/files#diff-1c89164c15586f6c599cedee5c6ca2c74546b94fa59ced69cb71869b5c0a8fc1R105 to drive the implementation.

andreaTP:
Sorry for the noise on the repo, the CI workflows should be fine now :crossed_fingers:

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
lbroudoux: I think that's ok. The only missing part here is that later on, `foo` can be reused as a variable in the response template for example:

```json
{
   "foo": "{{ foo }}"
}
```

Then the mock controller will inject the request context into the rendering engine to get the final `{"foo": "bar"}` response.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
lbroudoux: Got it. Thanks for the confirmation.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -0,0 +1,272 @@
...
+import org.apache.hc.core5.http.io.entity.EntityUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
+
+import java.util.Map;
+
+/**
+ * Utility class that holds methods for creating binding environments and evaluation context for a JSR 233 ScriptEngine.
+ * @author laurent
```
lbroudoux: Please put your name here 😉

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -0,0 +1,272 @@
...
+
+      /**
+       * Mimics the browser/node fetch API for GET requests. Is minimal at the moment, we can easily support more
+       * use-cases when needed
+       *
+       * @param url The URL to fetch
+       * @return A JS object with status and text() method
+       */
+      @HostFunction
+      public Object fetch(String url) {
```
lbroudoux: Saw your answers on JavaScript polymorphism and actual `fetch()` implementation. I agree this one can be completed later on but shouldn't we - at least - make more arguments available to avoid too many `fetch*()` methods later on and regression on existing consumers?

Maybe a simple signature like:
```java
public Object fetch(String url, String method, String body, Map<String, String> headers) {
   // ... 
}
```
will do the trick?

lbroudoux (commented):
Got 2 comments left ;-)

lbroudoux:
Hi @andreaTP thanks for the great submission! No noise here: only good stuff!

I have one more question regarding script caching: earlier on, you pointed to a change that was related to JSR233 support (but still [a draft PR](https://github.com/roastedroot/quickjs4j/pull/50/commits/1cf033cbd2e7ec476d5a13caa8c44e7b4001c29d)), and I also see [this commit](https://github.com/roastedroot/quickjs4j/commit/587711950bf1b1ed3ae62f28d72d1cfafc1012ba) related to JS module.

Is there any caching enabled at the time of your latest 0.0.7 release or is it a WIP that is planned?

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: Thanks for the context, in this case I believe the current implementation is "good enough" as the value out of the map is the expected one:
https://github.com/microcks/microcks/pull/1656/files#diff-1c89164c15586f6c599cedee5c6ca2c74546b94fa59ced69cb71869b5c0a8fc1R120

e.g. trying to store values as JsonNode produces double quoting of the strings value

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: Sounds good, I can implement the proposed signature :+1:

andreaTP:
> Is there any caching enabled at the time of your latest 0.0.7 release or is it a WIP that is planned?

I forgot to say it, but I already took care of caching with this change:
https://github.com/microcks/microcks/pull/1656/files#diff-ea5b79b063e2e16bf6ea94e55b0b08f809eab7894662f089eaf19caeeabef23aR54

With the bump to `0.0.7` you get first class `async` support instead :slightly_smiling_face:

lbroudoux:
> I forgot to say it, but I already took care of caching with this change

Ok! I'm reading https://github.com/roastedroot/quickjs4j/blob/main/core/src/main/java/io/roastedroot/quickjs4j/core/ScriptCache.java and the `Engine`, and I see no eviction policy in case the digested key is missing. Imagine the script being updated and replaced many times by people iterating on it via the UI or other means.

Wouldn't it be nice to use at least a `LinkedHashMap` as the internal `cache` to implement something like a LRU policy via `removeEldestEntry()`?

andreaTP:
> Imagine the script being updated and replaced many times by people iterating on it via the UI or other means.

Expecting very small scripts, and the fact that the cache only retains the QuickJs "compiled" version of those I thought it would be a "day 2" issue to clear the cache :sweat_smile: 
But you are correct, and hooking up an LRU or something is easy, I'll improve on this aspect!

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
andreaTP: Done, the handling of JSON should likely be improved, but there is a baseline now.

andreaTP:
Updated the PR:

- Added the `LRUCache` using: `org.codehaus.groovy.runtime.memoize.LRUCache`.
- Implemented the requested `fetch` signature (still missing proper handling of JSON, this PR is growing and I'd prefer to do a follow up in case requested)

I took the opportunity to make things nicer/easier starting in QuickJs4J, hence the version bump :slightly_smiling_face:

andreaTP:
Hi :wave: I'm back from vacation and able to iterate again on this when requested.
Looking through the comments I think that everything has been addressed, let me know if I missed anything :pray:

lbroudoux:
Hey @andreaTP
I'm also now back from vacations and ready to merge that! I plan to integrate this into the next `1.13.0` and just started a `1.13.x` development branch. Would you mind rebasing your work in `1.13.x` and change the PR to target this branch?

andreaTP:
Welcome back @lbroudoux and thanks for picking this up again :slightly_smiling_face: 

Rebased and re-targeted, should be good :+1:

andreaTP:
The Sonar check requires access to secrets, it's not going to pass, I opened #1690 to workaround it.

webapp/src/main/java/io/github/microcks/web/TestController.java
```diff
@@ -174,6 +178,38 @@ public ResponseEntity<TestCaseResult> reportTestCaseResult(@PathVariable("id") S
...
+      return ResponseEntity.ok("PATCH received with body: " + (body != null ? body : "empty"));
+   }
+
+   @GetMapping("/test-fetch-headers")
+   public ResponseEntity<String> testFetchHeaders(HttpServletRequest request) {
+      String customHeader = request.getHeader("X-Custom-Header");
+      String authHeader = request.getHeader("Authorization");
+      return ResponseEntity.ok("Headers received - Custom: " + customHeader + ", Auth: " + authHeader);
+   }
+
```
lbroudoux: I think we have to relocate this to another Controller/REST resource as `TestController` is dedicated to Microcks' TestResult entities management. Couldn't we put this in a dedicated controller under `/src/test/java` instead?

lbroudoux (changes requested):
Got a re-read and added a final concern / change request.

webapp/src/main/java/io/github/microcks/web/TestController.java
andreaTP: Makes sense, I'll have a look on how to turn this around :+1:

andreaTP: done here: https://github.com/microcks/microcks/pull/1656/commits/ab3923398a23c4bd81b381a5b6b3ba41bcd1bc1e

lbroudoux: Awesome! Thanks for the changes. Let's do merge this one!

andreaTP:
@lbroudoux thanks for the support in this initiative :slightly_smiling_face: 

Should we plan follow ups? On top of mind I have:

- json support in the `fetch` clone
- missing docs

lbroudoux:
Yes absolutely!

Documentation is the top priority IMHO and we should add something like we did for Groovy here: https://microcks.io/documentation/explanations/dispatching/#script-dispatcher. Documentation is living in the `microcks/microcks.io` Github repository.

JSON support is also more than welcome!

It would be awesome if you created an issue for each follow-up idea you have in mind so that we can keep them in sight!

Thanks a lot!

PR #1666, opened 2025-07-23
Fixed Schema Resolutions in WSDL files 

PranjaliBhardwaj:
Robust XSD Reference Handling:
The SoapUI importer was refactored to correctly resolve and replace all XSD schema references in WSDL files, handling all reference types (xs:import, xs:include, xs:redefine) and supporting both Windows and Unix path formats.
Comprehensive Testing:
New tests and sample SoapUI projects were added to ensure the new logic works for both Unix-style and Windows-style XSD references, improving reliability and cross-platform compatibility.


lbroudoux:
Hi @PranjaliBhardwaj !

Thanks a lot for this PR and for your patience regarding the review. It has been summer time here and I was in vacation 😉 

I'd like to start reviewing it as we started working on the `1.13.x`, but before that, it would be awesome if you could rebase your branch on the new `1.13.x` branch from the microcks origin and then re-target your PR to this branch.

Thanks again!

PR #1674, opened 2025-08-24
Refactor/Enchance Dialog Upload Component to have better UX

Apoorva64:
### Description
When a user wants to quickly import a specification into Microcks, he has to go though a lot of clicks and page changes.
For exemple when a user is on the homepage and wants to import a multi-part specification (ex: BeerCatalogApi), he has to:
- click on the "importers" page 
- click on the "Upload" button 
- select the BeerCatalogAPI-swagger.json
- click on the "upload" button
- select the BeerCatalogAPI-collection.json
- click on the "upload" button

https://github.com/user-attachments/assets/e142e767-0913-4dfe-8f03-1b35b0a19567

This PR aims to ease this workflows by implementing multi-file upload as well as a drag and drop system

https://github.com/user-attachments/assets/9440ae26-1272-4524-9998-785e4de224ee


Apoorva64:
linked to #1677

webapp/src/main/webapp/src/app/pages/importers/importers.page.ts
```diff
@@ -105,13 +103,20 @@ export class ImportersPageComponent implements OnInit {
...
     this.notifications = this.notificationService.getNotifications();
     this.getImportJobs();
     this.countImportJobs();
 
+    // Register refresh callback for this page
+    this.uploaderDialogService.registerPageRefreshCallback('/importers', () => {
+      this.getImportJobs();
+      this.countImportJobs();
+    });
+
```
lbroudoux: I'm not sure this is necessary, as the upload of new artifacts impacts the services list but not the importers list.

webapp/src/main/webapp/src/app/pages/services/services.page.html
```diff
@@ -6,14 +6,17 @@
 
 <h1>
   APIs &amp; Services
-  <div class="pull-right">
+  <div class="pull-right" >
+    <button class="btn btn-default" (click)="openArtifactUploader()">
+      <span class="fa fa-upload"></span> Upload
+    </button>&nbsp;
```
lbroudoux, line 12: This has to be discussed in favor of a global Upload button in the toolbar. One caveat of having it in the toolbar may be managing the page refresh callback. You may need to check the current location and whether it needs to be refreshed. 🤔

webapp/src/main/webapp/src/app/pages/importers/importers.page.ts
Apoorva64: yep indeed this is not necessary, i am removing it

lbroudoux: Thanks for the quick fix!

webapp/src/main/webapp/src/app/pages/services/services.page.html
lbroudoux: Hey @Apoorva64 
Can you easily propose this addition in the toolbar? I'd really like to see what it looks like.

Also, thinking about it, I have a new idea... 💡 
You may know that Microcks can also download remote artifacts directly without having to create an importer. However, this feature is not directly reachable from the UI. I think it could be nice to also add a second tool in the toolbar that pops up a modal and allow pasting a URL (optionally, picking up a Secret) to have Microck download the artifact.

That way, we would add 2 new tools: **Upload** (which triggers the dialog you created) and **Download** (which triggers a new dialog to be created). I think it would make the user experience drastically simpler. What do you think?

Apoorva64: The refresh logic is already handled globally using callbacks https://github.com/microcks/microcks/pull/1674/files#diff-c318b30f1aed49cc6573950ccb15d6d0d0a5fc9c70eedf8eca8d11104d042a39R69 . I think it should be straightforward to add the upload button anywhere in the UI. Could you make a quick mock-up of your idea (even on paper or in Paint)? I’m not entirely sure where and how you’d like the tools to be added. I was thinking of something similar to [the MUI Speed Dial](https://mui.com/material-ui/react-speed-dial/#basic-speed-dial)

Apoorva64: But first i need to refactor the css to use the variables xD

lbroudoux: I was thinking at something like the sceenshot below.

I made several attempts with different icons here:
* The "Upload" one => that would trigger your "multiple upload dialog"
* The "Download" one => that may trigger a new "paste your artifact URL for microcks to download it"  dialog
* The "Link" one as an alternative of the "download" that may be confusing with upload or a backup feature.

What do you think?

Capture d’écran 2025-08-29 à 10 28 46

Apoorva64: i think the "Upload" and "Download" button is kind of confusing to the user. When a new user goes to the web page, he'll think about "importing" a specification.

i think it would be better to have a "Direct Import" button
image

Which opens a Tabbed Dialog, 
a bit like 
image

Just for clarification:
Upload -> Uploading Files directly from the browser to Microcks using the Upload Dialog
Download -> Make microcks download a file using the hidden import
What does the `link` one do?

lbroudoux: The `link` icon was just there to provide an alternative to the `download` icon, which I also found confusing.

I like the tabbed approach in your modal dialog! I would see only two tabs though but it opens up for other alternatives in the future. So good!

I see two options: 
1. Would you pursue on this PR and build something on top of your `upload-dialog-component`? In that case, it would be better to rename it as `quick-import-dialog-component` maybe?
2. Do you prefer validating this PR as is and continuing in another issue?

Personally, I prefer option 1. but it's up to you.

Apoorva64: i'll take option 1 xD and rename it

Apoorva64: @lbroudoux i've done the changes, but it feels weird to have the Quick Import button as well as the `Upload` button or is it just me xD?

lbroudoux: Do you mean on the top left corner of the navigation toolbar? Or if you consider the "Quick import" from the nav bar + the "Upload" button in the importer page?

Apoorva64: image
On the Service and Importer pages, we have two groups of buttons placed very close to each other, which makes it feel like there are two toolbars. In my opinion, this makes the page look a bit cluttered but idk how to fix it

lbroudoux: Yes, see what you mean. As we now have the "+" button in the navigation bar, I think we can remove the "Upload" button from services page (and also importers page). WDYT?

Apoorva64: The problem is that removing the "Upload" button would break the existing user flow in the UI. Existing users would be confused when trying to import an artifact. I think we should keep both options for now and update the documentation to explain the different methods of importing artifacts in Microcks. Later, we can deprecate the "Upload" button.

Issue #1427, opened 2024-11-28
🧛‍♂️  Accessibility > Please add (or explain how to add a theme) Dark theme (for the :nerd_face: )


adriens:
### Reason/Context

During an internal demo I missed a tab due to the fact I did not see there were two of them, because of a _"white on white"_ effect.

:point_right: The main goal of this issue is to **enhance de UX**

### Description

Please implement a dark theme, see explanations and context in the previous section

### Implementation ideas

- Give a documentation that explains how to implement custom `css` themes that community could offer
- Implement the classical Dark theme
- Implement the [:vampire: Dracula theme](https://draculatheme.com/)

lbroudoux:
Thanks for the suggestion. By the beginning of next year, we should work on upgrading/refactoring our UI framework to switch to something more modern. We'll add this as a requirement for new UI design. Stay tuned!

adriens:
> . Stay tuned!

Of course we do :heavy_heart_exclamation:

lbroudoux:
See https://github.com/orgs/microcks/discussions/1458 and feel free to comment and help us 😉

lbroudoux:
@adriens A first version of this new dark theme is now available in the `nighlty` tagged container image. I'd love to get your feedback on it!

adriens:
We're about to laucnh a maintenance pass on many projects, and we'll put microcks in the loop, so I'll let you know our feedbacks 👐

lbroudoux:
Everything is not perfect yet, but we have some decent stuff to release! Let close this one and treat enhancement requests as iterations on releases after the `1.13.0`.

PR #1675, opened 2025-08-25
Refactor/use variables for css colors

Apoorva64:
### Description
Refactor css to use variables for color so that we can later implement a darkmode
#1427

### DIFF
--- a/webapp/src/main/webapp/src/app/components/grade-index/grade-index.component.css
+++ b/webapp/src/main/webapp/src/app/components/grade-index/grade-index.component.css
@@ -9,33 +9,33 @@
   margin-left: 0;
 }
 .grade-box {
-  background-color: #fff;
-  color: #6a6e73;
-  border-top: calc(1px * 2) solid #f5f5f5;
+  background-color: var(--white);
+  color: var(--text-secondary);
+  border-top: calc(1px * 2) solid var(--surface-subtle);
 }
 .grade-box:last-child {
-  border-right: calc(1px * 2) solid #f5f5f5;
+  border-right: calc(1px * 2) solid var(--surface-subtle);
 }
 
 .grade-box.a {
-  --pfe-health-index--accent: #3f9c35;
-  --pfe-health-index--accent--text: var(--pfe-theme--color--text--on-dark, #fff);
+  --pfe-health-index--accent: var(--status-success);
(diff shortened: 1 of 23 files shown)


lbroudoux:
Thanks for the PR! This is great! I'd love to see the dark them coming for the `1.13.0`release! 🤩 

Could you please rebase your branch on the latest changes from `1.13.x` so that it won't be polluted by changes on `install/*` files? Thanks!

webapp/src/main/webapp/src/app/components/grade-index/grade-index.component.css
```diff
@@ -9,33 +9,33 @@
   margin-left: 0;
 }
 .grade-box {
-  background-color: #fff;
-  color: #6a6e73;
-  border-top: calc(1px * 2) solid #f5f5f5;
+  background-color: var(--white);
```
lbroudoux: Don't you think we should put here a variable name that isn't related to the color but much more to the use of it? I would have seen something like `var(--background)` so that it can later be turned into a dark color?

lbroudoux (commented):
This looks awesome but I have a concern regarding the naming of the `--white` var. Please check my comment.

webapp/src/main/webapp/src/app/components/grade-index/grade-index.component.css
Apoorva64: Yes it would be better to have a name that makes more sense, but the problem is that the white color is not used exclusively for background like here 
https://github.com/microcks/microcks/pull/1675/files#diff-8d36062b4813984732cb1306719396d9c8f2f9e6e1feef838bf0a34c293b7ed3R192

lbroudoux: And couldn't we make the distinction between the different usages? We could introduce different variables for the different usages, even if their value is still `#ffffff` I presume.

Apoorva64: i was looking though it xD, it seems that we use the white color as an accent for text. I'll make 2 new variables `--accent-text` and `--background`

Apoorva64: done in https://github.com/microcks/microcks/pull/1675/commits/217802b30093096a5671be16b649e87d463e8c61

lbroudoux (approved):
This looks great. Thanks!

PR #1690, opened 2025-08-26
Disable not working sonar check on forks

andreaTP:
At best of my understanding Sonar requires access to the `SONAR_TOKEN` secret which is not available on forks for security reasons.

I noticed that some CI jobs are already disabled and made the exclusion consistent with them.


lbroudoux:
You're absolutely right. I think we should also update the branches list. I'm going to do it just after the merge. 

Thanks,

Issue #1728, opened 2025-09-09
Request Explanation & Debugging in Microcks


Apoorva64:
### Description

Currently, when working with Microcks, it can be challenging to understand why a request was matched (or not matched) and to debug issues during dispatching.

#### Proposal

Introduce a **Request Explanation / Debug Panel** within Microcks, supported by a lightweight **per-request tracing system**, that would:

* Expose dispatcher logs and script errors directly in the UI.
* Clearly show which request was executed by Microcks.
* Allow developers to follow the dispatcher’s decision-making process step by step.
* Provide clear feedback when a request fails to match or when no mock exists.

#### Implementation Direction

Leverage **OpenTelemetry** for this per-request tracing:

* Each incoming request could get a **correlation ID / trace ID**.
* Dispatcher decisions, logs, and errors could be attached as **spans or events**.
* Traces could be viewed directly in Microcks UI (for quick debugging) and/or exported to an external tracing system like Jaeger/Tempo/Zipkin for deeper analysis. Traces are to be visualized: 
  * in the service page using a Collapsible,
  * in the Edit operation properties page to enchance the debugging of dispatchers
  * in a new page as a Graph to grasp the whole picture of who and how requests are called
* REQUIRED: Anywhere there are traces in the UI we must provide an easy way to expose the trace ID, or link to another observability system (W3C)

### Analysis
Discussion on the subject: https://github.com/orgs/microcks/discussions/1700
#### Overview

```mermaid
sequenceDiagram
  autonumber
  participant C as Client (HTTP)
  participant RC as RestController
  participant RIP as RestInvocationProcessor
  participant CDC as computeDispatchCriteria (child span)
  participant OTel as OpenTelemetry SDK
  participant P as CustomExplainTraceProcessor
  participant S as SpanStorageService
  participant E as Spring Event Bus
  participant SSE as TracingSseManager
  participant SC as Client (SSE)

  C->>RC: REST request /rest/{service}/{version}/...
  RC->>RC: Tag current span (explain-trace=true,<br/>service/operation attrs)
  RC->>RIP: processInvocation()
  RIP->>CDC: create INTERNAL span (explain-trace=true)
  CDC-->>RIP: dispatch criteria + events
  RIP-->>RC: ResponseResult (mock/proxy)
  RC-->>C: HTTP response

  note over RC,RIP: Spans finish after logic completes
  RC-->>OTel: End server span
  RIP-->>OTel: End internal spans
  CDC-->>OTel: End child span

[...]

PR #1730, opened 2025-09-10
refactor: migrate OpenTelemetry configuration from agent to Spring Boot Starter

Apoorva64:
### Description

Why?
We need to migrate to spring boot starter because:
- it allows us to deeply customize the opentelmetry SDK to add a span processor
- it allows us to have opentelemetry in native image

What do we lose?
- We lose some instrumentation that are included in the JAVA_Agent, see https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/

### Breaking
This PR currently breaks the default opentelemetry configuration. As the Java Agent and the Spring Boot Starter does not use the same ENV vars for enabling and disabling the opentelemetry configuration

The JAVA Agent uses: "OTEL_JAVAAGENT_ENABLED" to enable the agent
But spring boot starter uses: "OTEL_SDK_DISABLED" to disable the sdk

Futhermore, we cannot completely disable the Spring Boot Start with OTEL_SDK_DISABLED, as we want to use Opentelemetry to collect the traces in memory for #1728 that is why in this PR the SDK is not disabled but the exporters are set to none. The best option would be to add something like
```
otel.traces.exporter=${OTEL_JAVAAGENT_ENABLED:true ? 'otlp' : 'none'}
otel.metrics.exporter=${OTEL_JAVAAGENT_ENABLED:true ? 'otlp' : 'none'}
otel.logs.exporter=${OTEL_JAVAAGENT_ENABLED:true ? 'otlp' : 'none'}
```
in the application.properties but it doesn't work that way xD. We could also implement an entrypoint to do this but before going futher i want to discuss a bit about it

### Related issue(s)

See #1728

### DIFF
--- a/webapp/pom.xml
+++ b/webapp/pom.xml
@@ -29,7 +29,7 @@
     <grpc.version>1.70.0</grpc.version>
     <protoc-jar.version>3.21.8</protoc-jar.version>
     <openai-gpt3-java.version>0.18.2</openai-gpt3-java.version>
-    <opentelemetry.version>1.33.6</opentelemetry.version>
+    <opentelemetry.version>2.19.0</opentelemetry.version>
   </properties>
 
   <dependencyManagement>
@@ -50,7 +50,7 @@
       <dependency>
         <groupId>io.opentelemetry.instrumentation</groupId>
         <artifactId>opentelemetry-instrumentation-bom</artifactId>
-        <version>2.19.0</version>
+        <version>${opentelemetry.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -588,15 +588,11 @@
             <version>${spring-boot.version}</version>
             <configuration>
               <environmentVariables>
-                <!-- <JAVA_TOOL_OPTIONS>-javaagent:${user.home}/.m2/repository/io/opentelemetry/javaagent/opentelemetry-javaagent/${opentelemetry.version}/opentelemetry-javaagent-${opentelemetry.version}.jar</JAVA_TOOL_OPTIONS> -->
-                <JAVA_TOOL_OPTIONS>-javaagent:${user.home}/Development/repository/io/opentelemetry/javaagent/opentelemetry-javaagent/${opentelemetry.version}/opentelemetry-javaagent-${opentelemetry.version}.jar</JAVA_TOOL_OPTIONS>
-                <OTEL_JAVAAGENT_ENABLED>true</OTEL_JAVAAGENT_ENABLED>
                 <OTEL_METRICS_EXPORTER>otlp</OTEL_METRICS_EXPORTER>
                 <OTEL_LOGS_EXPORTER>otlp</OTEL_LOGS_EXPORTER>
                 <OTEL_TRACES_EXPORTER>otlp</OTEL_TRACES_EXPORTER>
                 <OTEL_METRIC_EXPORT_INTERVAL>10000</OTEL_METRIC_EXPORT_INTERVAL>
                 <OTEL_EXPORTER_OTLP_TIMEOUT>7000</OTEL_EXPORTER_OTLP_TIMEOUT>
-                <OTEL_RESOURCE_ATTRIBUTES>service.name=${project.artifactId},service.namespace=${project.artifactId}-ns,service.instance.id=${project.artifactId}-dev,service.version=${project.version}</OTEL_RESOURCE_ATTRIBUTES>
               </environmentVariables>
             </configuration>
           </plugin>


webapp/pom.xml
```diff
@@ -47,6 +47,13 @@
         <artifactId>snakeyaml</artifactId>
         <version>${snakeyaml.version}</version>
       </dependency>
+      <dependency>
+        <groupId>io.opentelemetry.instrumentation</groupId>
+        <artifactId>opentelemetry-instrumentation-bom</artifactId>
+        <version>2.19.0</version>
```
lbroudoux: Please introduce a variable here

webapp/pom.xml
```diff
@@ -84,17 +91,18 @@
...
-      <scope>provided</scope>
+      <groupId>io.opentelemetry.instrumentation</groupId>
+      <artifactId>opentelemetry-spring-boot-starter</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.opentelemetry.instrumentation</groupId>
-      <artifactId>opentelemetry-instrumentation-annotations</artifactId>
-      <version>${opentelemetry.version}</version>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-otlp</artifactId>
```
lbroudoux, line 101: Is the `opentelemetry-javaagent` pulled when including this one? I thought your proposition was to keep the Java agent and have it still enabled for the generic use case?

webapp/src/main/docker/Dockerfile
```diff
@@ -19,18 +19,6 @@ RUN microdnf install java-21-openjdk-headless openssl curl-minimal ca-certificat
...
-# By default we deactivate the agent with OTEL_JAVAAGENT_ENABLED=false
-RUN curl https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v1.33.6/opentelemetry-javaagent.jar -Lo /opt/opentelemetry-javaagent.jar
-ENV JAVA_TOOL_OPTIONS=-javaagent:/opt/opentelemetry-javaagent.jar \
-    OTEL_JAVAAGENT_ENABLED=false \
-    OTEL_EXPORTER_OTLP_ENDPOINT=http://o11y_otel:4317 \
-    OTEL_METRICS_EXPORTER=otlp \
-    OTEL_LOGS_EXPORTER=otlp \
-    OTEL_TRACES_EXPORTER=otlp \
-    OTEL_RESOURCE_ATTRIBUTES=service.name=microcks,service.namespace=microcks-ns,service.instance.id=microcks-cnt,service.version=1.11.0
-
```
lbroudoux, line 32: I thought your proposition was to keep the Java agent and have it still enabled for the generic use case. Why removing it from here?

webapp/src/main/resources/config/application.properties
```diff
@@ -84,3 +84,26 @@ ai-copilot.openai.api-key=sk-my-openai-api-key
...
+otel.service.name=microcks
+otel.resource.attributes=service.namespace=microcks-ns,service.instance.id=microcks-cnt,service.version=${project.version}
+
+
+
+# Exporter configuration
+otel.traces.exporter=none
+otel.metrics.exporter=none
+otel.logs.exporter=none
+otel.exporter.otlp.endpoint=http://o11y_otel:4317
```
lbroudoux, line 99: Couldn't we use the OTEL_EXPORTER_OTLP_ENDPOINT environment variable here?

webapp/src/main/resources/config/application.properties
```diff
@@ -84,3 +84,26 @@ ai-copilot.openai.api-key=sk-my-openai-api-key
...
+# Service identification
+otel.service.name=microcks
+otel.resource.attributes=service.namespace=microcks-ns,service.instance.id=microcks-cnt,service.version=${project.version}
+
+
+
+# Exporter configuration
+otel.traces.exporter=none
+otel.metrics.exporter=none
+otel.logs.exporter=none
```
lbroudoux, line 98: Is there some way to condition value based on environment variable?

lbroudoux:
This topic is maybe a bit hard to follow in an asynchronous way and may deserve a dedicated design session. What do you think about organizing a specific chat/video call for that?

Apoorva64:
@lbroudoux I think it would be a great idea to make a call to design all of this. I am available everyday xD, when are you available?

lbroudoux:
> @lbroudoux I think it would be a great idea to make a call to design all of this. I am available everyday xD, when are you available?

I'm available tomorrow morning and afternoon. Let's ping via Discord.

webapp/src/main/resources/config/application.properties
lbroudoux: `$OTEL_TRACES_EXPORTER` and others have precedence on values defined in `application.properties`.

webapp/src/main/resources/config/application.properties
lbroudoux: `$OTEL_EXPORTER_OTLP_ENDPOINT` will have precedence.

webapp/src/main/docker/Dockerfile
lbroudoux: No. The proposition is to replace the Java agent so we can get rid of this.

webapp/pom.xml
lbroudoux: Nope. We will not use it anymore.

webapp/pom.xml
Apoorva64: done in 882ac5684c45c34d68677de60441e9a371d9e08a

Apoorva64:
i think that RestControllerIT.testProxyFallbackWithDelay is a flaky test, i can't replicate it in local

PR #1738, opened 2025-09-12
feat: strategy delay #1591

### LINKED ISSUE #1591: Make the "Default delay" variable dynamic for each endpoint
### Reason/Context

The aim of this evolution would be to make the performance of a component (such as a VAS or a BFF) using mocks more realistic.

### Description

Give the option of making the ‘default delay’ variable dynamic for each endpoint

### Implementation ideas

_No response_

SebastienDegodez:
### Description

- [x] Allows specifying the strategy with the `x-microcks-delay-strategy` header. (3 values: fixed, random and random-xx)
- [x] Introduced DelaySpec record / value object encapsulating delay value and strategy
- [x] Added DelayApplier interface and implementations (Fixed, Random, RandomRanged)
- [x] Implemented DelayApplierFactory for controlled instantiation
- [x] Refactored controllers and MetadataExtractor to use DelaySpec (record)
- [x] Preserved flat persistence structure for Operation (delay value and strategy)

### Related issue(s)

Resolves #1591

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java
+++ b/webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java
@@ -317,7 +317,7 @@ public class MockControllerCommons {
       if (headers.containsKey(MockControllerCommons.X_MICROCKS_DELAY_HEADER)) {
          String delayHeader = headers.getFirst(MockControllerCommons.X_MICROCKS_DELAY_HEADER);
          try {
-            String delayStrategyHeader = headers.getFirst(MockControllerCommons.X_MICROCKS_DELAY_HEADER);
+            String delayStrategyHeader = headers.getFirst(MockControllerCommons.X_MICROCKS_DELAY_STRATEGY_HEADER);
 
             return new DelaySpec(Long.parseLong(delayHeader), delayStrategyHeader);
          } catch (NumberFormatException nfe) {
--- a/webapp/src/test/java/io/github/microcks/web/MockControllerCommonsTest.java
+++ b/webapp/src/test/java/io/github/microcks/web/MockControllerCommonsTest.java
@@ -1,0 +1,45 @@
+package io.github.microcks.web;
+
+import io.github.microcks.util.delay.DelaySpec;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpHeaders;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class MockControllerCommonsTest {
+
+   @Test
+   void shouldUseDelayAndStrategyFromHeaders() {
+      HttpHeaders headers = new HttpHeaders();
+      headers.add(MockControllerCommons.X_MICROCKS_DELAY_HEADER, "500");
+      headers.add(MockControllerCommons.X_MICROCKS_DELAY_STRATEGY_HEADER, "random-10");
+
+      DelaySpec spec = MockControllerCommons.getDelay(headers, null, "fixed");
+      assertNotNull(spec);
+      assertEquals(500L, spec.getBaseValue());
+      assertEquals("random-10", spec.getStrategyName());
+   }
+
+   @Test
+   void shouldBeCaseInsensitiveForStrategyHeader() {
+      HttpHeaders headers = new HttpHeaders();
+      headers.add(MockControllerCommons.X_MICROCKS_DELAY_HEADER, "250");
+      headers.add(MockControllerCommons.X_MICROCKS_DELAY_STRATEGY_HEADER, "RaNdOm-20");
+
+      DelaySpec spec = MockControllerCommons.getDelay(headers, null, "fixed");
+      assertNotNull(spec);
+      assertEquals(250L, spec.getBaseValue());
+      assertEquals("RaNdOm-20", spec.getStrategyName()); // keep header value as provided
+   }
+
+   @Test
+   void shouldFallBackToParameterWhenHeaderValueInvalid() {
+      HttpHeaders headers = new HttpHeaders();
+      headers.add(MockControllerCommons.X_MICROCKS_DELAY_HEADER, "not-a-number");
+
+      DelaySpec spec = MockControllerCommons.getDelay(headers, 100L, "random");
+      assertNotNull(spec);
+      assertEquals(100L, spec.getBaseValue());
+      assertEquals("random", spec.getStrategyName());
+   }
+}


Apoorva64:
Hello, I've read around the code, would it be possible to extract the delay handling to another class so that we can reuse it?

Futhermore I think it would be cool if we could configure the delay using openapi extensions https://microcks.io/documentation/references/artifacts/openapi-conventions/

SebastienDegodez:
> Hello, I've read around the code, would it be possible to extract the delay handling to another class so that we can reuse it?
> 
> Futhermore I think it would be cool if we could configure the delay using openapi extensions https://microcks.io/documentation/references/artifacts/openapi-conventions/

I think it's possible, yes :).
This PR is a draft, I haven't really developed in Java since 2008 (school), and aside from a small feature earlier this year.
I'm having some trouble running the tests locally 😅 (Maven and all... I'm not very familiar with it).

SebastienDegodez:
For the next phase, I’m considering moving everything currently in util.delay (such as the Delay value object (actually rich business object), strategies, and factory) into the domain package. This would better reflect their role as core business objects rather than utilities.

However, I’m still hesitating about the mechanics with Jackson serialization/deserialization. I want to avoid any breaking changes, especially regarding how Operation is persisted and exposed in JSON. My goal is to keep the database and API structure flat (with defaultDelay and defaultDelayStrategy as separate fields).

I’ll need to carefully review the Jackson annotations and possibly add custom getters or setters to ensure backward compatibility for consumers and persistence.

Given my current level in Java, could you guide me @lbroudoux / @Apoorva64  ?

webapp/src/main/java/io/github/microcks/util/delay/Delay.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.util.delay;
+
+/**
```
lbroudoux, line 18: I think this "rich" class could be changed into a Java record that is immutable, with an already existing `toString()` method looking like your implementation and an `equals()` method that does exactly what you coded.

webapp/src/main/java/io/github/microcks/web/DynamicMockRestController.java
```diff
@@ -85,8 +87,9 @@ public DynamicMockRestController(ServiceRepository serviceRepository,
    @PostMapping(value = "/{service}/{version}/{resource}", produces = "application/json")
    public ResponseEntity<String> createResource(@PathVariable("service") String serviceName,
          @PathVariable("version") String version, @PathVariable("resource") String resource,
-         @RequestParam(value = "delay", required = false) Long delay, @RequestBody(required = true) String body,
-         HttpServletRequest request) {
+         @RequestParam(value = "delay", required = false) Long requestedDelay,
+         @RequestParam(value = "delayStrategy", required = false) String requestedDelayStrategy,
```
lbroudoux, line 92: I forgot we also had this way of specifying stuff... Maybe we should leave this aside for sake of simplicity now because it's now a bit redundant with the `x-microcks-delay` way of doing things... 🤔 `

webapp/src/main/java/io/github/microcks/web/DynamicMockRestController.java
```diff
@@ -212,12 +231,18 @@ public ResponseEntity<String> getResource(@PathVariable("service") String servic
    @PutMapping(value = "/{service}/{version}/{resource}/{resourceId}", produces = "application/json")
    public ResponseEntity<String> updateResource(@PathVariable("service") String serviceName,
          @PathVariable("version") String version, @PathVariable("resource") String resource,
-         @PathVariable("resourceId") String resourceId, @RequestParam(value = "delay", required = false) Long delay,
+         @PathVariable("resourceId") String resourceId,
+         @RequestParam(value = "delay", required = false) Long requestedDelay,
+         @RequestParam(value = "delayStrategy", required = false) String requestedDelayStrategy,
```
lbroudoux, line 237: See above comment on `createResource()` method.

webapp/src/main/java/io/github/microcks/util/delay/Delay.java
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * @author SebastienDegodez
+ */
+public class Delay {
```
lbroudoux: I would have called this differently. Maybe `DelaySpec` as this is a specification: a base value + a strategy.

webapp/src/main/java/io/github/microcks/util/delay/DelayStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.util.delay;
+
+/**
+ * 
+ * @author SebastienDegodez
+ */
+public interface DelayStrategy {
```
lbroudoux: I would have expect `DelayStrategy` being a enumeration of different supported strategies: `fixed`, `random`  and `range`

lbroudoux: In fact I would say that this interface (and implementing classes) are much more `DelayComputers` or `DelayAppliers` => they applied a strategy to a base value to get an exact duration.

webapp/src/main/java/io/github/microcks/util/delay/FixedDelayStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+   } // package-private
+
+   @Override
+   public long compute(long baseValue) {
+      return baseValue;
+   }
+
+   @Override
+   public String getName() {
+      return "fixed";
```
lbroudoux: We need to extract this to a constant here. Will be typically useful when creating a default delay with non specified strategy.

lbroudoux (commented):
Made some comments along the way. It looks good but I would have called things differently. Please have a look ant let me know what you think.

webapp/src/main/java/io/github/microcks/web/DynamicMockRestController.java
SebastienDegodez: > I forgot we also had this way of specifying stuff... Maybe we should leave this aside for sake of simplicity now because it's now a bit redundant with the `x-microcks-delay` way of doing things... 🤔 `

Do you want to remove `requestDelay` too ? Is it a breaking change?

webapp/src/main/java/io/github/microcks/util/delay/DelayStrategy.java
SebastienDegodez: Resolve: Rename interface to DelayApplier

webapp/src/main/java/io/github/microcks/util/delay/FixedDelayStrategy.java
SebastienDegodez: Ok :)

webapp/src/main/java/io/github/microcks/util/delay/Delay.java
SebastienDegodez: Resolve: Rename Delay to DelaySpec.

webapp/src/main/java/io/github/microcks/util/delay/Delay.java
SebastienDegodez: class to record ok ;)

SebastienDegodez:
> Made some comments along the way. It looks good but I would have called things differently. Please have a look ant let me know what you think.

Rename, constante, etc ... : Done.

Two subjects: 
- https://github.com/microcks/microcks/pull/1738#issuecomment-3289644533
- https://github.com/microcks/microcks/pull/1738#discussion_r2349958500

webapp/src/main/java/io/github/microcks/web/DynamicMockRestController.java
lbroudoux: Nope we have to keep it and not break the user space.

lbroudoux:
I think we're on track! What do you have in the mind regarding `Preserved flat persistence structure for Operation (delay value and strategy)` I'm not sure to understand...

webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java
```diff
@@ -306,32 +313,40 @@ public static EvaluableRequest buildEvaluableRequest(String requestBody, String
    }
 
    /** Retrieve delay header or default to the one provided as parameter. */
-   public static Long getDelay(HttpHeaders headers, Long delayParameter) {
+   public static DelaySpec getDelay(HttpHeaders headers, Long delayParameter, String strategyName) {
       if (headers.containsKey(MockControllerCommons.X_MICROCKS_DELAY_HEADER)) {
          String delayHeader = headers.getFirst(MockControllerCommons.X_MICROCKS_DELAY_HEADER);
          try {
-            return Long.parseLong(delayHeader);
+            String delayStrategyHeader = headers.getFirst(MockControllerCommons.X_MICROCKS_DELAY_HEADER);
```
lbroudoux: I think you should use `X_MICROCKS_DELAY_STRATEGY_HEADER` instead.

SebastienDegodez:
> I found a small glitch when you're extraction incoming request header for delay strategy. Please fix this one and we'll be good to go!

During the refactoring, I accidentally introduced this issue that wasn’t caught by the initial test used to implement the delay. I’ve now added a much more precise test specifically targeting `getDelay`

lbroudoux (approved):
Looks al good to me! Thanks a lot @SebastienDegodez 🎉

PR #1752, opened 2025-09-24
feat: add basic tracing storage and query

Apoorva64:
### Description

This PR adds the basic tracing implementation into microcks which contains the 'Processing','Storage' and query.

This PR does not contain the SSE endpoint but the prequisites for this endpoint
### Related issue(s)

#1728

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/otel/CustomExplainTraceProcessor.java
+++ b/webapp/src/main/java/io/github/microcks/util/otel/CustomExplainTraceProcessor.java
@@ -16,6 +16,7 @@
 package io.github.microcks.util.otel;
 
 import io.github.microcks.service.SpanStorageService;
+import io.github.microcks.util.tracing.TraceUtil;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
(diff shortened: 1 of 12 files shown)


webapp/src/main/java/io/github/microcks/util/otel/CustomExplainTraceProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+      return false; // We don't need to be notified when spans start
+   }
+
+   @Override
+   public void onEnd(@Nonnull ReadableSpan span) {
+      // Called when a span ends - delegate to storage service if span matches filter criteria
+      // if span has attribute explain-trace save it
+      Map<AttributeKey<?>, Object> attributes = span.toSpanData().getAttributes().asMap();
+      if (attributes.keySet().stream()
+            .anyMatch(key -> SpanStorageService.valuesEqualAttr(key, AttributeKey.booleanKey("explain-trace")))) {
```
lbroudoux: Shouldn't we put this `explain-trace` string as a constant somewhere? I suspect it will be also used by mock controllers to correctly initialize spans...

Apoorva64: @lbroudoux i've added a TraceUtil class which has the constant of "explain-trace" https://github.com/microcks/microcks/pull/1752/files#diff-8e815f5dd7d43468c14b37f0256811bcb1bf0b96c8f138d0d664cb5b72cc5583R30 and i've added a CommonEvents enum to centralize the otel event names

Waiting on this PR to refactor the #1753

webapp/src/main/java/io/github/microcks/util/tracing/CommonEvents.java
```diff
@@ -0,0 +1,17 @@
+package io.github.microcks.util.tracing;
```
lbroudoux, line 16: Could you please add the license header here?

Apoorva64: yep done in 

[8ce3f15](https://github.com/microcks/microcks/pull/1752/commits/8ce3f15cbb282cd97a7813bfe3b817abd2e0f5be)

lbroudoux (approved):
LGTM

PR #1753, opened 2025-09-24
Feat/add instrumentation to scripts (Soap, Rest)

Apoorva64:
### Description
This PR adds instrumentation to JS and Groovy Script in Soap And Rest Flows.

The JS and Grovy scripts are not handled in the same way. The Groovy script engine throws when an error happens but the JS engine encapsulates the error, This is why the JS script instrumentation is added to the engine and the groovy script instrumentation is added to the Processor. I think we need a refactor to cleaunp this difference.

### Related issue(s)
#1728

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/RestInvocationProcessor.java
+++ b/webapp/src/main/java/io/github/microcks/web/RestInvocationProcessor.java
@@ -38,10 +38,10 @@ import io.github.microcks.util.dispatcher.ProxyFallbackSpecification;
 import io.github.microcks.util.el.EvaluableRequest;
 import io.github.microcks.util.script.JsScriptEngineBinder;
 import io.github.microcks.util.script.ScriptEngineBinder;
-
 import io.github.microcks.util.tracing.CommonEvents;
-import io.github.microcks.util.tracing.OpenTelemetryResolver;
+import io.github.microcks.service.OpenTelemetryResolverService;
 import io.github.microcks.util.tracing.TraceUtil;
+
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
@@ -95,7 +95,7 @@ public class RestInvocationProcessor {
    private final ProxyService proxyService;
 
    private ScriptEngine scriptEngine;
-   private final OpenTelemetryResolver opentelemetryResolver;
+   private final OpenTelemetryResolverService opentelemetryResolverService;
 
    @Value("${mocks.enable-invocation-stats}")
    private Boolean enableInvocationStats;
@@ -109,13 +109,13 @@ public class RestInvocationProcessor {
     */
    public RestInvocationProcessor(ServiceStateRepository serviceStateRepository, ResponseRepository responseRepository,
          ApplicationContext applicationContext, ProxyService proxyService,
-         OpenTelemetryResolver opentelemetryResolver) {
+         OpenTelemetryResolverService opentelemetryResolverService) {
       this.serviceStateRepository = serviceStateRepository;
       this.responseRepository = responseRepository;
       this.applicationContext = applicationContext;
       this.proxyService = proxyService;
       this.scriptEngine = new ScriptEngineManager().getEngineByExtension("groovy");
-      this.opentelemetryResolver = opentelemetryResolver;
+      this.opentelemetryResolverService = opentelemetryResolverService;
    }
 
    /**
@@ -278,7 +278,8 @@ public class RestInvocationProcessor {
       String dispatchCriteria = null;
       Map<String, Object> requestContext = null;
       // Create an INTERNAL child span explicitly because Spring AOP does not apply to private/self-invoked methods.
-      Tracer tracer = opentelemetryResolver.getOpenTelemetry().getTracer(RestInvocationProcessor.class.getName());
+      Tracer tracer = opentelemetryResolverService.getOpenTelemetry()
+            .getTracer(RestInvocationProcessor.class.getName());
       Span childSpan = tracer.spanBuilder("computeDispatchCriteria").setSpanKind(SpanKind.INTERNAL).startSpan();
       try (Scope ignored = childSpan.makeCurrent()) {
          TraceUtil.enableExplainTracing();
--- a/webapp/src/main/java/io/github/microcks/service/OpenTelemetryResolverService.java
+++ b/webapp/src/main/java/io/github/microcks/service/OpenTelemetryResolverService.java
@@ -1,19 +1,20 @@
-package io.github.microcks.util.tracing;
+package io.github.microcks.service;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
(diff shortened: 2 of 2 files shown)


webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -24,6 +24,7 @@
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import io.github.microcks.service.StateStore;
 import io.github.microcks.util.http.HttpHeadersUtil;
+import io.opentelemetry.api.trace.StatusCode;
```
lbroudoux, line 31: Can we preseve one-line break between microcks specific imports and other 3rd parties?

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
```diff
@@ -317,12 +334,29 @@ public static String invokeProcessFn(String script, Engine scriptContext) {
...
+                     .put("dispatch.type", "SCRIPT").put("dispatch.result", res).build());
+         return res;
+
       } catch (Exception e) {
          log.error("Error during JS evaluation", e);
+         Span.current().recordException(e);
+         Span.current().addEvent("dispatch_criteria_result",
+               Attributes.builder().put("message", "Failed to compute dispatch criteria using JS dispatcher")
+                     .put("dispatch.type", "JS").put("dispatch.result", "null")
+                     .put("dispatch.script", script == null ? "" : script).build());
```
lbroudoux: Shouldn't we put somewhere that event components are `separated_with_underscore` (like `dispatch_criteria_result`) and attribute components are `separated-with-dash` (like `dispatch.result`)? How to ensure we're consistent over the different controllers and utilities that produce spans?
Shouldn't we add common events (ex: `dispatch_criteria_result`) as constants somewhere?

These questions don't need to be solved today; we can refactor that later. I was just thinking out loud ;-)

webapp/src/main/java/io/github/microcks/util/script/ScriptEngineBinder.java
```diff
@@ -41,6 +45,57 @@ public class ScriptEngineBinder {
...
+               .put(AttributeKey.stringKey("script.engine"), "groovy")
+               .put(AttributeKey.stringKey("message"), "Script log message");
+         if (t != null) {
+            b.put(AttributeKey.stringKey("exception.type"), t.getClass().getName());
+            if (t.getMessage() != null) {
+               b.put(AttributeKey.stringKey("exception.message"), t.getMessage());
+            }
+         }
+         Span.current().addEvent("script.log", b.build());
+      }
```
lbroudoux: Couldn't we share that with the `JsScrpitEngineBinder`?

webapp/src/main/java/io/github/microcks/web/RestInvocationProcessor.java
```diff
@@ -268,78 +274,98 @@ private DispatchContext computeDispatchCriteria(Service service, String dispatch
...
-            case DispatchStyles.SCRIPT:
-               log.info("Use the \"GROOVY\" Dispatch Style instead.");
-               // fallthrough
-            case DispatchStyles.GROOVY:
-               requestContext = new HashMap<>();
-               Map<String, String> uriParameters = DispatchCriteriaHelper.extractMapFromURIPattern(uriPattern,
-                     resourcePath);
-               try {
+      // Create an INTERNAL child span explicitly because Spring AOP does not apply to private/self-invoked methods.
+      Tracer tracer = openTelemetry.getTracer("io.github.microcks.web.RestInvocationProcessor");
```
lbroudoux: Can't we use class name here?

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
Apoorva64: The events name and attributes should follow https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations . For the common events i'm working on it xD

lbroudoux:
@Apoorva64, don't forget to sign your commits! The DCO check is failing at the moment.

webapp/src/main/java/io/github/microcks/util/script/JsScriptEngineBinder.java
Apoorva64: done

webapp/src/main/java/io/github/microcks/web/RestInvocationProcessor.java
Apoorva64: done

webapp/src/main/java/io/github/microcks/util/script/ScriptEngineBinder.java
Apoorva64: i've added a util in the TraceUtil to avoid duplication

lbroudoux (dismissed):
Forgot to approve ;-) but the build is failing... It looks like the tests are broken

lbroudoux:
There's an issue in tests:
```
Error creating bean with name 'restInvocationProcessor' defined in file [/home/runner/work/microcks/microcks/webapp/target/classes/io/github/microcks/web/RestInvocationProcessor.class]: Unsatisfied dependency expressed through constructor parameter 4: No qualifying bean of type 'io.opentelemetry.api.OpenTelemetry' available: expected at least 1 bean which qualifies as autowire candidate.
```
It looks like there's no `OpenTelemetry` bean loaded in the test context. I wonder if this is mandatory as we should be able to totally disable live traces (when running in uber native within testcontainers for example). 

Maybe we should put the `RestInvocationProcess` constructor argument as `@Nullable` and add some extra guards to avoid this kind of error?

Apoorva64:
> There's an issue in tests:
> 
> ```
> Error creating bean with name 'restInvocationProcessor' defined in file [/home/runner/work/microcks/microcks/webapp/target/classes/io/github/microcks/web/RestInvocationProcessor.class]: Unsatisfied dependency expressed through constructor parameter 4: No qualifying bean of type 'io.opentelemetry.api.OpenTelemetry' available: expected at least 1 bean which qualifies as autowire candidate.
> ```
> 
> It looks like there's no `OpenTelemetry` bean loaded in the test context. I wonder if this is mandatory as we should be able to totally disable live traces (when running in uber native within testcontainers for example).
> 
> Maybe we should put the `RestInvocationProcess` constructor argument as `@Nullable` and add some extra guards to avoid this kind of error?

i've fixed the build by creating a resolver which  resolves the Opentelemetry bean depending of where it is. When disabling the traces with otel.sdk.disabled it still works, it ignores the traces as expected

lbroudoux:
> i've fixed the build by creating a resolver which resolves the Opentelemetry bean depending of where it is. When disabling the traces with otel.sdk.disabled it still works, it ignores the traces as expected

It works, but it creates yet another bean. Wouldn't it be possible to just use `@Nullable`? If there is no other comment, I'll merge it like this, but add this point as a refactoring before the release.

PR #1754, opened 2025-09-25
Feat/add sse endpoint

Apoorva64:
### Description

This PR adds the sse endpoint mentioned in #1728 with the addition of regex filtering and the feature of filtering over clientAddress
### Related issue(s)

#1728

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/TraceSubscriptionManager.java
+++ b/webapp/src/main/java/io/github/microcks/web/TraceSubscriptionManager.java
@@ -1,13 +1,14 @@
-package io.github.microcks.service;
+package io.github.microcks.web;
 
 import io.github.microcks.event.TraceEvent;
+import io.github.microcks.service.SpanStorageService;
 import io.github.microcks.util.tracing.SpanFilterUtil;
 
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import org.springframework.context.event.EventListener;
 import org.springframework.scheduling.annotation.Scheduled;
-import org.springframework.stereotype.Service;
+import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.mvc.method.annotation.SseEmitter;
 
 import java.io.IOException;
@@ -20,13 +21,13 @@ import java.util.concurrent.CopyOnWriteArrayList;
  * Service for managing subscriptions to trace updates via Server-Sent Events (SSE). Clients can subscribe with filters
  * on service name and operation name, and will receive updates when matching traces are updated.
  */
-@Service
-public class TraceSubscriptionService {
+@Component
+public class TraceSubscriptionManager {
 
    private final SpanStorageService spanStorageService;
    private final List<Subscription> subscriptions = new CopyOnWriteArrayList<>();
 
-   public TraceSubscriptionService(SpanStorageService spanStorageService) {
+   public TraceSubscriptionManager(SpanStorageService spanStorageService) {
       this.spanStorageService = spanStorageService;
    }
 
--- a/webapp/src/main/java/io/github/microcks/event/TraceEvent.java
+++ b/webapp/src/main/java/io/github/microcks/event/TraceEvent.java
@@ -1,4 +1,4 @@
 package io.github.microcks.event;
 
-public record TraceEvent(String traceId, String service, String operation) {
+public record TraceEvent(String traceId, String service, String operation, String clientAddress) {
 }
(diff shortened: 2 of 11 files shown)


lbroudoux:
I think some of the commits are already embedded in #1752 or #1753. Do you plan to rebase before finalizing?

Apoorva64:
> I think some of the commits are already embedded in #1752 or #1753. Do you plan to rebase before finalizing?

Yes i am planning of merging https://github.com/microcks/microcks/pull/1752 and https://github.com/microcks/microcks/pull/1753 before rebasing this one

Apoorva64:
@lbroudoux i have a problem when adding Integration tests for the TracingController.
When i make a request with 'restTemplate' it creates a span which becomes the root span of the current context. This span does not have the 'explain-trace' tag so the 'CustomExplainTraceProcessor' does not store this root span and the SpanStorageService does not send an event as it does not recieve the "root" span.

To solve this we could add another tag (an attribute in the span) which indicates which is the root span but it adds unecessary code

Apoorva64:
> @lbroudoux i have a problem when adding Integration tests for the TracingController. When i make a request with 'restTemplate' it creates a span which becomes the root span of the current context. This span does not have the 'explain-trace' tag so the 'CustomExplainTraceProcessor' does not store this root span and the SpanStorageService does not send an event as it does not recieve the "root" span.
> 
> To solve this we could add another tag (an attribute in the span) which indicates which is the root span but it adds unecessary code

i found how to do it. In the test we disable "otel.instrumentation.spring-web.enabled" which disables the restTemplate instrumentation

lbroudoux:
> i found how to do it. In the test we disable "otel.instrumentation.spring-web.enabled" which disables the restTemplate instrumentation

I don't see anything related to that property in what you committed. Is it already pushed?

webapp/src/main/java/io/github/microcks/web/TraceSubscriptionManager.java
```diff
@@ -0,0 +1,103 @@
...
+
+import io.github.microcks.event.TraceEvent;
+import io.github.microcks.util.tracing.SpanFilterUtil;
+
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import org.springframework.context.event.EventListener;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+import org.springframework.web.servlet.mvc.method.annotation.SseEmitter;
```
lbroudoux, line 12: I'm not a big fan of having the `Service` component being dependent on the web layer.... I think we should let this related logic into the `web` package.

webapp/src/main/java/io/github/microcks/web/TraceSubscriptionManager.java
```diff
@@ -0,0 +1,103 @@
...
+         emitter.send(SseEmitter.event().name("heartbeat").data(Collections.emptyList()));
+      } catch (IOException ignored) {
+      }
+   }
+
+   /**
+    * Handle trace updates and push data to matching subscribers.
+    */
+   @EventListener
+   public void onTraceUpdated(TraceEvent event) {
```
lbroudoux, line 66: I think it would be better to move this one to a dedicated listener that implements the logic of ending a trace. Then, all the other things related to subscription or hearbeat should move to the `web` layer in the `TracingController` IMHO.

lbroudoux (changes requested):
Thanks a lot once again for this big PR! I would move stuffs from the tracing service in other parts such as existing TracingController or another listener.

webapp/src/main/java/io/github/microcks/web/TraceSubscriptionManager.java
Apoorva64: The 'onTraceUpdated' method doesn't have logic that is not related to the subscriptions, it is the bridge from Spring events to the subscriptions. So i don't think we should extract it to a custom listener in the listener package. If we do extract the listener it would either be empty or contain the subscription logic

How about we rename the TraceSubscriptionService to TraceSubscriptionManager and move it to the web package and make it an "@Component"?

lbroudoux: Yeah. Looks like the good approach!

Apoorva64: done in https://github.com/microcks/microcks/pull/1754/commits/07d9b154f6724de301fb78b00223161c65ee03f5

lbroudoux (changes requested):
All good. Thanks for applying the requested changes.

PR #1766, opened 2025-10-02
Feat/add live traces component

Apoorva64:
### Description
This PR adds the live traces component to the operation override page. The component is also available on the /traces page with additional customization.

I've just added the component to the page but i think we need to rethink the layout of the page to integrate it

### Related issue(s)

#1728 

https://github.com/user-attachments/assets/3ade8b65-7568-44e3-a2d6-80df1feababa

### DIFF
--- a/webapp/src/main/webapp/src/app/components/live-traces/live-traces.component.ts
+++ b/webapp/src/main/webapp/src/app/components/live-traces/live-traces.component.ts
@@ -85,7 +85,6 @@ export class LiveTracesComponent implements OnInit, OnDestroy {
       .subscribe({
         next: (spans) => {
           if (spans && spans.length > 0) {
-            console.log("Received spans", spans);
             // Get trace ID from first span
             const traceId = spans[0]?.spanContext()?.traceId;
             if (traceId && !this.seenTraceIds.has(traceId)) {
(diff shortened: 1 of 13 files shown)


webapp/src/main/webapp/src/app/components/live-traces/trace-group-list/trace-group-list.component.ts
```diff
@@ -0,0 +1,189 @@
+import { Component, Input } from "@angular/core";
```
lbroudoux, line 1: Could you add the license header on top of this page, please?

webapp/src/main/webapp/src/app/components/live-traces/live-traces.component.ts
```diff
@@ -0,0 +1,256 @@
+import { Component, Input, OnDestroy, OnInit } from "@angular/core";
```
lbroudoux, line 1: Could you add the license header on top of this page, please?

webapp/src/main/webapp/src/app/utils/trace-utils.ts
```diff
@@ -0,0 +1,70 @@
+import { ReadableSpan } from "@opentelemetry/sdk-trace-base";
```
lbroudoux, line 1: Could you add the license header on top of this page, please?

lbroudoux (approved):
Looks all good to me. I'm gonna add the missing headers while playing around with the code.

PR #1770, opened 2025-10-06
feat: add live traces component to service detail

Apoorva64:
### Description

- This PR adds the live traces component to the service page

https://github.com/user-attachments/assets/d1584056-5b4f-493f-9c79-9c7d97575099

### Related issue(s)

#1728

### DIFF
--- a/webapp/src/main/webapp/src/app/components/collapsible-live-traces/collapsible-live-traces.component.css
+++ b/webapp/src/main/webapp/src/app/components/collapsible-live-traces/collapsible-live-traces.component.css
@@ -1,0 +1,70 @@
+/* Collapsible Live Traces Component Styles */
+
+.panel-container {
+  border-left: 1px solid #d1d1d1;
+  padding-left: 20px;
+  position: relative;
+  height: 100%;
+  flex-shrink: 0;
+}
+
+.resize-handle {
+  position: absolute;
+  left: 0;
+  top: 0;
+  bottom: 0;
+  width: 5px;
+  cursor: ew-resize;
+  background: transparent;
+  z-index: 10;
+}
+
+.resize-indicator {
+  position: absolute;
+  left: 2px;
+  top: 50%;
+  transform: translateY(-50%);
+  width: 1px;
+  height: 50px;
+  background: #ccc;
+}
+
+.toggle-button {
+  position: absolute;
+  left: -30px;
+  top: 50%;
+  transform: translateY(-50%);
+  padding: 8px 6px;
+  border-radius: 3px;
+  z-index: 20;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.2);
+  writing-mode: vertical-rl;
+  text-orientation: mixed;
+  white-space: nowrap;
+}
+
+.toggle-button-text {
+  transform: rotate(180deg);
+  display: inline-block;
+}
+
+.toggle-button-icon {
+  margin-bottom: 5px;
+}
+
+.content-container {
+  position: relative;
+  height: 100%;
+  overflow-y: auto;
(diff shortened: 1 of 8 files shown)


lbroudoux:
At first thought, I would have put this new "Live Traces" block under the mocks - being folded by default. What do you think?

Apoorva64:
> At first thought, I would have put this new "Live Traces" block under the mocks - being folded by default. What do you think?

just saw the message, i'm moving it

Apoorva64:
moved down and collapsed by default 
image

Apoorva64:
@lbroudoux There is a formatting error in https://github.com/microcks/microcks/blob/1.13.x/webapp/src/main/java/io/github/microcks/event/TraceEvent.java on 1.13.x

lbroudoux:
> @lbroudoux There is a formatting error in https://github.com/microcks/microcks/blob/1.13.x/webapp/src/main/java/io/github/microcks/event/TraceEvent.java on 1.13.x

Oops... let me fix that.

Apoorva64:
Moving the LiveTrace component down makes a lot of whitespace tho
image

It makes the UI feel empty at that place

lbroudoux:
Done. Sorry for this.

Apoorva64:
I think we should keep the live traces before the mocks, as it makes an empty block of space. Futhermore the "Response" of an example can be really long which would make navigating to the Live Traces harder

lbroudoux:
Oh yes it's a bit quirky... hum let me think about that... maybe we should think about another way of displaying this like a sliding panel coming from the bottom of right side of the page...

lbroudoux:
A sliding panel could be also useful in the operation properties page. Something like a "debug" panel or something.

Apoorva64:
Could you make a quick drawing of your idea? The way I’m imagining the sliding panel, the traces wouldn’t be as easily correlated with the operation and would make the UI more complicated to navigate.

lbroudoux:
We could add the "operation" select box you have on the raw component but also try to track the currently focused/displayed operation to change it dynamically.

lbroudoux:
Or the panel could be included into the operation section panel...

lbroudoux:
In fact, it would be better to have the panel within the operation panel. As most spans are made of short key/value pairs, it doesn't need to have the full width.

Apoorva64:
Could you make a quick drawing of where to component should be inserted?

lbroudoux:
I was thinking of something like that for the service details page:

Capture d’écran 2025-10-07 à 18 23 31
Capture d’écran 2025-10-07 à 18 23 43

And for the "Edit properties" page:

Capture d’écran 2025-10-07 à 18 24 00
Capture d’écran 2025-10-07 à 18 24 19

Don't know if it's easy doable though.... Ideally, this would be something that can be resized externally but would have scrolling bars in case the content from the live trace component is too large or too high

Apoorva64:
@lbroudoux i think i got something. What do you think about this approach?

https://github.com/user-attachments/assets/2bea736f-f241-48e8-ae57-f9e19123a96b

lbroudoux:
I love this! It looks nice and neat! 
How the left panel behaves in case you have a small screen? Can you expand the live traces panel to the full width without making the mocks panel too messy?

Apoorva64:
https://github.com/user-attachments/assets/e011e4a3-0f90-4f37-96f2-04b0fb9575d4

Apoorva64:
@lbroudoux i fixed the responsive layout

lbroudoux:
That's awesome! Thank you very much!

lbroudoux:
What do you think of adding this kind of icon? Like in Intellij IDE?

Capture d’écran 2025-10-09 à 15 26 57

PR #1772, opened 2025-10-07
Feat/add explain event to rest invocation processor

Apoorva64:
### Description
This PR adds the "explain" events to the rest of the dispatchers as well as the fallback logic in the rest invocation Processor
### Related issue(s)
#1728


lbroudoux (approved):
LGTM. Thanks!

PR #1802, opened 2025-10-23
fix: Raise the max number of traces

Apoorva64:
### Description

In the live traces graph component raise the number of traces to a more useful size

### Related issue(s)

#1700


lbroudoux (approved):
LGTM

lbroudoux:
Thanks for being cautious @Apoorva64 , I didn't spot these ones during the review.

Issue #1801, opened 2025-10-23
Serialization of traces: No serializer found and no properties discovered to create `BeanSerializer`


Apoorva64:
### Describe the bug

when quering traces i get a serialization error:

```bash
2025-10-23 16:26:33.917 | exec java -Dspring.aot.enabled=true -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/app.jar
2025-10-23 16:26:35.991 | 14:26:35,988 |-WARN in ch.qos.logback.classic.model.processor.ConfigurationModelHandlerFull - Missing watchable .xml or .properties files.
2025-10-23 16:26:35.992 | 14:26:35,991 |-WARN in ch.qos.logback.classic.model.processor.ConfigurationModelHandlerFull - Watching .xml files requires that the main configuration file is reachable as a URL
2025-10-23 16:26:36.098 | 
2025-10-23 16:26:36.098 |   .   ____          _            __ _ _
2025-10-23 16:26:36.098 |  /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
2025-10-23 16:26:36.098 | ( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
2025-10-23 16:26:36.098 |  \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
2025-10-23 16:26:36.098 |   '  |____| .__|_| |_|_| |_\__, | / / / /
2025-10-23 16:26:36.098 |  =========|_|==============|___/=/_/_/_/
2025-10-23 16:26:36.098 | 
2025-10-23 16:26:36.098 |  :: Spring Boot ::                (v3.5.5)
2025-10-23 16:26:36.098 | 
2025-10-23 16:26:37.299 | 14:26:37.281  INFO 1 --- [      main] i.g.microcks.MicrocksApplication         : Starting AOT-processed MicrocksApplication using Java 21.0.8 with PID 1 (/deployments/app.jar started by ? in /deployments)
2025-10-23 16:26:37.302 | 14:26:37.301  INFO 1 --- [      main] i.g.microcks.MicrocksApplication         : The following 1 profile is active: "uber"
2025-10-23 16:26:39.266 | 14:26:39.266  INFO 1 --- [      main] i.g.microcks.config.WebConfiguration     : Starting web application configuration, using profiles: [uber]
[...]

PR #1803, opened 2025-10-23
QuickFix: jackson config to not fail on empty beans that happens in the Groovy exceptions

Apoorva64:
…

### Description
- This PR is a quick fix of the Serilization bug in the groovy exception

### Related issue(s)

#1801


lbroudoux:
Thanks @Apoorva64 for the quick fix!

Issue #1807, opened 2025-10-26
Enhancement/Optimisation: Reduce the data sent to the frontend in Live Traces


Apoorva64:
### Reason/Context

To make the frontend faster we need to reduce the size of the data parsed by the browser. 

### Description

Here is what a span looks like currently:

```json
    {
      "attributes": {
        "explain-trace": true
      },
      "totalAttributeCount": 1,
      "totalRecordedEvents": 1,
      "totalRecordedLinks": 0,
      "status": {
        "statusCode": "UNSET",
        "description": ""
      },
      "name": "computeDispatchCriteria",
      "endEpochNanos": 1.7614761712541005e+18,
      "resource": {
        "schemaUrl": "https://opentelemetry.io/schemas/1.24.0",
        "attributes": {
          "data": [
            {
              "type": "STRING",
              "key": "host.arch",
              "keyUtf8": "aG9zdC5hcmNo"
            },
            "amd64",
            {
              "type": "STRING",
              "key": "host.name",
              "keyUtf8": "aG9zdC5uYW1l"
            },
            "pewpewpew",
            {
              "type": "STRING",
              "key": "os.description",
              "keyUtf8": "b3MuZGVzY3JpcHRpb24="
            },
            "Windows 11 10.0",
            {
              "type": "STRING",
              "key": "os.type",
              "keyUtf8": "b3MudHlwZQ=="
            },
            "windows",
            {
              "type": "STRING",
              "key": "process.command_line",
              "keyUtf8": "cHJvY2Vzcy5jb21tYW5kX2xpbmU="
            },
[...]

PR #1808, opened 2025-10-26
Optimisation: Add a SpanData dto to remove the resource field sent to the user

Apoorva64:
See #1807

### Description

This PR creates a DTO of the SpanData to omit the repetitive resource field

A Span now looks like
```
    {
      "name": "computeDispatchCriteria",
      "attributes": {
        "explain-trace": true
      },
      "events": [
        {
          "name": "dispatch_criteria_computed",
          "attributes": {
            "data": [
              {
                "type": "STRING",
                "key": "dispatch.criteria",
                "keyUtf8": "ZGlzcGF0Y2guY3JpdGVyaWE="
              },
              "/name=Millefeuille",
              {
                "type": "STRING",
                "key": "dispatcher",
                "keyUtf8": "ZGlzcGF0Y2hlcg=="
              },
              "URI_PARTS",
              {
                "type": "STRING",
                "key": "dispatcher.rules",
                "keyUtf8": "ZGlzcGF0Y2hlci5ydWxlcw=="
              },
              "name",
              {
                "type": "STRING",
                "key": "message",
                "keyUtf8": "bWVzc2FnZQ=="
              },
              "Computed dispatch criteria using URI_PARTS dispatcher"
            ],
            "empty": false
          },
          "epochNanos": 1.761478323771033e+18,
          "totalAttributeCount": 4,
          "droppedAttributesCount": 0
        }
      ],
      "links": [],
      "spanContext": {
        "traceId": "f0403307cd19ec6a68246cd94a5c9f62",
        "spanId": "5ceb202ad4914bc2",
        "traceFlags": {
          "sampled": true
        },
        "traceState": {
          "empty": true
        },
        "remote": false,
        "valid": true,
        "sampled": true,
        "traceIdBytes": "8EAzB80Z7GpoJGzZSlyfYg==",
        "spanIdBytes": "XOsgKtSRS8I="
      },
      "kind": "INTERNAL",
      "status": {
        "statusCode": "UNSET",
        "description": ""
      },
      "spanId": "5ceb202ad4914bc2",
      "traceId": "f0403307cd19ec6a68246cd94a5c9f62",
      "instrumentationScopeInfo": {
        "name": "io.github.microcks.web.RestInvocationProcessor",
        "attributes": {
          "data": [],
          "empty": true
        }
      },
      "totalRecordedLinks": 0,
      "endEpochNanos": 1.7614783237710418e+18,
      "totalRecordedEvents": 1,
      "parentSpanId": "1cba8440104ed47c",
      "startEpochNanos": 1.76147832377088e+18,
      "parentSpanContext": {
        "traceId": "f0403307cd19ec6a68246cd94a5c9f62",
        "spanId": "1cba8440104ed47c",
        "traceFlags": {
          "sampled": true
        },
        "traceState": {
          "empty": true
        },
        "remote": false,
        "valid": true,
        "sampled": true,
        "traceIdBytes": "8EAzB80Z7GpoJGzZSlyfYg==",
        "spanIdBytes": "HLqEQBBO1Hw="
      },
      "totalAttributeCount": 1,
      "instrumentationLibraryInfo": {
        "name": "io.github.microcks.web.RestInvocationProcessor"
      }
    },
```
The size has been reduce by a factor of 2

### Related issue(s)

#1807 and #1823

### DIFF
--- a/webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
+++ b/webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
@@ -150,11 +150,11 @@ public class SpanStorageService {
     */
    public List<String> queryTraceIdsByPatterns(String serviceName, String operationName, String clientAddress) {
       // Copy current spans to avoid null pointer exceptions during filtering/sorting
-     Map<String, List<SpanData>> snapshot;
-     synchronized (spansByTraceId) {
-       snapshot = new HashMap<>(spansByTraceId);
-     }
-     return snapshot.entrySet().stream()
+      Map<String, List<SpanData>> snapshot;
+      synchronized (spansByTraceId) {
+         snapshot = new HashMap<>(spansByTraceId);
+      }
+      return snapshot.entrySet().stream()
             .map(entry -> SpanFilterUtil.extractTraceEvent(entry.getKey(), entry.getValue()))
             .filter(event -> SpanFilterUtil.matchesTraceEvent(event, serviceName, operationName, clientAddress))
             .map(TraceEvent::traceId)


webapp/src/main/java/io/github/microcks/service/SpanDataDTO.java
```diff
@@ -0,0 +1,158 @@
...
+public class SpanDataDTO implements SpanData {
+   // Essential fields for trace analysis
+   private final String name;
+   private final SpanKind kind;
+   private final SpanContext spanContext;
+   private final SpanContext parentSpanContext;
+   private final StatusData status;
+   private final long startEpochNanos;
+   private final long endEpochNanos;
+   private final Attributes attributes;
```
lbroudoux, line 45: Have you checked these one don't hold a reference on delete `SpanData` so that we're sure delegate and unwanted fields values will be garbage collected?

Apoorva64: i am doing some benchmarking to check if this change has a memory impact.
This PR also fixes https://github.com/microcks/microcks/issues/1823

Apoorva64:
1.13.0
before bench:
image


after bench (filled):
image



Current branch

before bench:
image

after bench (filled):
image

The change does not seem to have the expected memory reduction

The benchmark that was used:

```javascript
import http from 'k6/http';
import { check, sleep } from 'k6';

export let options = {
  // total number of requests (one request per iteration)
  iterations: 10000,

  // concurrency: number of virtual users running in parallel.
  // Increase for higher concurrency, decrease to throttle.
  vus:  20,

  // optional thresholds to fail the test if too many errors/slow responses
  thresholds: {
    'http_req_failed': ['rate<0.01'], // <1% errors allowed
    'http_req_duration': ['p(95)<1000'] // 95% requests < 1000ms
  }
};

// target URL
const URL = 'http://localhost:8080/rest/FIWARE-NGSI+v2/1.0/v2';

export default function () {
  const res = http.get(URL, {
    headers: {
      'Accept': 'application/json',
    }
  });

  check(res, {
    'status is 2xx': (r) => r.status >= 200 && r.status < 300,
  });

  // small sleep to avoid hammering too tightly; remove or change if you want max throughput
  sleep(0.1);
}
```

I think there must be some java black magic somewhere

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -84,21 +84,34 @@ public void storeSpan(ReadableSpan span) {
...
-      if (spans.size() > MAX_SPANS_PER_TRACE) {
-         // Remove oldest span to cap memory
-         spans.removeFirst();
+      // Add the span to our collection. Use a synchronized list for per-trace spans so concurrent
+      // additions/removals are thread-safe. Also synchronize on the list when doing size+remove
+      // operations to make them atomic.
+      List<SpanData> spans = spansByTraceId.computeIfAbsent(traceId,
+            k -> Collections.synchronizedList(new ArrayList<>()));
+
+      synchronized (spans) {
```
lbroudoux: I'm not sure this synchronization block is required as here you're just calling unitary `add()` and `remove()` that are already synchronized due to line 91. To me, an additional synchronized block is only necessary when iterating over many elements.

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -81,26 +82,40 @@ public SpanStorageService(ApplicationEventPublisher publisher) {
...
+      // synchronizedMap's keySet must be done while holding the map lock to avoid
+      // ConcurrentModificationException.
+      synchronized (spansByTraceId) {
+         if (spansByTraceId.size() > MAX_TRACES) {
+            // Remove oldest trace - LinkedHashMap preserves insertion order
+            var it = spansByTraceId.keySet().iterator();
+            if (it.hasNext()) {
+               it.next();
+               it.remove();
+            }
```
lbroudoux: I would have keep something like:
```java
// Limit total number of traces to prevent memory leaks
      if (spansByTraceId.size() > MAX_TRACES) {
         // Remove oldest trace this works because LinkedHashMap maintains insertion order
         synchronized (spansByTraceId) {
            String oldestTraceId = spansByTraceId.keySet().iterator().next();
            spansByTraceId.remove(oldestTraceId);
         }
      }
```

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -136,13 +151,21 @@ public void clearAll() {
     */
    public List<String> queryTraceIdsBySpanAttributes(Map<AttributeKey<?>, Object> requiredAttributes) {
       if (requiredAttributes == null || requiredAttributes.isEmpty()) {
-         return new ArrayList<>(spansByTraceId.keySet());
+         synchronized (spansByTraceId) {
+            return new ArrayList<>(spansByTraceId.keySet());
```
lbroudoux: I don't think keySet() needs to be synchonized as it is an unitary method with no iteration.

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -136,13 +151,21 @@ public void clearAll() {
...
+         }
+      }
+
+      // Make a snapshot copy of entries to avoid holding the map lock for the whole
+      // streaming/processing and to prevent ConcurrentModificationException.
+      List<Map.Entry<String, List<SpanData>>> entries;
+      synchronized (spansByTraceId) {
+         entries = new ArrayList<>(spansByTraceId.entrySet());
```
lbroudoux: I don't think entrySet() needs to be synchronized as it is a unitary method with no iteration.

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -158,8 +181,12 @@ public List<String> queryTraceIdsBySpanAttributes(Map<AttributeKey<?>, Object> r
     *         first)
     */
    public List<String> queryTraceIdsByPatterns(String serviceName, String operationName, String clientAddress) {
-      return spansByTraceId.entrySet().stream()
-            .map(entry -> SpanFilterUtil.extractTraceEvent(entry.getKey(), entry.getValue()))
+      // Snapshot entries to avoid concurrent modification while streaming
+      List<Map.Entry<String, List<SpanData>>> entries;
+      synchronized (spansByTraceId) {
+         entries = new ArrayList<>(spansByTraceId.entrySet());
```
lbroudoux: I don't think entrySet() needs to be synchronized as it is a unitary method with no iteration.

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
```diff
@@ -185,16 +212,18 @@ public static boolean valuesEqualAttr(Object actual, Object expected) {
     * @return Set of trace IDs
     */
    public Set<String> getAllTraceIds() {
-      return spansByTraceId.keySet();
+      synchronized (spansByTraceId) {
+         return Set.copyOf(spansByTraceId.keySet());
```
lbroudoux: I don't think entrySet() needs to be synchronized as it is a unitary method with no iteration. Also, why would we do a copy here?

lbroudoux:
Hello, 

I've built and run a micro-benchmark on my side and see the effects of not saving the `resource` attribute of the span.

So basically, the process is as follow:
1. Start a test with fixed memory ergonomics (-Xms and -Xmx)
2. Run the GC multiple times (3) to be sure we evict short-lived objects from the heap
3. Save the `Runtime.getRuntime().freeMemory()` as "start free mem"
4. Build and store 60000 spans (with or without resource attributes)
5. Run the GC multiple times (3) to be sure we evict short-lived objects from the heap
6. Save the `Runtime.getRuntime().freeMemory()` as "end free mem"
7. Deduced memory used for storing the spans

Here are the results I get:
* With `resource` attributes       - time: `30 ms`, mem used: `10075 KB`
* Without `resource` attributes - time: `34 ms`, mem used: `6175 KB`

So there's a little time penalty in filtering the span (re-creating and copying to another object to do so) but there's a nice memory usage enhancement here.

> Note that for this test, I didn't change the implementation of the `spansByTraceId` but just put `synchronized` blocks on iterations (not unitary method calls).

Apoorva64:
@lbroudoux i have removed the unnecessary synchronized blocks. I also removed the method queryTraceIdsBySpanAttributes as it wasn't used.

Btw: What tool did you use to setup the test? For exemple to trigger the GC?

lbroudoux:
> Btw: What tool did you use to setup the test? For exemple to trigger the GC?

IT was all hand-written in pure Java ;-) And it just calls `System.gc()` multiple times to guarantee objects move from one heap space to another, and finally become evicted and memory reclaimed.

lbroudoux:
@Apoorva64 For some reason I don't explain, I had some comments that were still in a pending review... I think you can remove even more synchronized blocks or switch back to a simple unitary command that makes them unnecessary. Can you check again please? 

BTW Sorry for the trouble.

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
Apoorva64: we need the synchronized here as in compareSpansByEndTime we get the traces from the IDs. If the trace  has been removed since, we get a null pointer exception

lbroudoux: OK got it. Wouldn't it be better to handle the null pointer case or to proceed with a copy of the original entry set? (more memory but fewer synchronization)

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
Apoorva64: fixed in 249541a0417105833453ed566b298925524d075f

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
Apoorva64: fixed in 31931a57193b7867a58b35b28f49fd411f0d7cd8

webapp/src/main/java/io/github/microcks/service/SpanStorageService.java
Apoorva64: i have replaced it with a copy. But we still need the sync as the copy of the hashmap iterates over the entryset.

```java
java.util.ConcurrentModificationException: null
	at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:1024)
	at java.base/java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1059)
	at java.base/java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:1056)
	at java.base/java.util.HashMap.putMapEntries(HashMap.java:520)
	at java.base/java.util.HashMap.<init>(HashMap.java:493)
	at io.github.microcks.service.SpanStorageService.queryTraceIdsByPatterns(SpanStorageService.java:154)
```

Apoorva64:
@lbroudoux thanks a lots for taking your time to review and giving advice on this PR, I learned a lot about java and memory management xD

Issue #1042, opened 2024-01-15
Use external value to serve non-utf8 encodable data


Apoorva64:
### Reason/Context

Why we need this improvement?
To be able to serve PDF, Images, ByteStreams in Microcks
How will this change help?
This will greatly open what microcks can do in terms of REST API Mocking
What is the motivation?
I have an API that returns PDFs but the exemples cannot be encoded in an OpenApi file.

### Description

Upon receving a request.
Microcks gets the value of the "externalValue" field in an OpenApi specification
Does the request to the url of the external value
Override headers of the response and return it.

see https://swagger.io/docs/specification/adding-examples/#external

lbroudoux:
This is being discussed here in this thread: #1093

lbroudoux:
Now resolved with the merge of #1627. Thanks a lot to @Apoorva64!

PR #1844, opened 2025-12-01
Feat: add relative and remote data handling in external value

Apoorva64:
### Description

- This PR add a way to handle relative an remote refs in externalValue

For example we can now do:

```yaml
openapi: 3.0.3
info:
   title: Get Image Api
   version: 1.0.0

paths:
   /image/{id}:
      get:
         summary: Get an Image By Id
         parameters:
            -  name: id
               in: path
               required: true
               description: ID of the image to retrieve
               schema:
                  type: string
               examples:
                  EXAMPLE1:
                     value: 1
         responses:
            '200':
               description: The resized image
               content:
                  image/png:
                     schema:
                        type: string
                        format: binary
                     examples:
                       EXAMPLE1:
                         externalValue: ./microcks-banner.png
            '404':
               description: Image not found
            '500':
               description: Server error
```

### Related issue(s)
#1042

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/ReferenceResolver.java
+++ b/webapp/src/main/java/io/github/microcks/util/ReferenceResolver.java
@@ -161,7 +161,7 @@ public class ReferenceResolver {
          return DataUriUtil.buildDataUri(contentBytes);
       }
 
-      return Files.readString(referenceFile.toPath(), encoding);
+      return new String(contentBytes, encoding);
    }
 
    /**
--- a/webapp/src/main/java/io/github/microcks/util/UTF8ContentTypeChecker.java
+++ b/webapp/src/main/java/io/github/microcks/util/UTF8ContentTypeChecker.java
@@ -15,6 +15,9 @@
  */
 package io.github.microcks.util;
 
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CoderResult;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
 import java.nio.charset.StandardCharsets;
@@ -37,8 +40,10 @@ public class UTF8ContentTypeChecker {
       return matcher.matches();
    }
 
+
    /**
-    * Checks if the given byte array is valid UTF-8.
+    * Checks if the given byte array is valid UTF-8. Uses incremental decoding with a small buffer to minimize memory
+    * allocation while still failing early on invalid input.
     * @param data The byte array to check.
     * @return true if valid UTF-8, false otherwise.
     */
@@ -47,9 +52,19 @@ public class UTF8ContentTypeChecker {
       decoder.onMalformedInput(CodingErrorAction.REPORT);
       decoder.onUnmappableCharacter(CodingErrorAction.REPORT);
 
+      ByteBuffer in = ByteBuffer.wrap(data);
+      CharBuffer out = CharBuffer.allocate(1024);
+
       try {
-         decoder.decode(java.nio.ByteBuffer.wrap(data));
-         return true;
+         while (in.hasRemaining()) {
+            CoderResult result = decoder.decode(in, out, true);
+            if (result.isError()) {
+               return false;
+            }
+            out.clear();
+         }
+         CoderResult result = decoder.flush(out);
+         return !result.isError();
       } catch (Exception e) {
          return false;
       }


webapp/src/test/java/io/github/microcks/util/openapi/OpenAPIImporterTest.java
```diff
@@ -2116,33 +2115,78 @@ void testOpenAPIWithExternalValueDataUriBinaryResponse() {
...
-                  "Response should contain base64-encoded PNG data or data URL");
-         } else {
-            fail("Exchange has the wrong type. Expecting RequestResponsePair");
-         }
+   }
+
+   @Test
+   void testOpenAPIWithRemoteExternalValueBinaryResponse() {
+      ReferenceResolver resolver = new ReferenceResolver(
+            "https://raw.githubusercontent.com/StartUpNationLabs/microcks/refs/heads/feat/add-relative-and-remote-data-in-external-value/webapp/src/test/resources/io/github/microcks/util/openapi/get-image-openapi-remote-externalvalue.yaml",
```
Apoorva64, line 2148: We need to change this test to use the microcks main repository

Apoorva64:
@lbroudoux what are your thoughts on this PR?

lbroudoux:
> @lbroudoux what are your thoughts on this PR?

Hey, I'm just back from a 2-weeks break. Looking at it now.

webapp/src/main/java/io/github/microcks/util/ReferenceResolver.java
```diff
@@ -154,6 +154,12 @@ public String getReferenceContent(String relativePath, Charset encoding) throws
       }
       // Keep track on how we resolved this relativePath.
       relativeResolvedReferences.put(relativePath, referenceFile);
+      // Check for binary content.
+      byte[] contentBytes = Files.readAllBytes(referenceFile.toPath());
+      if (!UTF8ContentTypeChecker.isValidUTF8(contentBytes)) {
+         // If binary, return a data URI representation.
+         return DataUriUtil.buildDataUri(contentBytes);
+      }
```
lbroudoux, line 162: This one requires loading the whole file again and can be memory/time-intensive. Isn't there a better way to do this?

webapp/src/main/java/io/github/microcks/util/UTF8ContentTypeChecker.java
```diff
@@ -29,4 +32,22 @@ public static boolean isUtf8Encodable(String contentType) {
       Matcher matcher = UTF8_ENCODABLE_PATTERN.matcher(contentType.trim());
       return matcher.matches();
    }
+
+   /**
+    * Checks if the given byte array is valid UTF-8.
+    * @param data The byte array to check.
+    * @return true if valid UTF-8, false otherwise.
+    */
+   public static boolean isValidUTF8(byte[] data) {
```
lbroudoux, line 50: We have some naming inconsistencies with the above `isUtf8Encodable()`. I think we should rename above to `isUTF8Encodable()`. What to you think?

lbroudoux (changes requested):
I'm a bit concerned by encoding detection overhead... I think there'd might be a better way to do this but not sure how....

webapp/src/main/java/io/github/microcks/util/UTF8ContentTypeChecker.java
Apoorva64: fixed in 4ce4d5768b7d18597f2d7ec62c8b6e8fc94e47c0

webapp/src/main/java/io/github/microcks/util/ReferenceResolver.java
Apoorva64: i've looked a bit into it.
First of all the `getReferenceContent` methods returns a String so the file has to be loaded in memory at least once.
In the case of a binary file (case 1):
- The file is first loaded with `readAllBytes` in the byte array 'contentBytes'.
- Then in the `isValidUTF8` a char array of the size of 'contentBytes' is allocated in `CharsetDecoder.decode` as a buffer
- Then the final string is encoded in `buildDataUri`

In the case of a isValidUTF8 file (case 2):
- The file is first loaded with `readAllBytes` in the byte array 'contentBytes'.
- Then in the `isValidUTF8` a char array of the size of 'contentBytes' is allocated in `CharsetDecoder.decode` as a buffer
- Then the file is loaded for a second time with the `Files.readString`

Some optimization that we can do:
- Replace `Files.readString` with `new String(contentBytes, encoding)` instead of reading the file for a second time in case 2
- we can rewrite `isValidUTF8` so that it uses `CharsetDecoder.decode(ByteBuffer in, CharBuffer out, boolean endOfInput)` to reduce the buffer size

lbroudoux: Good points. Let's apply those 2 changes.

Apoorva64: done in 96a930694c75a024cf2ac94b9e81c5b79321d5fa

lbroudoux (approved):
LGTM. Let it roll!

Issue #1714, opened 2025-09-03
chore: Review and fix quality issues raised by Sonar on 1.13.x for 1.13.1


lbroudoux:
See https://sonarcloud.io/summary/new_code?id=microcks_microcks&branch=1.13.x and the main reference containing all issues: https://sonarcloud.io/summary/overall?id=microcks_microcks

Include some chore PR here if you'd like to contribute!

AnubPrasad:
I'm new to the project but would love to contribute. Let me know if there are any guidelines or pointers to help me get started.

rana-rajat:
Hi @lbroudoux,  I have made some code changes to fix the high-severity vulnerabilities reported by Sonar. Could you guide me on how I can verify whether these vulnerabilities have been resolved in my code?

lbroudoux:
> Hi [@lbroudoux](https://github.com/lbroudoux), I have made some code changes to fix the high-severity vulnerabilities reported by Sonar. Could you guide me on how I can verify whether these vulnerabilities have been resolved in my code?

Hi @Rajat !
I think you should be able to check using one of the SonarQube plugins for IDE (see https://docs.sonarsource.com/sonarqube-for-ide/vs-code/ for example)

lbroudoux:
> I'm new to the project but would love to contribute. Let me know if there are any guidelines or pointers to help me get started.

This project use standard Java rules coming from SonarQube with formatting rules that are enforced running `mvn spotless:apply` on your project. You can use one of SonarQube IDE plugin to get the list of issues, fix and check them. Then apply coding conventions using the `mvn spotless:apply` Maven command.

lbroudoux:
Let's move this one to `1.13.1`!

lbroudoux:
Closing this one just before the release of `1.13.1`. Will reopen another one for the next milestone.

PR #1847, opened 2025-12-03
 Fix SonarQube code smells - add private constructor and reduce regex complexity

Caesarsage:
Fixes two SonarQube issues in `UTF8ContentTypeChecker`:

1. Add private constructor to hide implicit public one
2. Simplify regular expression to reduce complexity from 21 to 20

## Changes
- Added private constructor to prevent instantiation of utility class
- Refactored regex pattern to reduce cognitive complexity while maintaining functionality
- All existing tests pass

image

image

 
### Related issue(s)
#1714


lbroudoux (approved):
LGTM but I have to admit I'm not 100% sure it's still working just by reading the regexp 😉  Let's see how the CI with tests goes! Thanks a lot!

PR #1851, opened 2025-12-04
fix: Enable support for specifying a PubSub emulator host

### LINKED ISSUE #1846: Enable Google PubSub Emulator
In normal PubSub emulator usage we would expect these two environment variables `PUBSUB_EMULATOR_HOST` and `PUBSUB_PROJECT_ID` to configure a client to use the emulator.

But, currently we cannot configure the async minion to publish to a pubsub emulator due to two issues:
- We cannot remove the default value for the configuration `googlepubsub.service-account-location`
- We need to specify a local channel as per the google guide [here](https://docs.cloud.google.com/pubsub/docs/emulator#accessing_environment_variables)

I propose that within `GooglePubSubProducerManager.construct` we detect the presence of the `PUBSUB_EMULATOR_HOST` variable, and if set, configure the client to connect to the emulator, otherwise we use the service account and the cloud client.

ghost:
### Description

- Use `PUBSUB_EMULATOR_HOST` on the PubSub producer to configure for an emulator
- Use a host name in the endpointUrl spec for the PubSub consumer to consume from an emulator

### Related issue(s)

Fixes #1846

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/consumer/GooglePubSubMessageConsumptionTask.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/consumer/GooglePubSubMessageConsumptionTask.java
@@ -56,13 +56,19 @@ public class GooglePubSubMessageConsumptionTask implements MessageConsumptionTas
    private final Logger logger = Logger.getLogger(getClass());
 
    /** The string for Regular Expression that helps validating acceptable endpoints. */
-   public static final String ENDPOINT_PATTERN_STRING = "googlepubsub://((?<emulator>[^:]+(:\\d+)?)/)?(?<projectId>[a-zA-Z0-9-_]+)/(?<topic>[a-zA-Z0-9-_\\.]+)(\\?(?<options>.+))?";
+   public static final String ENDPOINT_PATTERN_STRING = "googlepubsub://(?<projectId>[a-zA-Z0-9-_]+)/(?<topic>[a-zA-Z0-9-_\\.]+)(\\?(?<options>.+))?";
    /** The Pattern for matching groups within the endpoint regular expression. */
    public static final Pattern ENDPOINT_PATTERN = Pattern.compile(ENDPOINT_PATTERN_STRING);
 
    private static final String CLOUD_OAUTH_SCOPE = "https://www.googleapis.com/auth/cloud-platform";
 
    private static final String SUBSCRIPTION_PREFIX = "-microcks-test";
+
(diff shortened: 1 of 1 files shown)


minions/async/src/main/java/io/github/microcks/minion/async/consumer/GooglePubSubMessageConsumptionTask.java
```diff
@@ -51,7 +56,7 @@ public class GooglePubSubMessageConsumptionTask implements MessageConsumptionTas
    private final Logger logger = Logger.getLogger(getClass());
 
    /** The string for Regular Expression that helps validating acceptable endpoints. */
-   public static final String ENDPOINT_PATTERN_STRING = "googlepubsub://(?<projectId>[a-zA-Z0-9-_]+)/(?<topic>[a-zA-Z0-9-_\\.]+)(\\?(?<options>.+))?";
+   public static final String ENDPOINT_PATTERN_STRING = "googlepubsub://((?<emulator>[^:]+(:\\d+)?)/)?(?<projectId>[a-zA-Z0-9-_]+)/(?<topic>[a-zA-Z0-9-_\\.]+)(\\?(?<options>.+))?";
```
lbroudoux: I would prefer managing the emulator stuff with an addition option like we did for LocalStack on AWS.

If you look [here](https://github.com/microcks/microcks/blob/master/minions/async/src/main/java/io/github/microcks/minion/async/consumer/AmazonSNSMessageConsumptionTask.java#L74), you'll see that we have defined an `overrideUrl` option. Here in the case of PubSub, it could be `emulatorHost`.

This allows to keep consistent URIs:
* `googlepubsub://my-project-test/my-topic`
* `googlepubsub://my-project-test/my-topic?emulatorHost=emulator:8585`

When initializing connection and channels, you can then use the `hasOption(EMULATOR_HOST_OPTION)` like we did [here](https://github.com/microcks/microcks/blob/master/minions/async/src/main/java/io/github/microcks/minion/async/consumer/AmazonSNSMessageConsumptionTask.java#L189) to create the additional channel provider.

ghost: ok, i don't have time today, but i'll get these changes done tomorrow 👍

ghost: done (i think)

Issue #1866, opened 2025-12-08
Feature Request [Import/UI]: Support drag-and-drop of URLs to automatically open the file import dialog


Apoorva64:
### Reason/Context

Microcks already supports a very convenient feature: **drag-and-drop of files anywhere in the UI** to trigger the import dialog.
However, in many workflows, users are not importing files from their local machine they are importing via URLs (internal repositories, GitHub raw links, public API specs, etc.).

Today, users must manually:
1. Copy the link,
2. Open the import dialog,
3. Click on "Download Artifact"
4. Paste it into the “From URL” field.
5. Click import

#### Why do we need this improvement?
* Reduces friction in importing workflow

### Description

Extend the existing drag-and-drop system to handle *dropped URLs* in addition to files.

1. User drags a link (e.g. from browser address bar, GitHub link, S3 link, etc.) onto the Microcks UI.
2. Microcks detects the dropped item is a URL (not a file).
3. Microcks automatically:
   * Opens the **Import → Download Artifact** modal,
   * Prefills the URL text field with the dropped link.
4. User just clicks *“Import”*.

i think we could also extend the importer to handle *pasted* files and urls

lbroudoux:
Sounds cool! I didn't know it was possible to detect that the drag and dropped element is a text/URL and not a file!

lbroudoux:
Available with the merge of #1871

PR #1871, opened 2025-12-09
feat: add drag n drop for urls and add copy-n-paste userflow

Apoorva64:
### Description
This PR adds:
- Drag-n-Drop and Copy-n-Paste user flows for "Download" specifications
  - Which means that we can now drag and drop (or paste them) urls directly onto the ui which will open the import window on the Download tab
- Copy-n-Paste user flow for the files

### Related issue(s)

#1866

Drag-n-drop url:

https://github.com/user-attachments/assets/96d9dc78-cf8f-414f-b1b4-f65329f8aa85


lbroudoux:
A quick question: as the listened event is `paste`, does it work when simply copy/pasting (cmd+c/cmd+v) a URL as a text in the UI?

Apoorva64:
> A quick question: as the listened event is `paste`, does it work when simply copy/pasting (cmd+c/cmd+v) a URL as a text in the UI?

Yes it works with paste also

Issue #1744, opened 2025-09-17
Feature Request: Add native support for Bruno collections in Microcks


sgkiokas:
### Reason/Context

Currently, Microcks supports Postman collections but not Bruno collections directly. While Bruno offers an export-to-Postman option, this conversion is incomplete and lacks support for important features such as request/response events or scripts that define dynamic mocking logic.

For example, in Bruno a pre-request event might contain logic for generating payloads, computing signatures, or setting variables dynamically. When exported to Postman, these events are not properly converted into Postman-compatible scripts. This leads to broken or missing functionality in Microcks when working with these collections.

This improvement would help teams who already use Bruno as their primary API client to integrate with Microcks seamlessly, without having to rely on lossy conversions. It would also reduce the overhead of manually rewriting or revalidating event scripts after export, which is time-consuming and error-prone.

The motivation is to extend Microcks’ usability to a broader community of API practitioners who are adopting Bruno as a lightweight, modern alternative to Postman, and to ensure that dynamic mocking scenarios defined in Bruno collections are preserved and respected when imported into Microcks.

### Description

The feature request is to add native support for Bruno collections in Microcks, alongside the existing Postman support.

At the moment, Bruno collections can only be imported into Microcks by first exporting them as Postman collections. However, this process does not handle advanced features such as event blocks, pre-request/post-response scripts, or Bruno-specific variable resolution. For example, Bruno’s event system allows dynamic payload generation and cryptographic signing of requests, but these scripts are either ignored or exported in a format that is not Postman-compatible.

The proposed change would allow Microcks to:

- Import Bruno collections directly, without requiring a Postman export step.
- Preserve event scripts, variables, and dynamic mocking logic defined in Bruno collections.
- Provide parity with Postman support so users can switch between the two tools without losing functionality.

This would not be a breaking change, as existing Postman collection support remains unaffected. Instead, it would broaden the scope of supported tools and reduce friction for teams already using Bruno.

lbroudoux:
Hi @sgkiokas and thanks for raising this feature request/discussion! 

There are 2 use-cases of seeing Microcks <> Bruno integration as an alternative to what we're doing with Postman:
* The mocking side involves gathering request/response pairs examples from a Bruno collection - and it seems to be an opened issue here https://github.com/usebruno/bruno/issues/1235 in the Bruno repo,

[...]

PR #1888, opened 2025-12-24
Added initial set of classes to import open collection

krisrr3:
### Description

- Enabling opencollection
- ...
- ...


lbroudoux:
Thanks a lot for pushing this!

A few questions and comments:
* How is it related to #1744? Does Bruno already use Open Collection as its internal export format?
* As far as I understand it, OpenCollection does not seem to support the concept of examples for the moment but Bruno claims to have it (see [here](https://github.com/usebruno/bruno/issues/1235#issuecomment-3511102988)). Isn't it a mismatch between Open Collection and Bruno format?
* The current development branch is `1.13.x`. Please start your PR from this one and target this one if you plan to finalize it for `1.13.2` or the next ones
* The DCO check fails as the commit is not signed. Could you please ensure that you sign it correctly? This is a prerequisite for the CNCF projects.

Issue #1895, opened 2026-01-04
Add error handling in TestController.java


PranjaliBhardwaj:
I saw a TODO comment: decide whether to return an error when a secret is specified but not found? in TestController.java
work: Add validation to return a clear error if a secret is required but missing.

lbroudoux:
Now done with the merge of #1908 
Thanks to @PranjaliBhardwaj! 🎉

PR #1908, opened 2026-01-11
[TODO/ENHC]: Error handling in TestController.java implemented validation

PranjaliBhardwaj:
[ENHANCEMENT]
issue: https://github.com/microcks/microcks/issues/1895

Description
Added validation: If a secret name is provided but not found in the repository, the method now returns an error instead of continuing with null.
Clear error message: Returns HTTP 400 (BAD_REQUEST) with: "Secret '{secretName}' not found. Please ensure the secret exists before creating the test."
Logging: Added a warning log when a secret is specified but not found for debugging.
Removed TODO: The TODO comment is resolved by implementing the validation.

The implementation:
Validates that if a secret name is provided, it must exist
Returns a clear, actionable error message
Uses appropriate HTTP status code (400 BAD_REQUEST)
Follows existing error handling patterns in the codebase
When a test request includes a secret name that doesn't exist, the API will return a clear error instead of silently proceeding.

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/SafeLogger.java
+++ b/webapp/src/main/java/io/github/microcks/util/SafeLogger.java
@@ -167,6 +167,16 @@ public class SafeLogger {
    }
 
    /**
+    * Log a message at the WARN level.
+    * @param message the message string to be logged
+    */
+   public void warn(String message) {
+      if (log.isWarnEnabled()) {
+         log.warn(encode(message));
+      }
+   }
+
+   /**
     * Log an exception (throwable) at the WARN level with an accompanying message.
     * @param message the message string to be logged
     * @param t       the exception (throwable) to log


PranjaliBhardwaj:
@lbroudoux new pr created, i closed old ones with changes

webapp/src/main/java/io/github/microcks/web/TestController.java
```diff
@@ -124,8 +124,13 @@ public ResponseEntity<TestResult> createTest(@RequestBody TestRequestDTO test) {
          List<Secret> secrets = secretRepository.findByName(test.getSecretName());
          if (!secrets.isEmpty()) {
             secretRef = new SecretRef(secrets.getFirst().getId(), secrets.getFirst().getName());
+         } else {
+            log.warn("Secret '{}' specified in test request but not found", test.getSecretName());
```
lbroudoux: You must return a `ResponseEntity<TestResult>` here. As a consequence, the build doesn't compile. You'd probably have to change the signature of the method so that we return `ResponseEntity<TestResult>` in case we launched the test or `ResponseEntity<String>` when something is wrong. Setting `ResponseEntity<Object>` should be ok.

lbroudoux:
> Microcks uses JBoss Logging, which does not support SLF4J-style {} placeholders.

In fact, it's a bit more complex 😉  JBoss Logging is used in the `async-minion` that is written in  Quarkus but the main app component (`webapp`) is written in Spring and uses SL4J.

Moreover, in the `webapp` component, we've wrapped common SL4J with a `SafeLogger` to prevent logging user-entered input (a security best practice). The remaining error is because the `warn()` method with the signature you chose doesn't exist on `SafeLogger`.

Actually, it's better to check that your PR locally compiles before submitting them...

PranjaliBhardwaj:
ok, new approach to handle this problem: 
The Problem
SafeLogger only had warn(String message, Throwable t), not warn(String message). When TestController called log.warn("Secret '" + test.getSecretName() + "' specified..."), the compiler tried to match warn(String, Throwable) and failed because a String can't be converted to Throwable.
The Solution
Added the missing warn(String message) method to SafeLogger, matching the pattern used for debug(), info(), and error(). This allows TestController to call log.warn() with a single string argument.

As I was learning the project, I may have made a few mistakes. I hope they did not come across negatively.
Thank you for mentoring me @lbroudoux  and @dipak0000812

PR #1965, opened 2026-02-19
fix(ui): add missing SOAPAction header when generating SOAP curl commands

### LINKED ISSUE #1952: Add SOAPAction in cURL content-type
### Describe the bug

When clicking on a generated curl command for a SOAP operation in Microcks, the SOAPAction in "content-type" header is never present. However, this header is required for some SOAP servers to correctly process the request. It would be helpful to either add it automatically or allow the user to specify it, to ensure SOAP calls work as expected.

### Expected behavior

The generated curl command for SOAP operations should always include the appropriate SOAPAction in content-type header, or provide a way for the user to specify it, to ensure compatibility and avoid errors during development phase.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

_No response_

_No response_

### Additional information

_No response_

dipak0000812:
Fixes #1952

This PR ensures that generated curl commands for SOAP operations include the required action header when it is missing.

Behavior:
- SOAP 1.1: adds `SOAPAction: "actionUri"` if missing.
- SOAP 1.2: appends `; action="actionUri"` to `Content-Type: application/soap+xml` if missing.
- Existing headers are preserved and not duplicated.

The change is minimal, scoped to `formatCurlCmd()`, and only applies to `ServiceType.SOAP_HTTP`.

### DIFF
--- a/webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts
+++ b/webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts
@@ -879,11 +879,51 @@ export class ServiceDetailPageComponent implements OnInit {
 
       cmd = 'curl -X ' + verb + ' \'' + mockUrl + '\'';
 
+      const isSoapService =
+        this.resolvedServiceView.service.type === ServiceType.SOAP_HTTP;
+
+      let hasSoapAction = false;
+      let hasSoap12ContentType = false;
+
       // Add request headers if any.
       if (exchange.request.headers != null) {
         for (const header of exchange.request.headers) {
+          const headerName = header.name.toLowerCase();
+
+          if (isSoapService && headerName === 'soapaction') {
+            hasSoapAction = true;
+          }
+
+          if (
+            isSoapService &&
+            headerName === 'content-type' &&
+            header.values.some(v => v.includes('application/soap+xml'))
+          ) {
+            hasSoap12ContentType = true;
+
+            if (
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts
```diff
@@ -879,13 +879,34 @@ export class ServiceDetailPageComponent implements OnInit {
 
       cmd = 'curl -X ' + verb + ' \'' + mockUrl + '\'';
 
+      let hasSoapAction = false;
+      let hasSoap12ContentType = false;
+
       // Add request headers if any.
       if (exchange.request.headers != null) {
         for (const header of exchange.request.headers) {
+          if (header.name.toLowerCase() === 'soapaction') {
```
SebastienDegodez: The code has unnecessary complexity with repeated ServiceType.SOAP_HTTP checks and indexOf() instead of includes(). 

The flag detection logic is mixed with header iteration, making it harder to read and maintain. Suggestion : Extract helper methods and cache the service type check to improve code quality and testability.

dipak0000812: Refactored the SOAP header handling to cache the service type check and simplify detection logic as suggested. Replaced indexOf() with includes() and separated SOAP 1.1 fallback logic from header iteration. Please let me know if anything else should be adjusted.

lbroudoux:
Sorry ... I missed this one and it has now be closed because fied into another one.

PR #1968, opened 2026-02-23
fix: normalize SOAP carriage returns on import and UI rendering

### LINKED ISSUE #1951: Carriage return characters (\r) in SOAP payloads cause validation errors
### Describe the bug

When importing SOAP contracts or responses into Microcks, carriage return characters (\r) are sometimes present in the payloads. This causes unexpected anomalies or validation errors during contract testing, as Microcks does not handle these characters properly.

### Expected behavior

Microcks should automatically clean or ignore carriage return characters (\r) during import or contract validation, so that SOAP payloads are processed correctly and contract tests do not fail due to these invisible characters.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

1.13.2

_No response_

### Additional information

_No response_

Caesarsage:
### Description

- Normalize \r\n / \r to \n during SoapUI import in SoapUIProjectImporter so persisted request/response payloads are clean.
- Add regression test testCRLFPayloadsAreNormalizedOnImport with a CRLF fixture SoapUI project.
- Add UI safety-net normalization in Service Detail display path using normalizeLineEndings from format-utils.

This prevents hidden carriage return anomalies in SOAP payload handling and display.

### Related issue(s)
Fixes #1951


Caesarsage:
cc @SebastienDegodez @lbroudoux

lbroudoux:
> cc @SebastienDegodez @lbroudoux 

I have the week off and will be back Monday 2nd.

webapp/src/main/java/io/github/microcks/util/soapui/SoapUIProjectImporter.java
```diff
@@ -728,7 +728,8 @@ private Response buildResponse(Element mockResponse, String dispatchCriteria)
          throws MockRepositoryImportException, Exception {
       Response response = new Response();
       response.setName(mockResponse.getAttribute(NAME_ATTRIBUTE));
-      response.setContent(getConfigUniqueDirectChild(mockResponse, "responseContent").getTextContent());
+      response.setContent(
+            normalizeLineEndings(getConfigUniqueDirectChild(mockResponse, "responseContent").getTextContent()));
```
lbroudoux: I would prefer not to change the imported content as it has to be "the source of truth". Couldn't we manage this only on client side?

webapp/src/main/java/io/github/microcks/util/soapui/SoapUIProjectImporter.java
```diff
@@ -743,7 +744,8 @@ private Request buildRequest(Element testRequest) throws MockRepositoryImportExc
       Request request = new Request();
       request.setName(testRequest.getAttribute(NAME_ATTRIBUTE));
       if (hasConfigDirectChild(testRequest, REQUEST_TAG)) {
-         request.setContent(getConfigUniqueDirectChild(testRequest, REQUEST_TAG).getTextContent());
+         request
```
lbroudoux: I would prefer not to change the imported content as it has to be "the source of truth". Couldn't we manage this only on client side?

webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts
```diff
@@ -832,22 +833,23 @@ export class ServiceDetailPageComponent implements OnInit {
   }
 
   public formatRequestContent(requestContent: string): string {
+    const normalizedContent = normalizeLineEndings(requestContent);
```
lbroudoux, line 836: Yes, if we're doing this here, there's no need to do it on server side IMO.

lbroudoux (changes requested):
Is it possible to only handle this on client side to keep the imported artifact as the source of truth? I wouldn't expect changes on server side importer.

SebastienDegodez:
> Is it possible to only handle this on client side to keep the imported artifact as the source of truth? I wouldn't expect changes on server side importer.

I just want to confirm: when you say client, do you specifically mean the JavaScript client?
If so, wouldn’t the same issue still occur when uploading through the REST API (e.g., via the test container), since the server-side importer wouldn’t change?

lbroudoux:
> I just want to confirm: when you say client, do you specifically mean the JavaScript client? If so, wouldn’t the same issue still occur when uploading through the REST API (e.g., via the test container), since the server-side importer wouldn’t change?

Yes, I mean on the JavaScript side only. IIRC, the problem we face is only for generating a correct CURL command - it does not impact mock features. So I would only change this part.

webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts
Caesarsage: okay. thanks

Caesarsage:
kindly take another look @lbroudoux

PR #1981, opened 2026-03-08
Fix/add explain tracing to operation service errors

### LINKED ISSUE #1950: Live Traces don't work if operation selection fails
### Describe the bug

When operation selection fails (ex: missing soap action header or bad resource path), it seems that no trace is actually stored and put into the global live traces panel. 

### Expected behavior

Of course, the trace cannot appear on per-operation panel because of wrong operation but it should appear on global per-service panel.

### Actual behavior

No Live traces at all.

### How to Reproduce?

_No response_

All

All

### Additional information

_No response_

Apoorva64:
### Description
Before this change, requests to non-existent services or unmatched operations returned HTTP errors silently with no trace data. This made it difficult to diagnose routing issues via the live-traces graph. 
This PR adds missing tracing in the Rest Controller and changes the Trace some of the Tracing logic to capture theses requests.

I also did a small refactor in the RestController to remove duplications see https://github.com/microcks/microcks/pull/1981/changes/dcaa90ce2fa84c6380570fdf91204a4da103b498

image

### Related issue(s)
Fixes #1950

### DIFF
--- a/webapp/src/test/java/io/github/microcks/web/TracingControllerIT.java
+++ b/webapp/src/test/java/io/github/microcks/web/TracingControllerIT.java
@@ -111,7 +111,7 @@ class TracingControllerIT extends AbstractBaseIT {
          assertThat(spansResponse.getBody().toString()).contains("/rest/pastry-details/1.0.0/pastry")
                .contains("processInvocation").contains("explain-trace")
                .contains("Selected dispatcher and rules for this invocation")
-               .contains("Received REST invocation GET /pastry");
+               .contains("Received REST invocation GET /rest/pastry-details/1.0.0/pastry");
       }
    }
 


Apoorva64:
There is an integration test to fix. I'll fix it tonight

lbroudoux:
> There is an integration test to fix. I'll fix it tonight

This looks good so far. Let me know when fix is pushed.

Apoorva64:
Sorry, for the late reply,
I changed the integration test from checking a message with the operation name to the URI of the resquest as we don't have the operation info when we do the "INVOCATION_RECIEVED" event
I think the AsyncAPITestManagerIT failing test is a flaky test as it works locally

Apoorva64:
@lbroudoux @SebastienDegodez are we good for this PR?

lbroudoux:
> @lbroudoux @SebastienDegodez are we good for this PR?

Thanks for your patience. Looking at it now.

PR #1991, opened 2026-03-13
OpenAI example generation and add gpt-5 family support [ #1989 #1990 ] 

### LINKED ISSUE #1989: [bug] AI Copilot bulk example generation
### Describe the bug

When using AI Copilot bulk sample generation on an API contract, the process may stop completely if generation for a single operation/response fails.

This makes the bulk generation workflow unreliable on medium or large contracts, because one malformed AI output can block processing of all remaining items.

### Why this is a bug

Bulk generation is expected to process a set of independent operations/responses. A failure on one item should not block processing of unrelated remaining items.

At the moment, a single invalid AI-generated sample can make the whole bulk workflow unusable, forcing manual restart/rework.

### Expected behavior

Bulk generation should be resilient to per-item failures.

Expected behavior:

1. If one item fails, Microcks should mark it as failed and continue with the next items.
2. Users should be able to resume an interrupted/failed bulk generation without reprocessing already successful items.
3. Users should be able to run generation only for missing examples.
4. Failed items should be optionally retried automatically.
5. On retry, Microcks should be able to include the previous parsing/validation error in the prompt so the model can avoid repeating the same mistake.

### Actual behavior

If one generated sample is invalid (for example malformed YAML/JSON or content that cannot be parsed/validated), the bulk generation flow becomes blocked from the user perspective.

As a result:

- remaining operations/responses are not processed,
- the job cannot be resumed from the last successful item,
- there is no option to skip the failed item and continue,
- there is no built-in retry strategy for failed generations.

### How to Reproduce?

_No response_

1.13.2

_No response_

### Possible directions for a fix

This issue could be addressed through one or more complementary improvements. These are not necessarily mutually exclusive and could also be combined:

1. **Per-item status tracking**
   Track generation status for each item independently, for example:
   - `pending`
   - `success`
   - `failed`
   - `skipped`

   This would make it possible to avoid treating the whole bulk generation as a single all-or-nothing process.

2. **Continue processing after single-item failure**
   If one item fails validation or parsing, skip it and continue generating the remaining items. 

3. **Resume interrupted generation**
   Allow users to restart a previously interrupted/failed bulk generation job and process only items still marked as `pending` or `failed`, without regenerating successful ones.

4. **Generate only missing examples**
   Add a mode that only targets operations/responses that currently do not already have examples, skipping the ones already populated.

5. **Retry failed items**
   Add a retry mode for failed items only, instead of forcing users to rerun the whole bulk generation.

6. **Retry with previous error context**
   When retrying a failed item, optionally include the previous parsing/validation error in the next prompt, so the model has a chance to correct the specific issue instead of repeating it.

   Example:

   > The previous output failed with this error: `<error message>`.  
   > Please regenerate a valid example and avoid this issue.

7. **Configurable retry behavior**
   Optionally expose controls such as:
   - `maxRetries`
   - retry only on parsing/validation errors
   - `includePreviousErrorInPrompt=true`

Any subset of the above would already improve the reliability of AI Copilot bulk generation, and the best solution may combine several of them.

### LINKED ISSUE #1990: [bug] OpenAI gpt-5 model support
### Describe the bug

When AI Copilot is configured to use GPT-5 family models, the OpenAI request payload appears to include parameters that are not supported for that model/API combination, such as `max_completion_tokens` and/or `logit_bias`.

This causes request failures and prevents AI Copilot from working correctly with GPT-5 models.

### Expected behavior

AI Copilot should build the request payload according to the selected OpenAI API style and model family.

For GPT-5 family models, Microcks should avoid sending unsupported parameters and should use the correct token limit parameter for the target API:

- use only parameters supported by the selected endpoint/model,
- avoid sending `logit_bias` if unsupported,
- avoid sending `max_completion_tokens` if the configured endpoint expects a different field,
- use the proper output token field depending on the API being used.

### Actual behavior

When using a GPT-5 family model in AI Copilot, the generated OpenAI-compatible request may include parameters that are not valid for the selected model or endpoint.

In practice, this results in request errors instead of successful sample generation.

### How to Reproduce?

_No response_

1.13.2

_No response_

### Additional information

OpenAI documentation differentiates parameter usage depending on the API surface. The Responses API uses `max_output_tokens`, while Chat Completions documents `max_completion_tokens` and also notes compatibility constraints for newer reasoning models.

Because of this, AI Copilot should not assume that legacy/default OpenAI parameters are universally valid for GPT-5 family models.

paulux84:
### Description

  - Refactor OpenAI sample generation to retry when the returned completion cannot be parsed, reusing the previous invalid output and parser error to ask for a corrected example instead of failing immediately.
  - Make AI sample generation more resilient at service level by skipping empty generations and logging per-operation failures without aborting the whole asynchronous run.
  - Add compatibility for GPT-5 family models by using `max_completion_tokens` and conditionally omitting `logit_bias` when the selected model does not support it, while preserving the existing behavior for previous models.

  ### Related issue(s)

  Fixes #1989
  Fixes #1990

### DIFF
--- a/webapp/src/test/java/io/github/microcks/service/AICopilotRunnerServiceTest.java
+++ b/webapp/src/test/java/io/github/microcks/service/AICopilotRunnerServiceTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.service;
+
+import io.github.microcks.domain.Exchange;
+import io.github.microcks.domain.Operation;
+import io.github.microcks.domain.Request;
+import io.github.microcks.domain.RequestResponsePair;
+import io.github.microcks.domain.Resource;
+import io.github.microcks.domain.Response;
+import io.github.microcks.domain.Service;
+import io.github.microcks.domain.ServiceType;
+import io.github.microcks.util.ai.AICopilot;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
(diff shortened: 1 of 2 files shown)


paulux84:
wrong target branch

Issue #1901, opened 2026-01-08
Adapting our branching strategy


lbroudoux:
### Reason/Context

As the project welcomes new contributors, we realize that our branching strategy is not easily communicated, understood, and applied. Most of the time, new contributors push code to the `master` branch, where we expect PR to target the current development branch (aka `1.13.x` at the time of writing). This brings friction and frustration in the contribution process! Sometimes, contributors just let things down when it comes to rebasing their work on another branch... 

### Description

As a reminder, this organization was historically set up because we have different kinds of things to manage here:
* very volatile ones like Java code => we don't want a change here to impact non-advanced users,
* stable ones like Helm chart and Docker Compose files => we want people to be able to quickly retrieve stable charts and files without having to worry about the branch to checkout/clone

This may not be the best option of doing things and we'd like to open the discussion of what could be a better way of working with more contributions. See implementation ideas below for first ideas...

### Implementation ideas

Many options are coming to my mind: 

1. Keep things as they are today, but add an extra layer of communication and maybe guardrails (see #1900). Maybe with checks and early feedback, we can help people: switching to the correct branch, moving their local changes to this branch and targeting the correct one when doing the PR?

2. Make the approach more generic. Instead of having a dedicated branch per version (`1.12.x`, then `1.13.x`, then `1.14.x`...), switch to a model where we just have `development` and `maintenance` branches. Adding an extra layer of communication, documentation, and guardrails on top of this may help as well - while making things more stable and understandable for infrequent contributors.
 
3. Move to an unstable trunk/default branch development approach => accepting all the PR into `master` and maybe pulling out a new branch for consolidation before releases. This will probably have other impacts like:
* The versioning of components on `master` branch. We might need to introduce (as Quarkus does it) a `999-SNAPSHOT` permanent version that relates trunk/unstable stuff.
* The need to move stable stuffs (Helm chart and docker compose files) to a dedicated repo?
* A strong investment in understanding/automating the Git mechanisms that must be used to cherry-pick and push wanted changes on consolidation branches + update the CI/CD and release processes

> As a personal note, my fear is that option #3 requires a lot of workforce and bandwidth we don't have today! Maybe the project is not big enough to handle this...

I'd like to receive as much feedback and ideas as possible! 
@Harsh4902 @SebastienDegodez @Apoorva64 as regular contributors your ideas are more than welcome!

YASHMAHAKAL:
@lbroudoux, i will be helping on this, can you assign me please ?

[...]

PR #1992, opened 2026-03-13
OpenAI example generation and add gpt-5 family support [ #1989 #1990 ] 

### LINKED ISSUE #1989: [bug] AI Copilot bulk example generation

### LINKED ISSUE #1990: [bug] OpenAI gpt-5 model support

paulux84:
### Description

  - Refactor OpenAI sample generation to retry when the returned completion cannot be parsed, reusing the previous invalid output and parser error to ask for a corrected example instead of failing immediately.
  - Make AI sample generation more resilient at service level by skipping empty generations and logging per-operation failures without aborting the whole asynchronous run.
  - Add compatibility for GPT-5 family models by using `max_completion_tokens` and conditionally omitting `logit_bias` when the selected model does not support it, while preserving the existing behavior for previous models.

  ### Related issue(s)

  Fixes #1989
  Fixes #1990

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
+++ b/webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
@@ -43,6 +43,7 @@ import org.springframework.web.client.RestTemplate;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;


lbroudoux:
Woot! It looks great and I'll try to review this on tomorrow! 
In the meantime, it happens that you have the DCO check failing because your commits are not signed. Can you please check https://github.com/microcks/microcks/pull/1992/checks?check_run_id=67031293428 and fix it so that I'll be able to merge it?
Thanks,

cc @paulux84

paulux84:
I've fixed the problem.

cc @lbroudoux

lbroudoux:
> I've fixed the problem.
> 
> cc @lbroudoux

This is a false positive. I removed the warning. Thanks for pushing the fix, i'll pursue the review asap.

paulux84:
> It looks all good. Thanks for taking the time to put this together @paulux84. I have some work to finalize on `1.13.x` as I'm in the middle of #1901 but will merge it asap.

thanks for your work @lbroudoux ...really useful for dev community

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
...
+            return parseCompletionContent(service, operation, content);
+         } catch (Exception e) {
+            retryHistory.add(new RetryTurn(safeErrorMessage(e), content));
+            if (retryIndex == MAX_PARSE_RETRIES) {
+               log.error(
+                     "Could not generate valid AI samples for service {} operation {} after {} retries. Last error: {}",
+                     service.getName(), operation.getName(), MAX_PARSE_RETRIES, safeErrorMessage(e));
+               return new ArrayList<>();
+            }
+            log.warn("Parsing OpenAI output failed for service {} operation {}. Retrying attempt {}/{}. Error: {}",
```
SebastienDegodez, line 282: Suggestion: Wait a bit before retrying (100-500ms random)
This gives OpenAI time to "cool down", without this, we'd call OpenAI 5 times in a row = bad for the API

```
waitRandom(100, 500) // backoff
```

@lbroudoux ?

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
       headers.set("Authorization", "Bearer " + apiKey);
       return headers;
    }
+
+   private List<? extends Exchange> suggestSampleExchangesWithRetry(Service service, Operation operation,
+         String basePrompt) throws Exception {
+      List<RetryTurn> retryHistory = new ArrayList<>();
+
+      for (int retryIndex = 0; retryIndex <= MAX_PARSE_RETRIES; retryIndex++) {
```
SebastienDegodez, line 264: Suggestion: Replace the FOR loop with a WHILE loop for clarity.


A FOR loop is designed to iterate a KNOWN number of times.
 Here it's misleading because:
 - If the 1st attempt succeeds, we exit immediately via return
 - We don't know in advance how many iterations we'll do
 - The exit condition (if retryIndex == MAX_PARSE_RETRIES) is hidden in the catch block
 
A WHILE is more explicit
- The intent is clear: retry as long as we haven't exceeded the limit
- The counter `attempts++` is at the same place as the check `attempts > MAX`
- Lower bug risk: single source of retry logic (no hidden `retryIndex`)

```
int attempts = 0;
while (attempts <= MAX_PARSE_RETRIES) {
   try {
      return parseCompletionContent(...);
   } catch (Exception e) {
      retryHistory.add(...);
      attempts++;                           // <---- explicite
      
      if (attempts > MAX_PARSE_RETRIES) {   // <---- Check at same line
         return new ArrayList<>();
      }
      waitRandomMS(100, 500);           // <----  
   }
}
```

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
...
+            }
+            log.warn("Parsing OpenAI output failed for service {} operation {}. Retrying attempt {}/{}. Error: {}",
+                  service.getName(), operation.getName(), retryIndex + 1, MAX_PARSE_RETRIES, safeErrorMessage(e));
+         }
+      }
+      return new ArrayList<>();
+   }
+
+   private String requestCompletionContent(String prompt) {
+      log.debug("Asking OpenAI to suggest samples for this prompt: {}", prompt);
```
SebastienDegodez, line 290:
```suggestion
   private String requestCompletionContent(String prompt) {
      if (prompt == null || prompt.trim().isEmpty()) {
         log.warn("Prompt is null or empty");
         return null;
      }
      log.debug("Asking OpenAI to suggest samples for this prompt: {}", prompt);
```

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
...
+
+      final List<ChatMessage> messages = new ArrayList<>();
+      final ChatMessage assistantMessage = new ChatMessage(ChatMessageRole.ASSISTANT.value(), prompt);
+      messages.add(assistantMessage);
```
SebastienDegodez:
```suggestion
   final List<ChatMessage> messages = Collections.singletonList(
         new ChatMessage(ChatMessageRole.ASSISTANT.value(), prompt)
   );
```

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
...
+         return null;
+      }
+
+      ChatCompletionChoice choice = completionResult.getChoices().get(0);
+      if (choice == null || choice.getMessage() == null) {
+         return null;
+      }
+
+      log.debug("Got this raw output from OpenAI: {}", choice.getMessage().getContent());
+      return choice.getMessage().getContent();
```
SebastienDegodez, line 313: Extract another method

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
```diff
@@ -280,4 +255,118 @@ private HttpHeaders createAuthenticationHeaders() {
...
+
+      final List<ChatMessage> messages = new ArrayList<>();
+      final ChatMessage assistantMessage = new ChatMessage(ChatMessageRole.ASSISTANT.value(), prompt);
+      messages.add(assistantMessage);
+
+      HttpEntity<Map<String, Object>> request = new HttpEntity<>(buildChatCompletionPayload(messages),
+            createAuthenticationHeaders());
+      ResponseEntity<ChatCompletionResult> responseEntity = restTemplate.exchange(apiUrl + "/v1/chat/completions",
+            HttpMethod.POST, request, ChatCompletionResult.class);
+      ChatCompletionResult completionResult = responseEntity.getBody();
```
SebastienDegodez, line 300: WARNING: 3 possible exception types: HttpClientErrorException (4xx), HttpServerErrorException (5xx) and ResourceAccessException (Connection/Timeout).

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
paulux84: I don’t think a backoff retry is necessary in this case. The error we’re dealing with is not a connection or transient network issue, but rather an invalid or unsatisfactory result returned by the OpenAI model, so a standard backoff strategy doesn’t really apply here.

Note: this mechanism was introduced because, during example generation, the LLM occasionally produces incorrect outputs. Instead of failing immediately the example generation, i've implemented a retry strategy that includes the previous error in the prompt. With this approach, the generation typically succeeds on the second attempt, and at most on the third.

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
paulux84: the prompt can't be null here due to OpenAICopilot::preparePromptForOpenAPI calling

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
paulux84: maybe yes, but all type of exception here and in previous code is correctly catched inside @lbroudoux code

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
paulux84: what do you mean?

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
lbroudoux: Yeah, agree. There's no specific reason to have a backoff retry here, but some wait can be nice to avoid "429 too many requests" style of issue. What do think?

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
lbroudoux: Not sure if it actually makes sense to distinguish the different kinds of exceptions. I think we can handle them as they are now.

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
lbroudoux: `List.of()` is even more readable than `Collections.singleList()`. I would appreciate the update.

paulux84: I Will try to do that...I'm full of work this week

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
lbroudoux: Agree. `while` loop is more explicit in that case. However, I wouldn't put this on the critical path for the review. It's nice to have if @paulux84 finds some time but we can edit as an after-step otherwise.

SebastienDegodez: Of course :)

lbroudoux: Let's merge this that way for now so that we'll be able to include it in next 1.14.0-cr2

lbroudoux:
We have a formatting error, but I can fix it just after the merge. I think I'll do it this morning and integrate it into the `1.14.0-rc2` to come.

PR #2018, opened 2026-03-28
fix: support literal URL callback keys per OpenAPI 3.x spec (fixes #2…

### LINKED ISSUE #2: Remove duplicate Test Results
On some kind of tests, results are duplicated and displayed 2 times (no more) on results page

### LINKED ISSUE #2017: OpenAPI 3.x literal URL callback keys silently dropped (spec non-compliance)
## Summary
Microcks silently ignores OpenAPI callback definitions when the callback key is a literal URL 
(e.g. `http://host.docker.internal:6030/webhook`). Only `{$request.*}` runtime expressions 
are evaluated. This is non-compliant with the OpenAPI 3.x specification.

## OpenAPI Spec Reference
[OpenAPI 3.x §4.8.20](https://spec.openapis.org/oas/v3.1.0#callback-object) states:
> The key that identifies the Path Item Object is a runtime expression that can be evaluated 
> in the context of a runtime HTTP request/response **or a literal URL**.

Literal URL keys are valid per spec but do not fire in Microcks.

## Steps to Reproduce
1. Define an OpenAPI with a hardcoded callback key:
```yaml
callbacks:
  onCompleted:
    'http://host.docker.internal:6030/webhook':
      post:
        x-microcks-callback:
          order: 1
          delay: 5000
        ...
```
2. Import into Microcks
3. Trigger the operation

Expected: Microcks POSTs to http://host.docker.internal:6030/webhook after 5 seconds
Actual: Log shows No callback URL found for callbackInfo — callback never fires

## Root Cause
CallbackTrigger only resolves expressions matching {$request.*} patterns.
When the expression is a valid absolute URI it falls through to null → silent skip.

## Use Case / Impact
This blocks a key async mock testing pattern: environment-fixed callbacks
(e.g. test echo servers, internal services) where the callback target is infrastructure
config, not something the API client should provide per-request.
The consuming service (e.g. marketplace) should not need to know or send the mock's
callback URL.

## Proposed Fix
In CallbackTrigger, before the expression evaluator:

if (expression != null && expression.startsWith("http")) {
    URI uri = URI.create(expression);
    if (uri.isAbsolute()) return expression; // use literal URL directly
}
~10 lines. Fully backward compatible. No new YAML syntax required.

ramakrishna-89:
## What

`CallbackTrigger.extractCallbackUrlFromRequest` only handled OpenAPI runtime expressions
(`{$request.*}`). When the callback key was a literal URL (e.g.
`http://host.docker.internal:6030/webhook`), the expression was silently dropped with
`"No callback URL found"` — the callback was never fired.

Callback body templates also had no access to values generated by the mock response itself.
Only the original request context was available, so server-generated values (e.g. a
transaction ID returned by the mock) could not be echoed in the callback body.

Finally, `sendCallbackRequest` used a hard-coded 2 000 ms `HttpRequest.timeout`. When the
callback consumer performs non-trivial work before returning a response, Microcks closes the
TCP connection mid-flight — causing `context.canceled` on the consumer side.

### Commit 1 — literal URL callback keys

OpenAPI 3.x §4.8.20 explicitly allows callback keys to be **either** a runtime expression
**or a literal URL**. The missing literal-URL path is a spec compliance gap.

The most common use case is infrastructure-fixed callback targets: in integration testing
and local development, the callback destination is a known host/port determined by the
deployment environment, not something each API client supplies per request. Forcing the
callback URL into the request body adds noise to the API contract and requires the client
to know internal infrastructure addresses.

### Commit 2 — rendered response body in callback templates

Once the callback URL is hardcoded in the YAML, the client no longer needs to send it — but
it also no longer sends server-generated correlation values. In a typical async flow the mock
generates a value in its response (e.g. a transaction ID). The downstream callback must echo
that same value back so the consumer can correlate the two events. Without access to
`response.body`, the callback body can only reference fields from the original request —
making server-generated correlation IDs impossible to use.

### Commit 3 — configurable response timeout

**Root cause: `context.canceled` in the callback consumer.**

`sendCallbackRequest` builds an `HttpRequest` with `.timeout(Duration.ofMillis(2000))` and
dispatches it via `httpClient.sendAsync(...)`. `sendAsync` is non-blocking on the calling
thread, but the underlying `HttpClient` still holds an open TCP connection and waits up to
`timeout` milliseconds for the consumer to respond.

A realistic callback consumer performs work before it returns: database lookups, cache
reads, external API calls, signature verification, and so on. When this work takes longer
than 2 000 ms — which is common under a debugger, on a cold cache, or during any outbound
network call — Microcks fires a `TimeoutException` on the Java side, which closes the TCP
connection with a RST. The consumer's HTTP server detects the abrupt close and cancels the
request context. Every subsequent operation that propagates that context then returns
`context.canceled`, surfacing as a misleading internal error rather than a timeout.

The 2 000 ms value cannot be increased globally without affecting all callbacks in every
spec. The fix is to expose it as an optional per-callback field so each spec can declare the
timeout it actually needs.

### Commit 1 — literal URL callback keys

Added a short-circuit at the top of `extractCallbackUrlFromRequest`: if the expression is
already an absolute URL (`http://` or `https://` prefix), return it directly — no expression
evaluation needed. 7 lines in one method. Fully backward compatible — all existing
`{$request.*}` expression paths are unchanged.

### Commit 2 — rendered response body in callback templates

- **`RestInvocationProcessor`**: forwards the already-computed `responseContent` string into
  `handleCallbackTrigger` instead of discarding it after sending the HTTP response.
- **`CallbackTriggerEvent`**: adds a `renderedResponseBody` field; existing 2-arg constructors
  delegate with `null` — no breaking change for other event publishers.
- **`CallbackTrigger`**: passes `renderedResponseBody` to `renderContent` and propagates it
  when re-emitting chained callback events (multi-step `order:` sequences).
- **`ListenerCommons`**: adds a `renderContent` overload that registers the rendered response
  JSON as a `response` template variable, enabling `{{ response.body/<field> }}` in callback
  body templates. The existing 2-arg method delegates with `null` — no behaviour change for
  existing callers (SOAP, gRPC, GraphQL invocation processors are unaffected).

### Commit 3 — configurable response timeout

- **`CallbackInfo`**: adds a `responseTimeout` (`Long`) field with getter/setter.
- **`OpenAPIImporter`**: reads `timeout:` from the `x-microcks-callback` vendor extension
  block (alongside the existing `order:` and `delay:` fields) and stores it on
  `CallbackInfo.responseTimeout`.
- **`CallbackTrigger.sendCallbackRequest`**: accepts `responseTimeoutMs` as a parameter.
  Uses `(value != null && value > 0) ? value : 2000L` — absent or zero falls back to the
  existing default, making the change fully backward compatible.

In the YAML spec the field is declared as:

```yaml
x-microcks-callback:
  order: 1
  delay: 15000
  timeout: 30000   # ms — overrides the 2 000 ms default
```

The `x-` prefix means this is a vendor extension. It does not affect OpenAPI validation or
tooling that ignores unknown extensions.

## Testing

**`CallbackTriggerTest`** — 8 cases:

| Test | What it verifies |
|------|-----------------|
| `testLiteralHttpUrlReturnedDirectly` | `http://` key returned as-is (new) |
| `testLiteralHttpsUrlReturnedDirectly` | `https://` key returned as-is (new) |
| `testQueryParameterExpressionUnchanged` | `{$request.query.*}` backward compat |
| `testBodyExpressionUnchanged` | `{$request.body#/...}` backward compat |
| `testHeaderExpressionUnchanged` | `{$request.header.*}` backward compat |
| `testResponseTimeoutUsesConfiguredValueWhenSet` | configured value used (new) |
| `testResponseTimeoutFallsBackTo2000WhenNull` | null → 2 000 ms default (new) |
| `testResponseTimeoutFallsBackTo2000WhenZero` | zero → 2 000 ms default (new) |

**`OpenAPIImporterTest#testOpenAPIWithCallbacks`** — asserts `responseTimeout == 5000L`
from the `timeout: 5000` in the updated `callback-example-openapi.json` fixture.

**`ListenerCommonsTest`** — 8 existing cases, all passing.

## Use Case

Enables environment-fixed callback targets to be hardcoded in the OpenAPI YAML — useful for
integration testing where the callback destination is infrastructure config, not something
each API client should supply per request.

With all three changes together, a full async callback flow works with no client-side
changes:

1. The YAML hardcodes where the callback fires (e.g. `http://host.docker.internal:8080/webhook`)
2. `{{ response.body/<field> }}` lets the callback body echo server-generated values for
   correct correlation at the receiving end
3. `timeout: <ms>` gives the consumer enough time to complete its work before Microcks
   closes the connection — eliminating spurious `context.canceled` errors

Fixes #2017

### DIFF
--- a/webapp/src/main/java/io/github/microcks/listener/ListenerCommons.java
+++ b/webapp/src/main/java/io/github/microcks/listener/ListenerCommons.java
@@ -47,6 +47,15 @@ public class ListenerCommons {
     * @return The rendered content or the original template if no evaluation was possible.
     */
    public static String renderContent(HttpServletRequestSnapshot requestSnapshot, String contentTemplate) {
+      return renderContent(requestSnapshot, null, contentTemplate);
+   }
+
+   /**
+    * Render callback (or async) content using the incoming request and optionally the rendered mock HTTP response body.
+    * When {@code renderedResponseBody} is JSON, templates may use e.g. {@code response.body/transId}.
+    */
+   public static String renderContent(HttpServletRequestSnapshot requestSnapshot, String renderedResponseBody,
+         String contentTemplate) {
       if (contentTemplate != null && contentTemplate.contains(TemplateEngine.DEFAULT_EXPRESSION_PREFIX)) {
 
          TemplateEngine engine = TemplateEngineFactory.getTemplateEngine();
@@ -61,8 +70,11 @@ public class ListenerCommons {
                .filter(e -> e.getValue() != null && !e.getValue().isEmpty())
                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getFirst())));
 
-         // Register the request variable and evaluate the response.
          engine.getContext().setVariable("request", evaluableRequest);
+         if (renderedResponseBody != null && !renderedResponseBody.isBlank()) {
+            EvaluableRequest evaluableResponse = new EvaluableRequest(renderedResponseBody, new String[] {});
+            engine.getContext().setVariable("response", evaluableResponse);
+         }
          try {
             return engine.getValue(contentTemplate);
          } catch (Exception e) {
--- a/commons/model/src/main/java/io/github/microcks/domain/CallbackInfo.java
+++ b/commons/model/src/main/java/io/github/microcks/domain/CallbackInfo.java
@@ -32,6 +32,7 @@ public class CallbackInfo implements Serializable {
    private String method;
    private Integer order;
    private Long defaultDelay;
+   private Long responseTimeout;
 
    public CallbackInfo() {
    }
@@ -71,5 +72,13 @@ public class CallbackInfo implements Serializable {
 
    public void setDefaultDelay(Long defaultDelay) {
       this.defaultDelay = defaultDelay;
+   }
+
+   public Long getResponseTimeout() {
+      return responseTimeout;
+   }
+
+   public void setResponseTimeout(Long responseTimeout) {
+      this.responseTimeout = responseTimeout;
    }
 }
--- a/webapp/src/main/java/io/github/microcks/event/CallbackTriggerEvent.java
+++ b/webapp/src/main/java/io/github/microcks/event/CallbackTriggerEvent.java
@@ -28,6 +28,8 @@ public class CallbackTriggerEvent extends ApplicationEvent {
    private final Operation operation;
    private final String responseName;
    private final HttpServletRequestSnapshot request;
+   /** JSON body actually returned to the client (rendered mock response); used for OpenAPI callback templates. */
+   private final String renderedResponseBody;
 
    private int step = 0;
 
@@ -41,11 +43,7 @@ public class CallbackTriggerEvent extends ApplicationEvent {
     */
    public CallbackTriggerEvent(Object source, String serviceId, Operation operation, String responseName,
          HttpServletRequestSnapshot request) {
-      super(source);
-      this.serviceId = serviceId;
-      this.operation = operation;
-      this.responseName = responseName;
-      this.request = request;
+      this(source, serviceId, operation, responseName, request, null, 0);
    }
 
    /**
@@ -59,11 +57,20 @@ public class CallbackTriggerEvent extends ApplicationEvent {
     */
    public CallbackTriggerEvent(Object source, String serviceId, Operation operation, String responseName,
          HttpServletRequestSnapshot request, int order) {
+      this(source, serviceId, operation, responseName, request, null, order);
+   }
+
+   /**
+    * Full constructor including rendered HTTP response body (for callback templates: {@code response.body/...}).
+    */
+   public CallbackTriggerEvent(Object source, String serviceId, Operation operation, String responseName,
+         HttpServletRequestSnapshot request, String renderedResponseBody, int order) {
       super(source);
       this.serviceId = serviceId;
       this.operation = operation;
       this.responseName = responseName;
       this.request = request;
+      this.renderedResponseBody = renderedResponseBody;
       this.step = order;
    }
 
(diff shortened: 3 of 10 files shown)


webapp/src/main/java/io/github/microcks/listener/ListenerCommons.java
```diff
@@ -61,8 +70,11 @@ public static String renderContent(HttpServletRequestSnapshot requestSnapshot, S
                .filter(e -> e.getValue() != null && !e.getValue().isEmpty())
                .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getFirst())));
 
-         // Register the request variable and evaluate the response.
          engine.getContext().setVariable("request", evaluableRequest);
+         if (renderedResponseBody != null && !renderedResponseBody.isBlank()) {
+            EvaluableRequest evaluableResponse = new EvaluableRequest(renderedResponseBody, new String[] {});
+            engine.getContext().setVariable("response", evaluableResponse);
```
dipak0000812, line 76: This is a nice addition for enabling `response.body` access in callback templates.

One question regarding thread safety: do we know whether `TemplateEngineFactory.getTemplateEngine()` returns a new instance per invocation? Since we mutate the engine context (`setVariable`), a shared engine could lead to cross-request leakage in concurrent callback executions.

If the engine is singleton-scoped, we may need to reset or clone the context before evaluation.

lbroudoux: > do we know whether TemplateEngineFactory.getTemplateEngine() returns a new instance per invocation? 

Yes, it's the case. We got a new instance each and every time.

lbroudoux:
Thanks for this PR but there's a major point to cover first: https://github.com/microcks/microcks/issues/2017#issuecomment-4160664282

Issue #1916, opened 2026-01-15
AsyncAPI Request/Reply


adamhicks-dojo:
### Reason/Context

At the moment Microcks only supports sending messages on a scheduled interval. Our applications are using a request/response pattern (that relies on a specific field "Correlation-ID" being copied from request to the response) which is not reproduce-able in Microcks.
We're planning to create a minimal (internal) tool to bridge this gap for us, but it would benefit microcks if we could add this. It would also make our specification files more consistent.

### Description

In this [comment](https://github.com/microcks/microcks/issues/402#issuecomment-1015174902) @lbroudoux mentioned:

> Regarding Request-Reply pattern: we don't have support for this for the moment as there's no standard way of specifying this yet. There's some ongoing discussions on how to add that to AsyncAPI and be sure we'll support it as soon as there is consensus on how to represent that.

I'm new to AsyncAPI so I don't know when [this](https://www.asyncapi.com/docs/concepts/asyncapi-document/reply-info) spec was added, I presume it was after the previous comment (i.e. in the last 3 years). If this spec is sufficient enough, can we discuss adding this feature(s) to Microcks?

lbroudoux:
Hello @adamhicks-dojo ,

Yes, AsyncAPI v3 has added the request/reply support. 

>  It would benefit microcks if we could add this. It would also make our specification files more consistent.
Totally agree, it would be awesome if we could co-develop this feature and have it supported in Microcks! 

This is the right timing as I would also like to revive a topic on [OpenAPI callbacks & webhooks + Sync-to-Async triggers features and design](https://github.com/orgs/microcks/discussions/1039) and the question or request/reply on async is correlated. If you haven't been through discussion #1039, I propose you have a read so you can see the kind of design brainstorming and decision-making we may have on this topic.

I'll also start to dedicate a some brain-time to this 😉

adamhicks:
That's great, I can see that there's some crossover in these features dependent on the implementation, I'll give that thread a read and give it some thought. 

Would you like me to create a design Idea similar in structure to #1039 but for request/reply?

(also, i have merged my work account -dojo into my personal account, hence the change of account and "ghost")

aviralgarg05:
I would like to solve the issue, please assign me

lbroudoux:
> I would like to solve the issue, please assign me

[...]

PR #2031, opened 2026-04-08
feat(async): Implement request-reply for kafka

adamhicks:
### Description

- Implement a request reply handler that can monitor (kafka) topics for messages to reply to
- Implement the request reply manager that can create monitors for operations (from the definitions streamed from the microcks API)
- Implement testcontainers test for a request-reply interaction
- Implement a message matcher based on body content, wasn't able to use Dispatch Criteria (open to discussion)

### Related issue(s)

See https://github.com/microcks/microcks/issues/1916

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
@@ -1,0 +1,80 @@
+package io.github.microcks.minion.async.handler;
+
+import com.fasterxml.jackson.core.exc.StreamReadException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.microcks.domain.EventMessage;
+import org.jboss.logging.Logger;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+/**
+ * Compute dispatch criteria for async request-reply events.
+ */
+public class EventData {
+
+   private static final Logger LOGGER = Logger.getLogger(EventData.class);
+   private static final ObjectMapper MAPPER = new ObjectMapper();
+
+   private EventData() {
+   }
+
+   public static boolean matchesExpectedMessage(EventMessage expMessage, String payload) {
+      if (expMessage == null) {
+         return false;
+      }
+      String expectedCriteria = expMessage.getContent();
+      if (expectedCriteria == null || expectedCriteria.isBlank()) {
+         return true;
+      }
+      try {
+         JsonNode expectedNode = MAPPER.readTree(expectedCriteria);
+         JsonNode incomingNode = MAPPER.readTree(payload);
+         return isJsonSubset(expectedNode, incomingNode);
+      } catch (StreamReadException e) {
+         LOGGER.errorf("message data is not valid JSON: %s", e.getMessage());
+         return false;
+
+      } catch (IOException e) {
+         LOGGER.errorf("Error reading expected message data: %s", e.getMessage());
+         return false;
+      }
+   }
+
+   private static boolean isJsonSubset(JsonNode expectedNode, JsonNode incomingNode) {
+      if (expectedNode == null || incomingNode == null) {
+         return expectedNode == incomingNode;
+      }
+
+      if (expectedNode.isObject()) {
+         if (!incomingNode.isObject()) {
+            return false;
+         }
+         Iterator<String> fieldNames = expectedNode.fieldNames();
+         while (fieldNames.hasNext()) {
+            String fieldName = fieldNames.next();
+            JsonNode expectedFieldValue = expectedNode.get(fieldName);
+            JsonNode incomingFieldValue = incomingNode.get(fieldName);
+            if (incomingFieldValue == null || !isJsonSubset(expectedFieldValue, incomingFieldValue)) {
+               return false;
+            }
+         }
+         return true;
+      }
+
+      if (expectedNode.isArray()) {
+         if (!incomingNode.isArray() || expectedNode.size() != incomingNode.size()) {
+            return false;
+         }
+         for (int i = 0; i < expectedNode.size(); i++) {
+            if (!isJsonSubset(expectedNode.get(i), incomingNode.get(i))) {
+               return false;
+            }
+         }
+         return true;
+      }
+
+      return expectedNode.equals(incomingNode);
+   }
+}
--- a/minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.github.microcks.minion.async.handler;
 
 import com.fasterxml.jackson.core.exc.StreamReadException;
@@ -11,11 +26,13 @@ import java.util.Iterator;
 
 /**
  * Compute dispatch criteria for async request-reply events.
+ *
+ * @author adamhicks
  */
 public class EventData {
 
-   private static final Logger LOGGER = Logger.getLogger(EventData.class);
-   private static final ObjectMapper MAPPER = new ObjectMapper();
+   private static final Logger logger = Logger.getLogger(EventData.class);
+   private static final ObjectMapper mapper = new ObjectMapper();
 
    private EventData() {
    }
@@ -29,15 +46,15 @@ public class EventData {
          return true;
       }
       try {
-         JsonNode expectedNode = MAPPER.readTree(expectedCriteria);
-         JsonNode incomingNode = MAPPER.readTree(payload);
+         JsonNode expectedNode = mapper.readTree(expectedCriteria);
+         JsonNode incomingNode = mapper.readTree(payload);
          return isJsonSubset(expectedNode, incomingNode);
       } catch (StreamReadException e) {
-         LOGGER.errorf("message data is not valid JSON: %s", e.getMessage());
+         logger.errorf("message data is not valid JSON: %s", e.getMessage());
          return false;
 
       } catch (IOException e) {
-         LOGGER.errorf("Error reading expected message data: %s", e.getMessage());
+         logger.errorf("Error reading expected message data: %s", e.getMessage());
          return false;
       }
    }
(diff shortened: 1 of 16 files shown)


adamhicks:
```
$ mvn test
...
[INFO] Reactor Summary for Microcks 1.14.0-SNAPSHOT:
[INFO] 
[INFO] Microcks ........................................... SUCCESS [  0.420 s]
[INFO] Microcks Model ..................................... SUCCESS [  0.801 s]
[INFO] Microcks Util ...................................... SUCCESS [  2.517 s]
[INFO] Microcks EL ........................................ SUCCESS [  0.520 s]
[INFO] Microcks App ....................................... SUCCESS [ 54.300 s]
[INFO] Microcks Async Minion .............................. SUCCESS [  1.537 s]
[INFO] Microcks Distros ................................... SUCCESS [  0.006 s]
[INFO] Microcks Uber App .................................. SUCCESS [  0.075 s]
[INFO] Microcks Uber Async Minion ......................... SUCCESS [  0.094 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:00 min
[INFO] Finished at: 2026-04-08T09:56:02+01:00
```

lbroudoux:
This one could be merged "just-in-time" for 1.14.0. Cool 👍

minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
```diff
@@ -0,0 +1,80 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.github.microcks.domain.EventMessage;
+import org.jboss.logging.Logger;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+/**
+ * Compute dispatch criteria for async request-reply events.
```
adamhicks, line 13: I tried to define a JSON_BODY dispatch criteria when parsing the example but I think I struggled to get AND behaviour. I was trying to create "equals" criterion for each leaf of the example json. e.g. if the example was:
```
{
    "payment_request_id": 123,
    "user": {
          "id": 456
    }
}
```
I wanted the criteria to behave like (pseudo code):
```
if body.payment_request_id == 123 && body.user.id == 456 then
   return "example-1"
else
   return "unmatched"
end if
```

minions/async/src/test/java/io/github/microcks/minion/async/producer/KafkaProducerManagerIT.java
```diff
@@ -126,6 +127,7 @@ void testProduceAvroMockMessages() throws Exception {
 
       Operation signedupOperation = new Operation();
       signedupOperation.setName("SUBSCRIBE user/signedup");
+      signedupOperation.addResourcePath(TOPIC_NAME);
```
adamhicks, line 130: this test actually broke in 9f78c5890d06412d4010e7feb3ffea627c7b9b80 where I changed `getTopicName` to use the channel name from the API doc (rather than a generated one)

adamhicks:
Tried to fix the sonar action as it was failing on getting coverage for an adjacent test to this PR.

Got a different [error](https://github.com/microcks/microcks/actions/runs/24144569910/job/70454913742?pr=2031) now that indicates auth with Sonar isn't set up or has expired
```
Error:  Error retrieving feature flags, using default analyzer behavior
Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to
```

adamhicks:
> This one could be merged "just-in-time" for 1.14.0. Cool 👍

I think this PR is ready for review, just making sure you're not waiting for me to do something more?

lbroudoux:
> I think this PR is ready for review, just making sure you're not waiting for me to do something more?

No that's cool. It's just me: I'm super busy and slow at the moment. Apologies for that 🙏

minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
```diff
@@ -234,29 +241,29 @@ public Set<Header> renderEventMessageHeaders(TemplateEngine engine, Set<io.githu
...
-      operationName = operationName.replace('/', '-');
+      if (resourcePaths == null || resourcePaths.size() != 1) {
+         throw new IllegalArgumentException(
+               String.format("Operation '%s' must have exactly one resource path to determine topic name, but has: %s",
+                     definition.getOperation().getName(), resourcePaths));
+      }
 
-      // Aggregate the 3 parts using '_' as delimiter.
-      return serviceName + "-" + versionName + "-" + operationName;
+      return resourcePaths.iterator().next();
```
lbroudoux, line 261: I don't understand why you actually needed to change that part... Existing behaviour allows us to generate different topic names when one resource path contains `{}` parts and uses an `URI_PARTS` dispatcher.  Here, you are no longer using the `eventMessage` to define the topic name. So dynamic topics seem no longer implemented. Am I guessing wrong?

minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class EventData {
+
+   private static final Logger LOGGER = Logger.getLogger(EventData.class);
```
lbroudoux, line 17: Could you use `logger` instead for better consistency with existing code?

minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
```diff
@@ -0,0 +1,80 @@
...
+   private static final ObjectMapper MAPPER = new ObjectMapper();
```
lbroudoux, line 18: Could you use `mapper` instead for better consistency with existing code?

minions/async/src/main/java/io/github/microcks/minion/async/handler/EventData.java
```diff
@@ -0,0 +1,80 @@
+package io.github.microcks.minion.async.handler;
```
lbroudoux, line 1: File is missing the copyright header. Can you please add it?

minions/async/src/test/java/io/github/microcks/minion/async/producer/KafkaProducerManagerTest.java
```diff
@@ -54,11 +56,11 @@ void testGetTopicName() {
       AsyncMockDefinition definition = new AsyncMockDefinition(service, operation, eventsMessages);
 
       String topicName = producerManager.getTopicName(definition, eventMessage);
-      assertEquals("StreetlightsAPI-0.1.0-receiveLightMeasurement", topicName);
+      assertEquals("smartylighting.streetlights.1.0.event.lighting.measured", topicName);
```
lbroudoux, line 59: This is not correct and breaks the existing conventions => we must ensure uniqueness of topic to avoid collisions between different versions or even different AsyncAPIs.

minions/async/src/test/java/io/github/microcks/minion/async/producer/KafkaProducerManagerTest.java
```diff
@@ -54,11 +56,11 @@ void testGetTopicName() {
...
    }
 
    @Test
-   void testGetDynamicTopicName() {
+   void testGetTopicNameThrowsWhenMultipleResourcePaths() {
```
lbroudoux, line 63: I think we have to review this part together. It looks like you removed support for dynamic topic name generation.

lbroudoux (changes requested):
Thanks @adamhicks for pushing and thanks for your patience! There are some minor change requests but a major concern here is related to the dynamic topic resolution - it seems to be broken/removed now and I don't understand why you have to do so.

minions/async/src/test/java/io/github/microcks/minion/async/producer/KafkaProducerManagerIT.java
lbroudoux: Yes, but channels/topic name override is super important to guarantee the isolation of mocks from different versions or even different APIs. We need to keep generated topic names with support of dynamic parts when using `{}` in the name.

minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
adamhicks: No you're right, this is no longer implemented here and I need to fix the dynamic topic naming using the message parts. I was mistaken in thinking that this is incompatible with request reply, because I assumed we'd have to listen on an arbitrary number of topics, but actually this is bounded by the examples given, so I think we can use the examples to fix the number of topics that we need to listen on. I can work on that.

adamhicks:
> Thanks @adamhicks for pushing and thanks for your patience! There are some minor change requests but a major concern here is related to the dynamic topic resolution - it seems to be broken/removed now and I don't understand why you have to do so.

The dynamic resolution (using parts of the example message) does need to come back.

The dynamic naming using service/operation/version info, I don't know how to make this compatible. How do clients currently find these topic names? E.g. if my application is expecting mocked events, does it need logic to be able to find microcks defined topic names? I'd expect the topic to be a fixed name defined in the spec, but I'm aware that I'm only really looking at my experience/use case here, so I'm keen to know how this is supposed to work 🙏

lbroudoux:
> The dynamic resolution (using parts of the example message) does need to come back.
> 
> The dynamic naming using service/operation/version info, I don't know how to make this compatible. How do clients currently find these topic names? E.g. if my application is expecting mocked events, does it need logic to be able to find microcks defined topic names? I'd expect the topic to be a fixed name defined in the spec, but I'm aware that I'm only really looking at my experience/use case here, so I'm keen to know how this is supposed to work 🙏

Ok, let me try to explain.

Let's reason from an example: imagine the channel address in `MyAsyncAPI`, version `1.0` is defined as `topic-{lightId}` for operation `SEND events`.

To allow isolation of APIs and versions, Microcks will ignore the address from the spec and define the mock topic as being `<service_name>-<service_version>-<operation_name>`, so it should be `MyAsyncAPI-1.0-events` (as we remove the operation verb - here `SEND`). This is the generic case when using static destinations. 

But here, the channel address is defined using a `{}` expression - which means it is dynamic and may change depending on event messages.  In that case, Microcks detects this during the import and sets the dispatcher as `URI_PARTS` - then we use a slightly different naming convention, being now `<service_name>-<service_version>-<dynamic_channel_name>`. The `lightId` value will actually be extracted from mocked events and used to form the topic name.

So imagine, your event sample has a `lightId` being `xyz`, then the topic name computed by Microcks will now be `MyAsyncAPI-1.0-topic-xyz`. 

> How do clients currently find these topic names? E.g. if my application is expecting mocked events, does it need logic to be able to find microcks defined topic names? 

The topic name (either static or dynamic) is rendered on the Microcks UI. So if you define multiple message samples in your AsyncAPI, you will have multiple thumbs in the Microcks UI with different topic names on each thumb.

Hope it makes it clearer. Let me know if it needs any further clarification.

adamhicks:
> The topic name (either static or dynamic) is rendered on the Microcks UI. So if you define multiple message samples in your AsyncAPI, you will have multiple thumbs in the Microcks UI with different topic names on each thumb.

I'm coming at this from the perspective of an application that is expecting to receive events and do something with them. So it needs to know where to listen for these events. Would the expected workflow be for the user to use the Microcks UI to get the topic name and then configure there application with something like:
```
HELLO_BLOG_EVENTS_TOPIC_NAME= MyAsyncAPI-1.0-topic-xyz
```
?

If so, I think I get it :)

lbroudoux:
Yes, that's it. Maybe my example with `{lightId}` is not that great because it would lead to a lot of topics... Common use cases are topics per type/category or per grouping criteria. But you get the point! Thanks for the new commit, I'll review on tomorrow morning.

lbroudoux (approved):
Looks all good to me!

Issue #1784, opened 2025-10-15
Helm chart: support TLS on microcks HTTP endpoint


rquinio1A:
### Reason/Context

In some cases we may need the mock endpoint to support HTTPS in addition to HTTP:

- .NET OIDC client enforces HTTPS is used by default during the download of public keys of identity provider.
- Kubernetes network policies may disallow non-TLS connections
- etc.

### Description

The microcks Helm chart support TLS certificate for the gRPC endpoint, but I didn't see the logic to mount a certificate on the spring-boot HTTP endpoint.

### Implementation ideas

In my case, the Certificate will be generated by cert-manager operator, as a k8s secret of type kubernetes.io/tls https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets
tls.crt and tls.key must be mounted into spring-boot web layer.

lbroudoux:
Hi @rquinio1A,
Thinking about that: TLS certs are put at the Ingress level with the current Helm chart - both Keycloak and Microcks HTTP endpoints can thus benefit from certs generated by cert-manager and placed in secrets to be used by their respective ingresses. Wouldn't it work if you make your .NET client point on these ingress points?

rquinio1A:
In our case Microcks and the microservices are in the same Kubernetes cluster, and I'd like to use directly the .svc.cluster.local K8s service DNS name rather than the ingress, as it's typically blocked by our network policies.

lbroudoux:
> In our case Microcks and the microservices are in the same Kubernetes cluster, and I'd like to use directly the .svc.cluster.local K8s service DNS name rather than the ingress, as it's typically blocked by our network policies.

OK got it!

sudheerr937-ai:
Hi @lbroudoux , @rquinio1A 
I'd like to work on this issue

My proposed approach:
Add an optional microcks.http.tlsSecretRef field in values.yaml to reference an existing kubernetes.io/tls secret
In the Deployment template, conditionally mount tls.crt and tls.key from that secret as a volume when the field is set
Configure Spring Boot server.ssl.certificate and server.ssl.certificate-private-key properties pointing to the mounted certificate path

sudheerr937-ai:
Hi @rquinio1A , @lbroudoux 
I have submitted a PR for this issue:  https://github.com/microcks/microcks/pull/2038
Please review and let me know if any changes are needed. Happy to update.

lbroudoux:
Hey @sudheerr937-ai and @rquinio1A 
[...]

PR #2038, opened 2026-04-15
feat: add TLS support for Microcks HTTP endpoint #1784

sudheerr937-ai:
Summary:-
Fixes -1784

 Changes:-
- Added httpEnableTLS and httpSecretRef options in values.yaml
- Added HTTPS port 8443 in deployment.yaml
- Added conditional volumeMount and volume for HTTP TLS certificate
- Updated liveness, readiness and startup probes to support HTTPS
- Added Spring Boot SSL configuration in `configmap.yaml`

 How to use:-
To enable TLS on the HTTP endpoint, set the following values:
```yaml
microcks:
  httpEnableTLS: true
  httpSecretRef: my-tls-secret
```
The secret must be of type 'kubernetes.io/tls' containing 'tls.crt' and 'tls.key'.

testing
- helm lint passed with 0 failures
- helm template verified all TLS properties are generated correctly

### DIFF
--- a/install/kubernetes/microcks/templates/configmap.yaml
+++ b/install/kubernetes/microcks/templates/configmap.yaml
@@ -147,6 +147,15 @@ data:
     grpc.server.privateKeyFilePath=/deployments/config/grpc/tls.key
     {{- end }}
 
+    {{- if .Values.microcks.httpEnableTLS }}
+
+    # HTTP server TLS properties
+    server.port=8443
+    server.ssl.enabled=true
+    server.ssl.certificate=/deployments/config/https/tls.crt
+    server.ssl.certificate-private-key=/deployments/config/https/tls.key
+    {{- end }}
+
     # AI Copilot configuration properties
     ai-copilot.enabled={{ .Values.features.aiCopilot.enabled }}
     ai-copilot.implementation={{ .Values.features.aiCopilot.implementation }}
--- a/install/kubernetes/microcks/templates/deployment.yaml
+++ b/install/kubernetes/microcks/templates/deployment.yaml
@@ -41,6 +41,11 @@ spec:
         - name: grpc
           containerPort: 9090
           protocol: TCP
+        {{- if .Values.microcks.httpEnableTLS }}
+        - name: https
+          containerPort: 8443
+          protocol: TCP
+        {{- end }}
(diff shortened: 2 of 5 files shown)


dipak0000812 (commented):
Nice addition, useful feature.

I had one question while checking the templates: this PR updates the app/probes to use HTTPS on `8443`, but the main Service still points to `8080`, and the Ingress also seems to use service port `8080`.

If enabling `httpEnableTLS=true` moves Microcks to `8443`, should Service/Ingress be updated too? Otherwise traffic may still go to the old port.

Maybe I’m missing another template that already handles this.

sudheerr937-ai:
Hi @dipak0000812, good catch!

You're right, the Service and Ingress were still using port 8080 even when 'httpEnableTLS=true'. I've updated both:

service.yaml - port and targetPort now use 8443 when TLS is enabled
ingress.yaml - service port now uses 8443 when TLS is enabled

Tested locally with helm template for both cases and it looks good. Let me know if you spot anything else

dipak0000812 (commented):
Thanks for the update — just checked the latest changes and the Service/Ingress alignment with `httpEnableTLS` looks good now 👍

One small note: DCO check is currently failing — I think commits might need a sign-off.

Nice improvement overall.

Vaishnav88sk:
Fixes #1784

install/kubernetes/microcks/templates/configmap.yaml
```diff
@@ -147,6 +147,15 @@ data:
     grpc.server.privateKeyFilePath=/deployments/config/grpc/tls.key
     {{- end }}
 
+    {{- if .Values.microcks.httpEnableTLS }}
+
+    # HTTP server TLS properties
+    server.port=8443
+    server.ssl.enabled=true
+    server.ssl.certificate=/deployments/config/https/tls.crt
+    server.ssl.certificate-private-key=/deployments/config/https/tls.key
```
lbroudoux, line 156: I think those properties are still available for backward compatibility, but the now recommended wya of doing things is to use `server.ssl.bundle` properties.

lbroudoux (approved):
Thanks a lot for pushing this. I added one comment but it's more for future enhancement.

PR #2046, opened 2026-04-20
feat(mcp): make SSE emitter timeout configurable

### LINKED ISSUE #2033: Make SSE emitter timeout in McpController configurable
### Reason/Context

The SSE emitter timeout in McpController is currently hardcoded to 600 seconds. This makes it difficult to adjust connection lifetime based on different deployment environments (e.g., reverse proxies, ingress controllers, or load balancers) that may require shorter or longer timeouts.

Allowing this value to be configurable would improve operational flexibility while preserving the existing default behavior.

### Description

In `McpController`, the SSE emitter is initialized with a fixed timeout:

SseEmitter emitter = new SseEmitter(Duration.ofSeconds(600).toMillis());

Externalizing this timeout into a configuration property would allow deployments to tune the value without modifying code. The default value can remain unchanged to ensure backward compatibility.

This would not introduce a breaking change.

### Implementation ideas

_No response_

dipak0000812:
Fixes #2033

Make SSE emitter timeout in McpController configurable.

- Add property `mcp.sse.timeout` (default: 600 seconds)
- Replace hardcoded timeout with configurable value
- Preserve existing behavior

No breaking changes.

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/McpController.java
+++ b/webapp/src/main/java/io/github/microcks/web/McpController.java
@@ -34,6 +34,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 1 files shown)


dipak0000812:
Hi @lbroudoux,

Quick follow-up on this PR.

I noticed the failing checks:
- SonarCloud
- License compliance (1 issue)

The change itself is minimal and limited to McpController, so these seem unrelated to the diff.

Could you confirm if these are baseline issues on master or if you’d like me to investigate further?

Happy to adjust if needed.

lbroudoux:
> Could you confirm if these are baseline issues on master or if you’d like me to investigate further?

Yes, I think it's unrelated. I currently have stability issues on the build due to migration to Java 25, but I hope I'll be able to merge soon.

lbroudoux:
Thanks for the contribution! This is great but is actually only part of the solution. How to make this configurable when installing Microcks? Can you set an environment variable to override this? Do you have to specify this in the application.properties file .... We also have to make this configurable when installing on Kubernetes. DO you want to follow up with these or may I just merge at this point?

dipak0000812:
Hi @lbroudoux, 
Makes sense, I was mainly focused on removing the hardcoded value first. I’ll check how other configurable properties are exposed for Kubernetes/container installs and follow up on this PR with the env/Helm side as well.

PR #2047, opened 2026-04-20
feat(mocks): make ServiceStateStore default TTL configurable

### LINKED ISSUE #1964: Make ServiceStateStore default ttl configurable
### Reason/Context

While the [docs](https://microcks.io/documentation/explanations/dispatching/#script-dispatcher) explain that one can be override this using the ttlInSeconds parameter, having to do this in a lot of different API specs SCRIPT dispatcher snippet is not easy to update or maintain in the long term.

> Allows you to access a service-scoped persistent store for string values. Such store elements can be later reused in other operations’ scripts to keep track of state or feed the requestContext. Store provides helpful methods like put(key, value), get(key) or delete(key). Store elements are subject to a Time-To-Live that is 10 seconds by default. This TTL can be overridden using the put(key, value, ttlInSeconds) method.

### Description

Make ServiceStateStore default ttl configurable keep the default to 10 seconds so it's not breaking change.

### Implementation ideas

Configure it either via a property or environment variable, like `keycloak.enabled`

dipak0000812:
Fixes #1964

Make ServiceStateStore default TTL configurable.

- Add property `mocks.service-state-store.default-ttl`
- Default remains 10 seconds
- Inject configurable TTL into invocation processors
- Add unit test validating custom TTL

No breaking changes.

### DIFF
--- a/install/kubernetes/microcks/values.yaml
+++ b/install/kubernetes/microcks/values.yaml
@@ -87,6 +87,10 @@ microcks:
     #- name: CORS_REST_ALLOW_CREDENTIALS
     #  value: 'true'
 
+    #- name: MCP_SSE_TIMEOUT
+    #  value: "900000"
+    #- name: MOCKS_SERVICE_STATE_STORE_DEFAULT_TTL
+    #  value: "30"
   #logLevel: TRACE | DEBUG | INFO | WARN
   logLevel: INFO
 
--- a/webapp/src/main/java/io/github/microcks/service/ServiceStateStore.java
+++ b/webapp/src/main/java/io/github/microcks/service/ServiceStateStore.java
@@ -33,6 +33,7 @@ public class ServiceStateStore implements StateStore {
 
    private final ServiceStateRepository repository;
    private final String serviceId;
+   private final int defaultSecondsTtl;
 
    /**
     * Build a ServiceStateStore for required elements.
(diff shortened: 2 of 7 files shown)


dipak0000812:
Hi @lbroudoux,

Quick follow-up on this PR.

This change introduces a configurable default TTL for ServiceStateStore while keeping the existing 10s default. The scope is limited to wiring the configuration through invocation processors and adding a unit test for validation.

I noticed the failing checks (SonarCloud and License Compliance). From what I can see, there are no new issues in the modified files, so this may be unrelated to the change — but I’m happy to investigate further if needed.

Also, since this touches configuration, please let me know if you'd prefer aligning it with Helm/operator values as part of this PR or in a follow-up.

Thanks!

lbroudoux (commented):
LGTM

lbroudoux:
Thanks a lot for this one but I have the same comment that in #2046 

> This is great but is actually only part of the solution. How to make this configurable when installing Microcks? Can you set an environment variable to override this? Do you have to specify this in the application.properties file .... We also have to make this configurable when installing on Kubernetes.

Do you want to follow up with these or may I just merge at this point? Let us know! We'd be happy to merge the complete resolution 😉

dipak0000812:
Hi @lbroudoux, thanks for the review and feedback,
You're right — the current change only covers the application-level configurability. I’d like to follow up by wiring this through deployment/configuration layers as well so the feature is fully usable in Kubernetes and container-based installations.
I’ll work on adding support through environment variables and Helm/operator configuration where appropriate, then update this PR with the complete solution.

dipak0000812:
Hi @lbroudoux ,
Followed up on the Kubernetes/runtime configuration side as well.
The chart already supports custom env injection through `microcks.env`, so I added examples for configuring:
* `MOCKS_SERVICE_STATE_STORE_DEFAULT_TTL`
* `MCP_SSE_TIMEOUT`
This keeps the existing Helm structure unchanged while making the new properties easier to configure in container/Kubernetes deployments.

SebastienDegodez (approved):

Issue #2058, opened 2026-04-29
Review and fix quality issues raised by Sonar on `1.14.x` for `1.15.0`


lbroudoux:
See https://sonarcloud.io/summary/overall?id=microcks_microcks&branch=master

Include some chore PR here if you'd like to contribute!

dipak0000812:
Hi @lbroudoux,

Thanks for opening this umbrella task. I’d like to help with one of the smaller maintainability fixes from the Sonar list.

I’m planning to start with a focused, low-risk cleanup in the commons module (one isolated issue with no functional behavior change), so the PR stays easy to review and validate.

I’ll comment again with the exact item once I’ve prepared the patch. Please let me know if there’s any specific issue from the list you’d prefer contributors to prioritize first.

lbroudoux:
No real priority. You can start with low hanging fruits.

Heyy-Himanshuu:
@lbroudoux Picking up the "unnamed pattern" findings (rule java:S????) across RandomStringELFunction.java, EvaluationContextTest.java, and TemplateEngineTest.java.

abhaygoudannavar:
Hi @lbroudoux 
Thanks again for the feedback on my previous PR regarding the SonarCloud cleanup. You were totally right about isolating the changes.I’ve gone ahead and closed that mixed PR and made up some time and split the work into 4 small, single-purpose PRs to make reviewing much easier.

They tackle a few different code smells/bugs: 
1. Fixing the indexOf bounds check for charset parsing #2101. 
2. Adding a missing default case to the McpController switch #2102. 
3. Removing an empty test class #2103. 
4. Cleaning up some leftover System.out.println debug statements in the test files #2104.

Let me know if this structure works better

PR #2060, opened 2026-04-30
chore: #2058 Clean up ExpressionParser.validateSuffixFound

Heyy-Himanshuu:
Address two Sonar maintainability findings on ExpressionParser:

- Remove unused 'expressionPrefix' parameter from the private validateSuffixFound helper and update its single call site.
- Extract template.substring(prefixIndex) into a local 'unparsed' variable so it is computed once and reused for both the log message and the ParseException message, avoiding the eager invocation that was being passed as a logger argument.

Behaviour is unchanged.

### Description

- `commons/util-el` — `ExpressionParser.validateSuffixFound`: removed the unused `expressionPrefix` parameter and updated the single private call site (rule: unused method parameter, L100).
- `commons/util-el` — `ExpressionParser.validateSuffixFound`: extracted `template.substring(prefixIndex)` into a local `unparsed` variable, eliminating the duplicate substring call and the eager method invocation passed as a logger argument (rule: invoke method only conditionally, L104).
- No public API surface changed; `validateSuffixFound` is `private static` with a single caller inside `parseExpressions`.
- `mvn spotless:check -pl commons/util-el` passes locally; existing tests in `commons/util-el` exercise both branches of `validateSuffixFound` (missing-suffix path) so behaviour is covered by the current suite.

### Related issue(s)

See also #2058

### DIFF
--- a/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
+++ b/commons/util-el/src/main/java/io/github/microcks/util/el/ExpressionParser.java
@@ -68,7 +68,7 @@ public class ExpressionParser {
             int afterPrefixIndex = prefixIndex + expressionPrefix.length();
             int suffixIndex = skipToCorrectEndSuffix(expressionSuffix, template, afterPrefixIndex);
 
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
LGTM. Thank you!

PR #2059, opened 2026-04-30
chore: #2058 Use unnamed patterns in commons exception handlers

Heyy-Himanshuu:
### Summary
Replace unused exception variables (nfe, t) with the Java 22+ unnamed pattern (_) across the commons/util-el module, addressing 6 of the maintainability findings raised by SonarCloud and aggregated under #2058.

### Sites changed
NowELFunction#isInteger
RandomIntELFunction#evaluate (two catches)
RandomStringELFunction#evaluate
TemplateEngineTest#testContextlessTemplate
TemplateEngineTest#testPostmanNotationCompatibility
Each catch body either has only a comment or calls fail(...) with a hard-coded message — the exception variable was never read.

### Notes
This rule (java:S???? for unnamed patterns) became applicable after the Java 25 upgrade in #2052 / #2057.
I did not include EvaluationContextTest's newInstance Sonar finding in this PR — it's a different rule with non-trivial replacement (Class.getDeclaredConstructor().newInstance() introduces additional checked exceptions). Happy to do it as a follow-up.

PR #2061, opened 2026-04-30
chore: #2058 Catch Exception instead of Throwable in AsyncAPITestManager

Heyy-Himanshuu:
### Description

- `minions/async` — `AsyncAPITestManager#runTest`: outer catch around `executorService.invokeAny` narrowed from `Throwable` to `Exception`; variable renamed `t` → `e` and the log message and test-return reason text updated to match (rule: `java:S1181`, L141).
- `minions/async` — `AsyncAPITestManager#runTest`: finally-block catch around `messageConsumptionTask.close()` narrowed from `Throwable` to `Exception`. The variable is unused (best-effort cleanup), so the unnamed pattern `_` is used (JEP 456, Java 22+), matching the convention introduced in #2059 (rule: `java:S1181`, L150).
- No behavioural change for the normal `Exception` path. JVM `Error`s will now propagate instead of being silently turned into `FAILURE` test returns, which is the intended behaviour for fatal conditions like `OutOfMemoryError`.
- `mvn spotless:check -pl minions/async` passes locally.

### Related issue(s)

See also #2058


Heyy-Himanshuu:
@lbroudoux

lbroudoux (approved):
LGTM

PR #2062, opened 2026-04-30
chore: #2058 Re-interrupt thread when catching InterruptedException

Heyy-Himanshuu:
### Description

- `minions/async` — `AsyncAPITestManager.AsyncAPITestThread#run` (L129): re-interrupt after logging that the test thread was interrupted.
- `minions/async` — `NATSMessageConsumptionTask#close` (L137): re-interrupt after logging the close-time exception so callers can still observe cancellation.
- `webapp` — `MockControllerCommons#waitForDelay` (L350): narrow the overly broad `catch (Exception)` to `catch (InterruptedException)` — `Object#wait()` only declares `InterruptedException`, and the surrounding `synchronized (semaphore)` block guarantees we hold the monitor, so the catch was wider than necessary — then re-interrupt.
- No behavioural change on the normal exception-free path. Interrupted threads will now correctly propagate the interrupt signal up the stack.

### Related issue(s)

See also #2058


lbroudoux (approved):
LGTM. Thank you!

PR #2064, opened 2026-05-01
chore: #2058 Guard Optional access with isPresent() before get()

Heyy-Himanshuu:
Sonar flagged four call sites that invoke `Optional#get()` without a preceding presence check (rule `java:S3655`). Any of these would throw `NoSuchElementException` at runtime if the Optional is empty. Each site is now guarded using the codebase's existing convention of `Optional<T> xxxOpt = ...; if (xxxOpt.isPresent()) { ... }`.

### Description

- `minions/async` — `KafkaProducerManager#renderEventMessageHeaders` (L226): wrap header value extraction in an `isPresent` guard so headers with no values are simply skipped instead of crashing the producer.
- `webapp` — `TestRunnerService#runTests` (L200): if a `TestCaseResult` references an operation that no longer exists on the service, log a warning and skip that test case rather than aborting the entire test run with `NoSuchElementException`.
- `webapp` — `ExamplesExporter#exportHeaders` (L174): wrap header value extraction in an `isPresent` guard so headers with no values are omitted from the exported document instead of crashing the export.
- `webapp` — `SchemaConformanceAssertion#assertResponse` (L63): if no WSDL resource is available for a SOAP service, log a warning and return `AssertionStatus.FAILED` with a clear error message rather than crashing the assertion.
- Happy path is unchanged in every case; only the previously unreachable empty-Optional path now degrades gracefully.

### Related issue(s)

See also #2058

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
@@ -223,16 +223,19 @@ public class KafkaProducerManager {
 
          for (io.github.microcks.domain.Header header : headers) {
             // For Kafka, header is mono valued so just consider the first value.
-            String firstValue = header.getValues().stream().findFirst().get();
-            if (firstValue.contains(TemplateEngine.DEFAULT_EXPRESSION_PREFIX)) {
-               try {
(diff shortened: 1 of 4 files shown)


lbroudoux (approved):
Awesome! Thanks a lot! LGTM.

PR #2065, opened 2026-05-01
chore: #2058 Override equals/hashCode/toString in ResponseResult

Heyy-Himanshuu:
`ResponseResult` is a Java record with a `byte[] content` field. Java's auto-generated record accessors use reference semantics on arrays — two records with byte-for-byte identical content are **not** `.equals()` to each other, their `hashCode()` values diverge, and `toString()` prints `content=[B@1a2b`. Sonar flags this with rule `java:S6218` ("override equals/hashCode/toString to consider array's content").

### Description

- `equals`: compare `status` and `headers` via `Objects.equals` and `content` via `Arrays.equals`, so two records with identical bytes compare equal.
- `hashCode`: combine `Objects.hash(status, headers)` with `Arrays.hashCode(content)`, so equal records share a `hashCode`.
- `toString`: print the content as `byte[<length>]` rather than dumping the whole array (which may be megabytes for binary responses) or emitting the unhelpful default `[B@hashref`.
- No behavioural change for callers that don't use equality, hashing, or `toString` on `ResponseResult`; strict improvement everywhere else.
- `mvn spotless:check -pl webapp` passes locally.

### Related issue(s)

See also #2058

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/ResponseResult.java
+++ b/webapp/src/main/java/io/github/microcks/web/ResponseResult.java
@@ -18,6 +18,9 @@ package io.github.microcks.web;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatusCode;
 
+import java.util.Arrays;
+import java.util.Objects;
+
 /**
  * A thin wrapper around a response result.
  * @param status  The HTTP status code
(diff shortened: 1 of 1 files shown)


lbroudoux:
That's a really good catch! Thanks for spending time on this!

lbroudoux (approved):
LGTM

PR #2067, opened 2026-05-02
chore: #2058 Handle the boolean returned by File.delete

Heyy-Himanshuu:
Sonar flagged four cleanup sites that call `File.delete()` but ignore the boolean return value (rule `java:S899`). When `delete()` returns false the file silently survives, which over time produces disk leaks of trust stores, resolved references, and uploaded artifacts. Each site now logs a warning when `delete()` fails so the leak is diagnosable from production logs.

### Description

- `minions/async` — `WebSocketMessageConsumptionTask#close` (L128): warn when the WebSocket trust store cannot be deleted.
- `webapp` — `ReferenceResolver#cleanResolvedReferences` (L187): warn for each resolved reference file that fails to delete.
- `webapp` — `UploadArtifactController#importArtifactFromUrl` (L111): warn when the temporary artifact downloaded from a URL cannot be deleted in the `finally` block.
- `webapp` — `UploadArtifactController#importArtifact` (L156): same for the multipart-uploaded artifact.
- `UploadArtifactController` uses a `SafeLogger` whose `warn(...)` overloads only accept `(String)` and `(String, Throwable)`, so the warning is emitted via plain string concatenation rather than slf4j placeholder substitution.
- Behaviour is unchanged on the happy path; failed deletions previously produced no signal at all.

### Related issue(s)

See also #2058

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
+++ b/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
@@ -107,8 +107,8 @@ public class UploadArtifactController {
             return new ResponseEntity<>(mrie.getMessage(), HttpStatus.BAD_REQUEST);
          } finally {
             // Cleanup and remove local file.
-            if (localFile != null) {
-               localFile.delete();
+            if (localFile != null && !localFile.delete()) {
+               log.warn("Failed to delete temporary artifact file " + localFile.getAbsolutePath());
             }
          }
 
@@ -153,7 +153,10 @@ public class UploadArtifactController {
          } finally {
             // Cleanup and remove local file.
             if (localFile != null) {
-               Paths.get(localFile).toFile().delete();
+               File tempFile = Paths.get(localFile).toFile();
+               if (!tempFile.delete()) {
+                  log.warn("Failed to delete temporary artifact file " + tempFile.getAbsolutePath());
+               }
             }
          }
 
--- a/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
+++ b/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
@@ -108,7 +108,7 @@ public class UploadArtifactController {
          } finally {
             // Cleanup and remove local file.
             if (localFile != null && !localFile.delete()) {
-               log.warn("Failed to delete temporary artifact file " + localFile.getAbsolutePath());
+               log.warn("Failed to delete temporary artifact file {}", localFile.getAbsolutePath());
             }
          }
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
```diff
@@ -107,8 +107,8 @@ public ResponseEntity<String> importArtifact(@RequestParam(value = "url", requir
             return new ResponseEntity<>(mrie.getMessage(), HttpStatus.BAD_REQUEST);
          } finally {
             // Cleanup and remove local file.
-            if (localFile != null) {
-               localFile.delete();
+            if (localFile != null && !localFile.delete()) {
+               log.warn("Failed to delete temporary artifact file " + localFile.getAbsolutePath());
```
lbroudoux: Shouldn't you use the template form with `{}` to be substituted by localFile.getAbsolutePath()?

webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
```diff
@@ -153,7 +153,10 @@ public ResponseEntity<String> importArtifact(@RequestParam(value = "file") Multi
          } finally {
             // Cleanup and remove local file.
             if (localFile != null) {
-               Paths.get(localFile).toFile().delete();
+               File tempFile = Paths.get(localFile).toFile();
+               if (!tempFile.delete()) {
+                  log.warn("Failed to delete temporary artifact file " + tempFile.getAbsolutePath());
```
lbroudoux: Shouldn't you use the template form with {} to be substituted by localFile.getAbsolutePath()?

lbroudoux (changes requested):
Just a small change to avoid future issues about not using the templatized form of log messages. Thanks a lot!

webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
Heyy-Himanshuu: Good point the reason I went with concatenation in the first commit was that `SafeLogger.warn` didn't expose any placeholder overloads (it only had `(String)` and `(String, Throwable)`). The placeholder forms existed for `debug`, `info` and `error`, but the matching three overloads for `warn` were missing.

Pushed `bc989702` which:

- Adds `warn(String, Object)`, `warn(String, Object, Object)` and `warn(String, Object...)` to `SafeLogger`, each routed through the existing `encode(...)` sanitiser exactly like the equivalent `info` overloads, so user-controlled data is still newline-encoded before being passed down to the underlying SLF4J logger.
- Switches the two `File.delete` warning sites in `UploadArtifactController` to the `{}` templated form now that the API supports it.

Both call sites now match the file's existing convention. Thanks for catching this. the asymmetry in the API was something to fix anyway.

lbroudoux: Awesome! Thanks!

lbroudoux (approved):
LGTM

PR #2068, opened 2026-05-02
chore: #2058 Override hashCode in StringToStringsMap to mirror equals

Heyy-Himanshuu:
`StringToStringsMap` extends `HashMap` and overrides `equals(Object)` with a mode switch driven by the `equalsOnThis` flag  reference equality when `true`, otherwise the inherited `HashMap` content equality. `hashCode()` was never overridden though, so it always returned HashMap's content-based hash.

Sonar flags this with rule `java:S1206`, and beyond the rule the behaviour is genuinely fragile: when `equalsOnThis` is `true`, two distinct instances with identical content satisfy `hashCode() == hashCode()` but `equals()` returns `false`, which produces surprising results inside hash-based collections.

### Description

- Added `hashCode()` override that mirrors the equals mode:
  - When `equalsOnThis = true`: return `System.identityHashCode(this)` so the hash is consistent with reference equality.
  - When `equalsOnThis = false`: delegate to `super.hashCode()` to stay consistent with the inherited content equality.
- Added the missing `@Override` annotation on the existing `equals` method so the pair is explicit.
- `mvn spotless:check -pl webapp` passes locally.

### Related issue(s)

See also #2058

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/script/StringToStringsMap.java
+++ b/webapp/src/main/java/io/github/microcks/util/script/StringToStringsMap.java
@@ -92,8 +92,14 @@ public class StringToStringsMap extends HashMap<String, List<String>> {
       this.equalsOnThis = equalsOnThis;
    }
 
+   @Override
    public boolean equals(Object o) {
       return this.equalsOnThis ? this == o : super.equals(o);
+   }
+
+   @Override
+   public int hashCode() {
+      return this.equalsOnThis ? System.identityHashCode(this) : super.hashCode();
    }
 
    public String[] getKeys() {
--- a/webapp/src/test/java/io/github/microcks/util/script/StringToStringsMapTest.java
+++ b/webapp/src/test/java/io/github/microcks/util/script/StringToStringsMapTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


Vaishnav88sk:
The `identityHashCode` approach is correct for the `equalsOnThis = true` mode, but is there actually a test that puts a `StringToStringsMap` with `equalsOnThis = true` into a `HashMap` or `HashSet` and verifies the contract holds? Without that, this is just Sonar-silencing, the broken behavior that originally existed would still be easy to reintroduce. What are your thoughts?

Heyy-Himanshuu:
makes sense, the override alone is just Sonar-silencing without a regression test to lock the contract in. let me add test for it.

Heyy-Himanshuu:
@Vaishnav88sk added. Have a look and let me know if any changes are required.

lbroudoux:
We have to be careful here as the `StringToStringsMap` is actually a port from the original SoapUI library to avoid us the pain of pulling this outdated lib. As a consequence, even if incorrect or fragile, we should keep it "as is" to avoid differences from the behaviour people can see in SoapUI. We may check the latest version of this class (don't know where is the original but there are some copies around there like in https://github.com/oysteing/soapui/blob/master/src/java/com/eviware/soapui/support/types/StringToStringsMap.java) to see how it evolves and then it may be better to ignore this issue at the Sonar level... Let's check.

Heyy-Himanshuu:
@lbroudoux good call! checked the linked SoapUI source and confirmed: SoapUI's upstream `StringToStringsMap` has the same `equalsOnThis` flag and `equals` override, and does **not** override `hashCode` either. So Microcks's current behavior already matches SoapUI exactly. Merging this PR would introduce a behavioral divergence between Microcks-hosted SoapUI scripts and real SoapUI, which defeats the purpose of the port.

Closing this PR. Will follow up with a small PR that suppresses the Sonar finding at the source (e.g. `@SuppressWarnings("java:S1206")` with a comment pointing to the SoapUI port rationale) so future contributors don't keep tripping over it.

PR #2070, opened 2026-05-02
fix: NPE risk in `ExamplesExporter.exportHeaders()` when header values are empty

ashnaaseth2325-oss:
### Summary

Prevent a runtime exception when exporting headers by safely handling empty header value sets in `ExamplesExporter`. This updates `exportHeaders()` to avoid unsafe `Optional#get()` usage.

---

### Sites changed

ExamplesExporter#exportHeaders

---

### Notes

Replaces unsafe `findFirst().get()` with `ifPresent(...)` to avoid `NoSuchElementException` when a header has no values. Behavior remains unchanged for valid inputs; empty headers are now safely skipped instead of causing a crash.

See #2058


ashnaaseth2325-oss:
Hello @yada @lbroudoux,
This PR fixes empty header values crash in `exportHeaders()` by replacing unsafe `Optional#get()` with `ifPresent()`

lbroudoux (approved):
LGTM

PR #2071, opened 2026-05-02
refactor: improve dependency injection and error handling in async producer managers

kishore08-07:
### Description

This PR addresses 13 SonarQube maintainability and reliability violations across three classes in `minions/async/src/main/java/io/github/microcks/minion/async/producer/`.

- Replaces all `@Inject` field injections with constructor injection
- Reduces cognitive complexity in `triggerAsyncMockMessages` by extracting the inner binding dispatch into a private helper method
- guards unsafe Optional access
- replaces broad `Throwable` catches with `Exception`, and returns an empty
collection instead of null.
-  A `ProducerDependencies` holder class is introduced to keep the `ProducerManager` constructor within the 7-parameter limit while preserving backward compatibility for existing tests via a suppressed 9-parameter constructor.
- passes `mvn spotless:check -pl minions/async`.

### Related issue(s)

See also #2058

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/KafkaProducerManager.java
@@ -61,8 +61,20 @@ public class KafkaProducerManager {
 
    private Producer<String, GenericRecord> registryProducer;
 
-   @Inject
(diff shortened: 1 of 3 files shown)


lbroudoux (approved):
Thanks for this refactoring. This looks all good to me!

PR #2072, opened 2026-05-03
chore: hide implicit constructor in TemplateEngineFactory

dipak0000812:
Refs #2058

Fix Sonar rule for utility classes by preventing instantiation of TemplateEngineFactory.

- Add private constructor to hide implicit public constructor
- Align with utility class design pattern

No functional changes.

### DIFF
--- a/commons/util-el/src/main/java/io/github/microcks/util/el/TemplateEngineFactory.java
+++ b/commons/util-el/src/main/java/io/github/microcks/util/el/TemplateEngineFactory.java
@@ -23,6 +23,11 @@ import io.github.microcks.util.el.function.*;
  */
 public class TemplateEngineFactory {
 
+   private TemplateEngineFactory() {
+      // Utility class 
+   }
+
+
    /**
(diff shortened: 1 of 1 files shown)


commons/util-el/src/main/java/io/github/microcks/util/el/TemplateEngineFactory.java
```diff
@@ -23,6 +23,11 @@
  */
 public class TemplateEngineFactory {
 
+   private TemplateEngineFactory() {
+      throw new IllegalStateException("Utility class");
```
lbroudoux: I understand that the final result will be the same, but, for better consistency with other classes in the same case, I would prefer just to put a comment here.

lbroudoux (changes requested):
Thanks for pushing this. I'd prefer a slightly different approach for better consistency with code base. Thanks!

dipak0000812:
Thanks for the suggestion. Updated the constructor to align with existing code style by removing the exception and adding a comment instead.

lbroudoux (approved):
LGTM

PR #2073, opened 2026-05-03
chore: handle return value of File.delete in ReferenceResolver

dipak0000812:
Refs #2058

Fix Sonar issue by handling the return value of File.delete().

- Check deletion success
- Log warning if deletion fails

No functional changes.


lbroudoux:
Sorry, I have to close this one as I already merged another PR with the same fix.

PR #2074, opened 2026-05-03
chore: handle return value of File.delete in UploadArtifactController

dipak0000812:
Refs #2058

Fix Sonar issue by handling the return value of File.delete().

- Check deletion success
- Log warning if deletion fails

No functional changes.

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
+++ b/webapp/src/main/java/io/github/microcks/web/UploadArtifactController.java
@@ -108,7 +108,9 @@ public class UploadArtifactController {
          } finally {
             // Cleanup and remove local file.
             if (localFile != null) {
-               localFile.delete();
+               if (!localFile.delete()) {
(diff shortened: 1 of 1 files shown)


Vaishnav88sk:
When deletion fails, a warning is logged, but the temp file is still left on disk. Is that intentional? In a high-traffic upload scenario this could quietly accumulate leftover files. Should we at least mention `deleteOnExit()` as a fallback, or is there a cleanup job somewhere already handling this?

dipak0000812:
> When deletion fails, a warning is logged, but the temp file is still left on disk. Is that intentional? In a high-traffic upload scenario this could quietly accumulate leftover files. Should we at least mention `deleteOnExit()` as a fallback, or is there a cleanup job somewhere already handling this?

Good point. Added a fallback using deleteOnExit() to avoid accumulation of temporary files in case deletion fails. This keeps behavior safe without introducing additional cleanup complexity.

lbroudoux:
Sorry, I have to close this one as I already merged another PR with same fix.

PR #2082, opened 2026-05-04
refactor: constructor injection, literal constants, and code cleanup across repository layer

kishore08-07:
### Description
This PR resolves maintainability issues flagged by SonarQube across five repository classes: `GenericResourceRepositoryImpl`, `TestConformanceMetricRepositoryImpl`, `DailyStatisticRepositoryImpl`, `ImportJobRepositoryImpl`, and `ServiceRepositoryImpl`.

## Critical
- Extract duplicated literals into named constants across `TestConformanceMetricRepositoryImpl`, `DailyStatisticRepositoryImpl`, `ImportJobRepositoryImpl`, and `ServiceRepositoryImpl` (S1192)

## Major

- Replace field injection with constructor injection across all five classes (S6813)
- Remove unused private log field in `GenericResourceRepositoryImpl` and `ImportJobRepositoryImpl` (S1068)
- Remove commented-out code block in `DailyStatisticRepositoryImpl` (S125)
- Replace string concatenation with format specifiers in log statements in `DailyStatisticRepositoryImpl` (S2629)
- Replace keySet() iteration with entrySet() to eliminate redundant map lookups in `ImportJobRepositoryImpl` and `ServiceRepositoryImpl` (S2864)

## Minor

- Replace deprecated MapReduceResults with aggregation pipeline equivalent in `DailyStatisticRepositoryImpl` (S1874)
- Inline unnecessary temporary variables across `ImportJobRepositoryImpl` and `ServiceRepositoryImpl` (S1488)
- Replace explicit generic types with diamond operator (S2293)

### Related issue(s)

See also #2058


webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
```diff
@@ -70,25 +75,34 @@ public void incrementDailyStatistic(String day, String serviceName, String servi
...
+      ProjectionOperation projectOp = project(DAILY_COUNT_FIELD).and("_id").as("day");
+      for (int i = 0; i < 24; i++) {
+         projectOp = projectOp.and("h" + i).as("hourlyCount." + i);
+      }
+
+      Aggregation aggregation = newAggregation(match(Criteria.where("day").is(day)), groupOp, projectOp);
 
-      // Execute a MapReduce command.
-      MapReduceResults<WrappedDailyStatistic> results = template.mapReduce(query, "dailyStatistic",
-            "classpath:mapDailyStatisticForADay.js", "classpath:reduceDailyStatisticForADay.js",
```
lbroudoux, line 77: Shouldn't you remove those files if they are no longer needed?

webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
```diff
@@ -70,25 +75,34 @@ public void incrementDailyStatistic(String day, String serviceName, String servi
...
+      GroupOperation groupOp = group("day").sum(DAILY_COUNT_FIELD).as(DAILY_COUNT_FIELD);
+      for (int i = 0; i < 24; i++) {
+         groupOp = groupOp.sum("hourlyCount." + i).as("h" + i);
+      }
+
+      ProjectionOperation projectOp = project(DAILY_COUNT_FIELD).and("_id").as("day");
+      for (int i = 0; i < 24; i++) {
+         projectOp = projectOp.and("h" + i).as("hourlyCount." + i);
+      }
+
```
lbroudoux, line 84: Have you compared the results with the previous version, especially checking the statistics diagram?

lbroudoux (commented):
Looks good to me but I think some other files need to be removed if you refactored the mapredue to an aggregation. Also, can you explain how you checked it produces the same result and doesn't introduce any regression?

webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
kishore08-07: Yes,removed both MapReduce JS files (`mapDailyStatisticForADay.js`, `reduceDailyStatisticForADay.js`) since the repository no longer uses MapReduce. Also removed the `WrappedDailyStatistic` class which was only used for MapReduce results.

kishore08-07:
> Looks good to me but I think some other files need to be removed if you refactored the mapredue to an aggregation. Also, can you explain how you checked it produces the same result and doesn't introduce any regression?

Yes, checked that aggregation is semantically identical with mapreduce and i validated its equivalence via:
- **Unit tests with assertions** : dailyCount sum, hourlyCount[0..23] sums across services, and empty-day behavior.
- **MongoDB query comparison:** ran the equivalent `$match` + `$group` aggregation in Mongo shell and compared results with `/api/metrics/invocations/global?day=...`

webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
kishore08-07: I verified the `aggregation` behavior against the prior `MapReduce` semantics by tightening `CustomDailyStatisticRepositoryTest`. Previously, `testAggregateDailyStatistics()` only ensured no exception was thrown. I added assertions to validate:

the aggregated day and dailyCount values, the presence of a 24-entry `hourlyCount` map, and correct hourly summation in a new case with non‑zero buckets.
All tests pass, confirming the aggregation output matches the expected results and should not regress the statistics diagram.

kishore08-07:
Thanks @lbroudoux  for your review and i have address the feedbacks, kindly re-review whenever you have time.

webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
```diff
@@ -15,51 +15,56 @@
  */
 package io.github.microcks.repository;
 
+import static org.springframework.data.domain.Sort.Direction.ASC;
+import static org.springframework.data.mongodb.core.aggregation.Aggregation.*;
```
lbroudoux: Can you keep original import order and move these ones to their original location?

lbroudoux (changes requested):
LGTM but need to restore the original import order before merging please.

webapp/src/main/java/io/github/microcks/repository/DailyStatisticRepositoryImpl.java
kishore08-07: Sure, I have moved those imports back to their original location and have also included the formatting fix that was causing the Spotless CI build failure.

lbroudoux (approved):
LGTM

PR #2084, opened 2026-05-04
fix: #2058 prevent possible NPE in JobController start/stop endpoints

dipak0000812:
Fix Sonar reliability issues (S2259 - possible NPE) in JobController.

Changes:
- Added null checks before accessing ImportJob in startJob() and stopJob()

Impact:
- Prevents potential NullPointerException when the job is not found
- No functional behavior change
- Aligns with existing null-check pattern used in activateJob()

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/JobController.java
+++ b/webapp/src/main/java/io/github/microcks/web/JobController.java
@@ -169,8 +169,8 @@ public class JobController {
    public ResponseEntity<ImportJob> startJob(@PathVariable("id") String jobId, UserInfo userInfo) {
       log.debug("Starting job with id {}", jobId);
       ImportJob job = jobRepository.findById(jobId).orElse(null);
-      if (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
-            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job)) {
+      if (job != null && (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
+            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job))) {
          job.setActive(true);
          initMetadataIfMissing(job);
          job.getMetadata().objectUpdated();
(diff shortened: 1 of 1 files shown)


webapp/src/main/java/io/github/microcks/web/JobController.java
```diff
@@ -169,8 +169,8 @@ public ResponseEntity<ImportJob> activateJob(@PathVariable("id") String jobId, U
    public ResponseEntity<ImportJob> startJob(@PathVariable("id") String jobId, UserInfo userInfo) {
       log.debug("Starting job with id {}", jobId);
       ImportJob job = jobRepository.findById(jobId).orElse(null);
-      if (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
-            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job)) {
+      if (job != null && (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
```
SebastienDegodez: I think better in this case to return an httpresult is ‘notfound’.

I suggest an early return on job.

webapp/src/main/java/io/github/microcks/web/JobController.java
```diff
@@ -184,8 +184,8 @@ public ResponseEntity<ImportJob> startJob(@PathVariable("id") String jobId, User
    public ResponseEntity<ImportJob> stopJob(@PathVariable("id") String jobId, UserInfo userInfo) {
       log.debug("Stopping job with id {}", jobId);
       ImportJob job = jobRepository.findById(jobId).orElse(null);
-      if (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
-            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job)) {
+      if (job != null && (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
```
SebastienDegodez: I think better in this case an httpresult is ‘notfound’.

I suggest an early return on job.

SebastienDegodez:
I think it would be interesting to be able to test the scenario with a in memory repository of the jobs.

SebastienDegodez (changes requested):

webapp/src/main/java/io/github/microcks/web/JobController.java
lbroudoux: Agreed with @SebastienDegodez's suggestion! +1

lbroudoux (approved):
LGTM

lbroudoux (approved):
LGTM

PR #2086, opened 2026-05-05
feat(openapi): #2037 Update overlay export shape for standard mergers

### LINKED ISSUE #2037: Overlay format for export examples | Proposal
### Reason/Context

In the context of generating examples via the copilot I want to extract them back, to be finally merged with my originating specification. The goal being to upload the merged version into my API portal.

 

### Description

Hello,

When extracting the examples generated via copilot with the overlay format, I was struggling to make them work with the `github.com/speakeasy-api/openapi/overlay/loader` Goland utility (to merge them back to the specification).  

While investigating the issue I finally tried to amend the overlay output format to comply with a new form given below.

```yaml
  - target: "$.paths['/my-endpoint'].post.requestBody.content['application/json']"
    update:
      examples:
        sample-string-returned:
          value: "XYZ"
          summary: Sample String Returned
```
This new format works better when no examples exists in the target.

Please note that the I had to convert the summary to output the name, somehow to be discussed.

Thoughts ?

### Implementation ideas

Code snippet below.

```java
   private void exportResponse(ArrayNode actions, Operation operation, Response response) {
      // Export the response part of the exchange.
      ObjectNode responseNode = actions.addObject();

      String[] operationParts = operation.getName().split(" ");

      responseNode.put("target", "$.paths['" + operationParts[1] + "']." + operationParts[0].toLowerCase()
            + ".responses['" + response.getStatus() + "'].content['" + response.getMediaType() + "']");
      createExampleUpdate(responseNode, toExampleKey(response.getName()), response.getName(), response.getContent());
   }

   private void createExampleUpdate(ObjectNode actionNode, String exampleKey, String summary, String value) {
      ObjectNode updateNode = actionNode.putObject("update");
      ObjectNode examplesNode = updateNode.putObject("examples");
      ObjectNode exampleNode = examplesNode.putObject(exampleKey);
      exampleNode.put("value", value);
      exampleNode.put("summary", summary);
   }

   private String toExampleKey(String name) {
      return name.toLowerCase().replace(" ", "-");
   }
```

Heyy-Himanshuu:
## Summary

Resolves #2037.

The previous overlay export targeted ....examples directly and placed each example as a top-level key under update. Standard OpenAPI overlay mergers (e.g. https://github.com/speakeasy-api/openapi/overlay/loader) struggle with this structure when the target node does not already contain an examples map, since the merger cannot create intermediate paths.

This PR restructures each emitted action so that:
- The target points to the parent node (parameter, requestBody content type, or response content type)
- The update explicitly includes an examples wrapper

This allows mergers to deep-merge cleanly, regardless of whether examples already exists.

### New format
```
yaml - target: "$.paths['/p'].post.requestBody.content['application/json']"   update:     examples:       my-example:         value: ...         summary: My Example 
```

### Backward Compatibility
Replaced the existing format in-place instead of introducing a new exporter (OPENAPI_OVERLAY_V2).

Reason:
- The previous format was incompatible with standard overlay mergers
- The new format remains compliant with the same overlay: 1.0.0 specification

---

### Summary Source
- Uses the request/response name
- Microcks does not currently store a separate description for examples
- Name provides the best available semantic value

---

### Slugification
Uses:
java name.toLowerCase().replace(" ", "-") 

- Matches the original proposal
- YAML is serialized via Jackson → UTF-8 safe
- Non-ASCII characters (e.g. Éclair Café → éclair-café) are preserved correctly

---

### Existing Examples Handling
- Shape correctness is validated via snapshot tests (OpenAPIOverlayExporterTest)
- Merge behavior is delegated to the overlay merger
- Exporter responsibility is limited to emitting a valid structure

---

## Notable Cleanups

- Extracted shared logic into:
  java   createExampleUpdate(actionNode, name, value)   
  This removes duplication across:
  - Query parameters
  - Request bodies
  - Responses

- Updated response path format:
    .responses.200.content[...]    → .responses['200'].content[...]  
  This aligns with consistent JSONPath bracket notation.

---

## Test Plan

- Updated OpenAPIOverlayExporterTest
  - All EXPECTED_EXPORT_ACTION_* snapshots rewritten to match new structure
  - Snapshot tests ensure future regressions are caught

- Verified locally:
    mvn spotless:check -pl webapp  

- Post-merge validation:
  - Export overlay from Microcks
  - Merge using speakeasy-api/openapi/overlay/loader
  - Confirm:
    - New examples are added
    - Existing examples remain intact

---

## Related Issues

- Resolves #2037

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
+++ b/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
@@ -113,32 +113,27 @@ public class OpenAPIOverlayExporter implements MockRepositoryExporter {
    }
 
    private void exportRequest(ArrayNode actions, Operation operation, Request request) {
+      String[] operationParts = operation.getName().split(" ");
+      String pathPart = operationParts[1];
+      String methodPart = operationParts[0].toLowerCase();
+
       // Export the parameters part of the exchange if present.
       if (request.getQueryParameters() != null && !request.getQueryParameters().isEmpty()) {
          for (Parameter parameter : request.getQueryParameters()) {
             // Export the query parameter part of the exchange.
             ObjectNode parameterNode = actions.addObject();
-
-            String[] operationParts = operation.getName().split(" ");
-
-            parameterNode.put("target", "$.paths['" + operationParts[1] + "']." + operationParts[0].toLowerCase()
-                  + ".parameters[?@.name=='" + parameter.getName() + "'].examples");
-            ObjectNode updateNode = parameterNode.putObject("update");
-            ObjectNode exampleNode = updateNode.putObject(request.getName());
-            exampleNode.put("value", parameter.getValue());
+            parameterNode.put("target",
+                  "$.paths['" + pathPart + "']." + methodPart + ".parameters[?@.name=='" + parameter.getName() + "']");
+            createExampleUpdate(parameterNode, request.getName(), parameter.getValue());
          }
       }
 
       // Export the requestBody part of the exchange if present.
       if (request.getContent() != null) {
          ObjectNode requestNode = actions.addObject();
-         String[] operationParts = operation.getName().split(" ");
-
-         requestNode.put("target", "$.paths['" + operationParts[1] + "']." + operationParts[0].toLowerCase()
-               + ".requestBody.content['" + getContentType(request) + "'].examples");
-         ObjectNode updateNode = requestNode.putObject("update");
-         ObjectNode exampleNode = updateNode.putObject(request.getName());
-         exampleNode.put("value", request.getContent());
+         requestNode.put("target",
+               "$.paths['" + pathPart + "']." + methodPart + ".requestBody.content['" + getContentType(request) + "']");
+         createExampleUpdate(requestNode, request.getName(), request.getContent());
       }
    }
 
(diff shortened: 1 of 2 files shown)


Heyy-Himanshuu:
@lbroudoux let me know if any changes are required.

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
```diff
@@ -154,15 +149,34 @@ private String getContentType(Request request) {
...
+      exampleNode.put("summary", name);
+   }
 
-      responseNode.put("target", "$.paths['" + operationParts[1] + "']." + operationParts[0].toLowerCase()
-            + ".responses." + response.getStatus() + ".content['" + response.getMediaType() + "'].examples");
-      ObjectNode updateNode = responseNode.putObject("update");
-      ObjectNode exampleNode = updateNode.putObject(response.getName());
-      exampleNode.put("value", response.getContent());
+   private String toExampleKey(String name) {
+      return name.toLowerCase().replace(" ", "-");
```
dipak0000812, line 180: The slug generation is pretty simple (lowercase + replace space).

Could this lead to collisions or issues with special characters?

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
```diff
@@ -154,15 +149,34 @@ private String getContentType(Request request) {
...
+    * @param name       The example's human-readable name, used both as the slugified key and as the {@code summary}.
+    * @param value      The example's value payload.
+    */
+   private void createExampleUpdate(ObjectNode actionNode, String name, String value) {
+      ObjectNode updateNode = actionNode.putObject("update");
+      ObjectNode examplesNode = updateNode.putObject("examples");
+      ObjectNode exampleNode = examplesNode.putObject(toExampleKey(name));
+      exampleNode.put("value", value);
+      exampleNode.put("summary", name);
+   }
```
dipak0000812, line 177: Good extraction of shared logic into createExampleUpdate(), improves readability and avoids duplication.

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
```diff
@@ -113,32 +113,27 @@ public String exportAsString() throws MockRepositoryExportException {
...
-            String[] operationParts = operation.getName().split(" ");
-
-            parameterNode.put("target", "$.paths['" + operationParts[1] + "']." + operationParts[0].toLowerCase()
-                  + ".parameters[?@.name=='" + parameter.getName() + "'].examples");
-            ObjectNode updateNode = parameterNode.putObject("update");
-            ObjectNode exampleNode = updateNode.putObject(request.getName());
-            exampleNode.put("value", parameter.getValue());
+            parameterNode.put("target",
+                  "$.paths['" + pathPart + "']." + methodPart + ".parameters[?@.name=='" + parameter.getName() + "']");
+            createExampleUpdate(parameterNode, request.getName(), parameter.getValue());
```
dipak0000812, line 127: This changes the export structure compared to the previous format.

Do we need versioning or a flag here, or is it safe for existing consumers?

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
```diff
@@ -113,32 +113,27 @@ public String exportAsString() throws MockRepositoryExportException {
    }
 
    private void exportRequest(ArrayNode actions, Operation operation, Request request) {
+      String[] operationParts = operation.getName().split(" ");
+      String pathPart = operationParts[1];
+      String methodPart = operationParts[0].toLowerCase();
```
dipak0000812, line 118: This relies on operation.getName().split(" "), which feels a bit fragile.
It assumes a strict "METHOD /path" format.

Might be safer to validate or avoid parsing the string if possible.

dipak0000812 (commented):
Looks like a solid improvement overall and the new structure makes sense for merging.

Left a few small comments around parsing, slug generation, and the change in export format—just want to make sure we’re safe on edge cases and existing usage.

Happy to take another look after updates 👍

Heyy-Himanshuu:
@dipak0000812 thanks for the review, however I'd like to wait for maintainer review before making any changes to current PR.

Heyy-Himanshuu:
@lbroudoux could u review this one also ?

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
lbroudoux: We have this assumption in other parts of the code, and as of today all the operations are built that way. I think we can assume it is safe for now.

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIOverlayExporter.java
lbroudoux: I think this is safe as it's just a diffferent structural form to express the same semantic patch.

lbroudoux (approved):
LGTM! Thank you.

PR #2088, opened 2026-05-05
fix: #2058 return empty map instead of null in Metadata domain

Manar-Elhabbal7:
## Description

1. Return `Collections.emptyMap()`from getAnnotations() when annotations are not set.
2. Return `Collections.emptyMap()` from getLabels() when labels are not set.
3. Keep both getters behavior consistent by always returning immutable, non-null maps to avoid extra null checks for callers.

### Related issue(s)
Part of #2058


SebastienDegodez (approved):

lbroudoux (approved):
LGTM. Thanks!

PR #2089, opened 2026-05-05
refactor #2058 : use java.nio.file.Files.delete for better error reporting

Manar-Elhabbal7:
### Description

This change switches cleanup to `java.nio.file.Files` delete so failures are more informative.
Instead of a simple true/false result, delete throws specific exceptions, which helps identify the exact reason deletion failed (for example: file missing, permission denied, or file in use).

### Related issue(s) #2058

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/consumer/WebSocketMessageConsumptionTask.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/consumer/WebSocketMessageConsumptionTask.java
@@ -124,7 +124,7 @@ public class WebSocketMessageConsumptionTask implements MessageConsumptionTask {
             logger.warn("Closing WebSocket session raised an exception", e);
          }
       }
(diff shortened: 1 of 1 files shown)


lbroudoux:
> delete throws specific exceptions, which helps identify the exact reason deletion failed (for example: file missing, permission denied, or file in use).

This change absolutely doesn't help regarding this topic. I can validate it to lower the number of Sonar issues by 1... but is that the point of raising a PR?

Manar-Elhabbal7:
> > delete throws specific exceptions, which helps identify the exact reason deletion failed (for example: file missing, permission denied, or file in use).
> 
> This change absolutely doesn't help regarding this topic. I can validate it to lower the number of Sonar issues by 1... but is that the point of raising a PR?

Thanks  a lot for the review and clarification. This is  my first time working with Sonar suggestions, so I assumed addressing the issue directly was the expected contribution as it's looks so easy . I will think again about that thanks again :)

PR #2098, opened 2026-05-07
Feat/2053 spring boot 4

### LINKED ISSUE #2053: Update to Spring Boot 4.0.6
Spring Boot 4.x represents the latest major enhancement of Spring that leverages Java 25 benefits (see #2052 as a pre-requisite). We must migrate to this version and keep the compatibility layer with Jackson 2.x as explained in #2048

Heyy-Himanshuu:
fixes #2053

### pom.xml
- Bump `spring-boot.version` 3.5.14 → 4.0.6
- Pin Jackson 2.x BOM ahead of the SB BOM (Quarkus 3 minion + commons modules still on 2.x, per your guidance)
- Drop `spring-boot-starter-aop` (project had zero AspectJ usage in main; SB4 renamed it to `-aspectj` anyway — added back as `spring-boot-starter-aspectj` later for `WithSpanAspect`)
- Add `spring-boot-starter-restclient` (`RestTemplateBuilder` moved out of `spring-boot` in SB4)
- Add `spring-boot-starter-kafka` (`KafkaProperties` moved out of `spring-boot-autoconfigure` in SB4)
- Drop `<junit.version>` override — SB4 manages Jupiter 6.0.3 + Platform 6.x consistently; the override was masking `junit-platform-launcher` riding along unpinned

### SB4 modularization — package renames (main)
- `org.springframework.boot.autoconfigure.kafka.KafkaProperties` → `org.springframework.boot.kafka.autoconfigure.KafkaProperties`
- `org.springframework.boot.autoconfigure.security.oauth2.client.servlet.OAuth2ClientAutoConfiguration` → `org.springframework.boot.security.oauth2.client.autoconfigure.OAuth2ClientAutoConfiguration` (also dropped `.servlet` — servlet+reactive merged)
- `org.springframework.boot.web.client.RestTemplateBuilder` → `org.springframework.boot.restclient.RestTemplateBuilder`

### Spring Framework 7 API churn (main)
- `HttpHeaders` no longer extends `MultiValueMap`: `containsKey()` → `containsHeader()`, `entrySet()` → `headerSet()`, `put()` → `addAll()`; 4 call sites that pass `HttpHeaders` where `Map<String,List<String>>` is expected now use `.asMultiValueMap()`
- `UriComponentsBuilder.fromHttpUrl()` → `fromUriString()`
- `HttpComponentsClientHttpRequestFactory.setConnectTimeout(int)` removed → `RequestConfig.custom().setConnectTimeout(...)`
- `RestTemplateBuilder.setReadTimeout(Duration)` → `readTimeout(Duration)`
- `KafkaProperties.Producer.buildProperties(SslBundles)` → `buildProperties()`
- `ResponseEntity` null-body constructor disambiguated in `SecretController`

### Spring Security 7 — Resource Owner Password Grant removed
Per your guidance went with option (b) — **vendored** `DefaultPasswordTokenResponseClient` + `OAuth2PasswordGrantRequest` into `io.github.microcks.security.oauth2.password.*` so the feature keeps working under SB4 and can be properly deprecated in 1.16.0. Sibling clients (`Default*TokenResponseClient` for Client Credentials and Refresh Token) got renamed to `RestClient*TokenResponseClient`.

### Test infrastructure (drop-in path, evaluation [here](https://github.com/microcks/microcks/issues/2053#issuecomment-XXXXXX))
- `TestRestTemplate` → `org.springframework.boot.resttestclient.TestRestTemplate` + `@AutoConfigureTestRestTemplate` on `AbstractBaseIT` / `MicrocksApplicationFuzz`
- `@SpyBean` → `@MockitoSpyBean` from `org.springframework.test.context.bean.override.mockito`
- `@AutoConfigureMockMvc` / `MockMvcPrint` → `org.springframework.boot.webmvc.test.autoconfigure.*`
- `ClientHttpRequestFactorySettings.Redirects` (removed) → top-level `HttpRedirects` from `org.springframework.boot.http.client`
- `spring.data.mongodb.uri` → `spring.mongodb.uri` (Mongo connection props moved out of `spring.data.*` in SB4 — testcontainer URI was silently ignored before this)

### OpenTelemetry — swapped community starter for Spring's official
- Dropped `io.opentelemetry.instrumentation:opentelemetry-spring-boot-starter` + `opentelemetry-instrumentation-bom`; pulled in `spring-boot-starter-opentelemetry`
- Spring's official starter does **not** consume the OTel SDK-autoconfigure SPI, so the legacy `@Bean AutoConfigurationCustomizerProvider` was rewired:
  - `CustomExplainTraceProcessor` is `@Component` → picked up by `OpenTelemetryTracingAutoConfiguration.spanProcessors(ObjectProvider<SpanProcessor>)`
  - `XTraceIdTextMapPropagator` exposed as `@Bean TextMapPropagator` → composed by `otelContextPropagators(ObjectProvider<TextMapPropagator>)`
- Added a local 30-line `WithSpanAspect` to replace the one the community starter brought in — the official starter intentionally doesn't ship one (rationale in the [resolved review thread](#discussion))
- `management.tracing.sampling.probability=1.0` to preserve the 100% sampling the project previously got via the SDK-autoconfigure-only `otel.traces.sampler.arg=1.0`

## Build status
✅ `build-install` — green
✅ `integration-tests` — green (61 ITs)
✅ `fuzzing-tests` — green
✅ `javadoc-check` — green
⬜ `package-native` — only runs post-merge on upstream (`event_name != 'pull_request'`). Audited locally: webapp `process-aot` runs cleanly under SB 4.0.6 + JDK 25; native profile config (`BP_JVM_VERSION=25`, builder pinned, reachability metadata wiring) unchanged. Async minion is Quarkus, decoupled.

## Follow-ups (separate issues)

- Migrate non-SSE ITs to `RestTestClient`; rewrite the 4 SSE tests with `RestClient.exchange(ExchangeFunction)` so they keep raw stream access. Then drop the `spring-boot-resttestclient` dep.
- Deprecate the vendored Password Grant in 1.16.0.


webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
```diff
@@ -19,22 +19,90 @@
...
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import io.github.microcks.web.AbstractBaseIT;
-
 /**
  * This is a test case for class JsScriptEngineBinder class.
  */
-class JsScriptEngineBinderTest extends AbstractBaseIT {
+class JsScriptEngineBinderTest {
```
lbroudoux, line 44: Not sure to understand why you changed this one...
Is the goal to remove `TestFetchController` (which is still there)?
Is it to make it a regular unit test instead of an integration one?
Moreover, I'm not sure if using ``com.sun.*` classes is a future-proof choice? Can you elaborate on this please?

webapp/src/main/java/io/github/microcks/util/otel/WithSpanAspect.java
```diff
@@ -0,0 +1,80 @@
...
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.reflect.MethodSignature;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.lang.reflect.Method;
+
+/**
+ * Weaves {@link WithSpan @WithSpan} annotations into proper tracer spans. Spring Boot 4's official OpenTelemetry
+ * starter does not bundle a {@code WithSpanAspect}, so we provide a minimal one ourselves to keep the existing
```
lbroudoux, line 36: Seems strange as it is a regression when switching from the community Oteal implementation to the SB4 starter... Any useful link on this choice from the SB community?

lbroudoux (commented):
Hey! Thanks a lot for a huge amount of work! It looks almost good to me even if I have a few questions. Can you check it out please?

webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
Heyy-Himanshuu: Good questions, sorry for not adding context in the commit message.

**Why the change:** the old test extended `AbstractBaseIT` only so `TestFetchController` could serve `/api/test-fetch*` as a fetch target for the JS scripts under test. Under SB4 the IT-style boot pulled in the full Spring context + testcontainers Mongo just to call a 3-line controller — and it was breaking during the test-infra module split. Quickest fix was to recognize this was always a unit-level test (it exercises `JsScriptEngineBinder.bindFetch()`, not the controller) and let it bring up its own in-process HTTP server.

**`TestFetchController` removal:** not the goal, but it is now orphaned (no remaining references, confirmed by grep). I'll delete it in a follow-up commit on this PR if you're happy with the refactor.

**`com.sun.net.httpserver` future-proofness:** it's the [`jdk.httpserver` module](https://docs.oracle.com/en/java/javase/25/docs/api/jdk.httpserver/module-summary.html) — an officially supported JDK module (not the `sun.misc.*` family), stable since Java 6, and the recommended way to spin up a tiny loopback HTTP server in tests without dragging in Jetty/Netty. It's used by JDK's own networking tests. Happy to switch to `WireMockServer` or `MockWebServer` if you'd rather avoid the `com.sun.*` package name aesthetically both are equivalent for this case.

If you'd prefer I revert and fix the IT-style test under SB4 instead.

webapp/src/main/java/io/github/microcks/util/otel/WithSpanAspect.java
Heyy-Himanshuu: You're right that it's a regression and an intentional. The split is by design on Spring's side:

- The community `io.opentelemetry.instrumentation:opentelemetry-spring-boot-starter` we used on SB3.5 bundled an `InstrumentationAnnotationsAutoConfiguration` + `WithSpanAspect` (both in the `internal.*` package, see [the OTel java-instrumentation source](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/internal/instrumentation/annotations/WithSpanAspect.java)). That's what wove `@WithSpan` on SB3.
- Spring's official `spring-boot-starter-opentelemetry` deliberately scopes itself to the OTel **SDK + exporters + Micrometer tracing bridge** and explicitly does not include auto-instrumentation / aspects. Spring's stance is that auto-instrumentation belongs to the OTel project and shouldn't be re-shipped — see the [Spring Boot 4 OTel observability reference](https://docs.spring.io/spring-boot/reference/actuator/observability.html) and the [SB issue spring-projects/spring-boot#43914](https://github.com/spring-projects/spring-boot/issues/43914) discussing the boundary. Mixing both starters is explicitly called out as unsupported.
- The community `WithSpanAspect` lives in an `internal` package, so we can't depend on it from user code without breaking compatibility on every OTel java-instrumentation release.

That left three options:
1. **Stay on the community starter** — keeps `@WithSpan` for free, but Spring's docs explicitly say don't mix it with their starter on SB4.
2. **Drop `@WithSpan`** and call the OTel `Tracer` API directly in the two `processInvocation` methods. Cleaner long-term but invasive (touches business code).
3. **Local 30-line aspect** — the current choice. Recreates exactly what the OTel internal aspect did, no business code changes, gated by the same `otel.instrumentation.annotations.enabled` property so it can be turned off identically.

I picked (3) because the community starter is `internal`-package and (2) is a bigger blast radius than the SB4 migration warrants. Happy to go (2) instead if you'd rather not carry the aspect — that's the most stable long-term path. (1) feels off-table given Spring's explicit guidance.

webapp/src/test/java/io/github/microcks/util/script/JsScriptEngineBinderTest.java
lbroudoux: I was not aware of `com.sun.net.httpserver` and never used it but that's a good catch!

PR #2101, opened 2026-05-07
correct indexOf bounds check for charset parsing [Fix]

abhaygoudannavar:
### Summary
Correct `indexOf` bounds check from `> 0` to `>= 0` during charset parsing, addressing 2 of the critical bug findings raised by SonarCloud and aggregated under #2058.

### Sites changed
- `OpenAPISchemaValidator#getMessageContentNode`
- `GraphQLTestRunner#buildTestReturn`

### Notes
This rule (`java:S2692` - indexOf checks should not ignore index 0) flagged logic where splitting the string via `.indexOf(";")` was erroneously ignoring the 0th index due to a strict `> 0` check.

### DIFF
--- a/commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
+++ b/commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
@@ -288,8 +288,11 @@ public class OpenAPISchemaValidator {
 
          // If no match here, it's maybe contentType contains charset but not the spec.
          // Remove charset information and try again.
-         if (contentType.contains("charset=") && contentType.indexOf(";") > 0) {
-            return getMessageContentNode(responseCodeNode, contentType.substring(0, contentType.indexOf(";")));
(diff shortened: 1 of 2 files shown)


Vaishnav88sk:
The `> 0` to `>= 0` fix is correct per `S2692` and technically sound. One thing worth thinking about though: if `indexOf(";")` does return `0` (malformed content-type like `;charset=utf-8`), whatever comes after this check - likely a `substring(0, idx)` call, would produce an empty string. The fix stops silently ignoring index 0, but doesn't make that case safe. Probably not a real concern since no valid Content-Type starts with `;`, but good to be aware of.

abhaygoudannavar:
@Vaishnav88sk I hadn't considered the edge case of a malformed Content-Type starting with a semicolon resulting in an empty string. I've pushed an updated commit that extracts the index, ensures it is != -1 (to satisfy Sonar S2692), and safely checks > 0 before extracting the substring to prevent any empty strings. Thanks for pointing that out

PR #2104, opened 2026-05-07
remove System.out.println from test classes [Fix]

abhaygoudannavar:
### Summary
Removed leftover `System.out.println` debug statements from test classes, addressing the code smell raised by SonarCloud and aggregated under #2058.

### Sites changed
- `AsyncAPISchemaValidatorTest`
- `OpenAPISchemaValidatorTest`

### Notes
This rule (`java:S106` - Standard outputs should not be used directly to log anything) flagged these debug print statements. Since these were just leftover prints inside tests that provide no long-term logging value, they have been cleanly removed (along with their enclosing `for` loops).


lbroudoux (approved):
This is a nice catch. Thank you!

PR #2115, opened 2026-05-11
refactor: sanitize log inputs and reduce complexity in service layer classes

kishore08-07:
### Description
This PR addresses maintainability and security violations across `ImportExportService`, `ProxyService`, `ServiceService` and `TestService` 

Critical:
- Decompose Brain Method at L202 in `ServiceService` to reduce cognitive complexity from 77 to 15 by extracting protocol-specific helpers (S3776)
- Reduce cognitive complexity at L588 in `ServiceService` from 18 to 15 (S3776)

Security:
- Sanitize user-controlled input in log statements across 7 occurrences in `ServiceService`, `ImportExportService` and `TestService` (S5145)
- Validate and sanitize URL and path construction from user-controlled data in `ProxyService` (S5144, S2083)

Major:
- Introduce parameter object to reduce 8-parameter method in `ServiceService` below limit (S107)
- Remove unused method parameter service in `ServiceService` and update all call sites (S1172)

Minor:
- Remove unused imports in `ImportExportService` (S1128)
- Replace named exception variables with unnamed pattern _ across 3 occurrences (S6910)
- Merge nested if statements in `TestService` (S1066)

### Related issue(s)

See also #2058


dipak0000812 (commented):
Nice cleanup overall, especially around the service layer refactors and log sanitization.
Left a couple of small comments around utility reuse and URL validation behavior, but overall this looks good to me.

webapp/src/main/java/io/github/microcks/service/ProxyService.java
```diff
@@ -73,4 +75,31 @@ public ResponseEntity<byte[]> callExternal(URI externalUrl, HttpMethod method, H
...
+         throw new IllegalArgumentException("External URL cannot be null");
+      }
+      String scheme = externalUrl.getScheme();
+      if (!"http".equalsIgnoreCase(scheme) && !"https".equalsIgnoreCase(scheme)) {
+         throw new IllegalArgumentException("Only HTTP(S) URLs are allowed");
+      }
+      if (externalUrl.getHost() == null || externalUrl.getHost().isBlank()) {
+         throw new IllegalArgumentException("External URL host is required");
+      }
+      if (externalUrl.getUserInfo() != null || externalUrl.getFragment() != null) {
```
dipak0000812, line 90: Could this be a bit restrictive for some proxy targets? just thinking about whether any existing URLs might include these parts.

webapp/src/main/java/io/github/microcks/service/TestService.java
```diff
@@ -311,11 +310,17 @@ private void waitSomeRandomMS(int min, int max) {
       long timeout = ThreadLocalRandom.current().nextInt(min, max + 1);
       try {
          Thread.sleep(Duration.ofMillis(timeout));
-      } catch (Exception e) {
+      } catch (Exception _) {
          log.debug("waitSomeRandomMS semaphore was interrupted");
       }
    }
 
+   private static String sanitize(String value) {
```
dipak0000812, line 318: Seeing the same sanitize() helper in multiple services now Maybe worth pulling into a small shared util at some point?

webapp/src/main/java/io/github/microcks/service/ProxyService.java
```diff
@@ -47,17 +48,18 @@ public class ProxyService {
     * @return The response entity returned by external service as is
     */
    public ResponseEntity<byte[]> callExternal(URI externalUrl, HttpMethod method, HttpHeaders headers, String body) {
```
lbroudoux, line 50: I think we'd better use something like Bean Validation or `@Valid` annotation here. I think this is a global topic we need to address due to the number of issues in Sonar related to this problem of input validation/sanitization.

webapp/src/main/java/io/github/microcks/service/TestService.java
```diff
@@ -119,7 +119,8 @@ public TestResult launchTests(Service service, String testEndpoint, TestRunnerTy
     * @return A completed TestCaseResult object
     */
    public TestCaseResult reportTestCaseResult(String testResultId, String operationName, List<TestReturn> testReturns) {
-      log.info("Reporting a TestCaseResult for testResult {} on operation '{}'", testResultId, operationName);
+      log.info("Reporting a TestCaseResult for testResult {} on operation '{}'", sanitize(testResultId),
```
lbroudoux, line 122: We also introduced a `SafeLogger` utility as a way to centralize this sanitization logic. However, the problem is cascading along the architecture layers ... because at no time do we transform or block invalid data format... I think this issue must be reviewed globally and not directly hidden within a PR.

webapp/src/main/java/io/github/microcks/service/TestService.java
lbroudoux: We also have a `SafeLogger` that has this kind of util but the problem also relates to validation...

lbroudoux:
We should first start thinking about why we got this issue and how to avoid cascading it along the layers ... Otherwise, we'll find out a ton of `sanitize()` method everywhere with still the same risk of having inconsistent/invalid data ...

PR #2117, opened 2026-05-12
fix(ci): #1594 remove dual Maven cache stomping local SNAPSHOT artifacts

### LINKED ISSUE #1594: ci: `package-webapp` and `integration-tests` compilation are failing when adding new classes
### Describe the bug

See https://github.com/microcks/microcks/actions/runs/14791356227

Even if the first compilation done in the `build-install` step is OK, the other ones done in subsequent steps are failing because of a missing new class.

### Expected behavior

If my understanding of the Maven caching is correct, this should work!

### Actual behavior

`build-install` compilation is done with `mvn -B clean install` so `webapp` is able to access new classes compiled in `util` and pushed in Maven local repo. Then, the Maven local repo is cached using `maven-repo-${{ github.run_id }}`.

We see in the post compilation step logs: 
```sh
Cache hit for: maven-repo-14791356227
Received 176160768 of 425392991 (41.4%), 167.8 MBs/sec
Received 297795584 of 425392991 (70.0%), 141.9 MBs/sec
Received 425392991 of 425392991 (100.0%), 142.1 MBs/sec
Cache Size: ~406 MB (425392991 B)
/usr/bin/tar -xf /home/runner/work/_temp/b7d97389-e032-463d-b769-fd551f0999d0/cache.tzst -P -C /home/runner/work/microcks/microcks --use-compress-program unzstd
Cache restored successfully
Cache restored from key: maven-repo-147564147[20](https://github.com/microcks/microcks/actions/runs/14791356227/job/41529021782#step:5:21)

------------------------

Post job cleanup.
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/microcks/microcks --files-from manifest.txt --use-compress-program zstdmt
Sent [2](https://github.com/microcks/microcks/actions/runs/14791356227/job/41529021782#step:8:2)2734334 of 425387518 (5.3%), 21.7 MBs/sec
Sent 224060926 of 425387518 (52.7%), 106.8 MBs/sec
Sent 425387518 of 425387518 (100.0%), 160.3 MBs/sec
Cache saved with key: maven-repo-14791[3](https://github.com/microcks/microcks/actions/runs/14791356227/job/41529021782#step:8:3)56227
Cache saved with key: maven-repo-14791356227
```

When `package-webapp` starts the Maven cache is restored using `maven-repo-${{ github.run_id }}` and then the compilation is launched using `mvn -B -Pprod -DskipTests package`. `package` should be enough at that level because project dependencies should be pulled from the restore local repo...

We see in the pre compilation step logs:
```sh
Cache hit for: maven-repo-14791356227
Received 0 of 425387518 (0.0%), 0.0 MBs/sec
Received 12[16](https://github.com/microcks/microcks/actions/runs/14791356227/job/41529104295#step:4:17)34816 of 425387518 (28.6%), 58.0 MBs/sec
Received 243269632 of 425387518 (57.2%), 77.3 MBs/sec
Received 348127232 of 4253875[18](https://github.com/microcks/microcks/actions/runs/14791356227/job/41529104295#step:4:19) (81.8%), 82.9 MBs/sec
Received 425387518 of 425387518 (100.0%), 84.7 MBs/sec
Cache Size: ~406 MB (425387518 B)
```

### How to Reproduce?

_No response_

_No response_

_No response_

### Additional information

_No response_

Heyy-Himanshuu:
### Summary

Closes #1594.

- Drop `cache: maven` from every `actions/setup-java` invocation in `build-verify.yml` and `build-container.yml`. The explicit `actions/cache@v5` block (keyed on `github.run_id`) becomes the single source of truth for `~/.m2/repository`.
- Remove the per-job `mvn -B -DskipTests install` workaround from `package-webapp`, `integration-tests`, `fuzzing-tests`, and `javadoc-check`. It was masking the symptom; with the dual cache gone, the cached SNAPSHOTs from `build-install` now restore cleanly.

## Why

Two caches were active for `~/.m2/repository`:

1. `actions/setup-java` with `cache: maven` (keyed on `**/pom.xml` hashes)
2. An explicit `actions/cache@v5` block (keyed on `github.run_id`)

`setup-java` runs first and restores its pom-hash-keyed cache — likely a stale repo from an earlier unrelated run since pom hashes rarely change. The explicit `actions/cache` step then layers on top, but the resulting state can miss the SNAPSHOT artifacts `build-install` just installed. That's why downstream jobs needed the workaround.

GitHub's own [setup-java caveat](https://github.com/actions/setup-java#caching-packages-dependencies) calls this anti-pattern out: when you manage `~/.m2/repository` with an explicit `actions/cache`, don't also set `cache: maven` on setup-java.

## Scope

- `build-verify.yml`: 5× `cache: maven` removed, 4× workaround `mvn install` removed.
- `build-container.yml`: 1× `cache: maven` removed. Its `mvn -Pprod -DskipTests install` step is the actual job (producing the prod jar with the Angular UI for the container image), not a workaround — left untouched.
- `package-native.yml`: unchanged. It uses `cache: maven` without an explicit `actions/cache` block, so it isn't exhibiting the dual-cache conflict.

## Test plan

- [x] Branch CI on fork — all 5 jobs (`build-install`, `package-webapp`, `integration-tests`, `fuzzing-tests`, `javadoc-check`) green without the workaround.
- [ ] CI on this PR after open.

### DIFF
--- a/.github/workflows/build-verify.yml
+++ b/.github/workflows/build-verify.yml
@@ -32,9 +32,6 @@ jobs:
           architecture: x64
           cache: maven
 
-      - name: Build Java components
-        run: mvn -B clean install
-
       - name: Cache Maven dependencies
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
@@ -42,6 +39,9 @@ jobs:
           key: maven-repo-${{ github.run_id }}
           restore-keys: |
             maven-repo-
+
+      - name: Build Java components
+        run: mvn -B clean install
 
   package-webapp:
     runs-on: ubuntu-latest
@@ -70,7 +70,6 @@ jobs:
 
       - name: Build Angular app
         run: |
-          mvn -B -DskipTests install
           cd ${{github.workspace}}/webapp
           mvn -B -Pprod -DskipTests install
           
@@ -129,7 +128,6 @@ jobs:
 
       - name: Run Integration tests
         run: |
-          mvn -B -DskipTests install
           cd ${{github.workspace}}/webapp
           mvn -Pit test
           cd ${{github.workspace}}/minions/async
@@ -162,7 +160,6 @@ jobs:
 
       - name: Run Fuzzing tests
         run: |
-          mvn -B -DskipTests install
           cd ${{github.workspace}}/webapp
           mvn -Pfuzz test
 
@@ -193,7 +190,6 @@ jobs:
 
       - name: Verify Javadoc completion
         run: |
-          mvn -B -DskipTests install
           mvn -B javadoc:javadoc
 
   container-images-build:


Heyy-Himanshuu:
CI is green on my fork : https://github.com/Git-HimanshuRathi/microcks/actions/runs/25726875878

lbroudoux:
> CI is green on my fork : https://github.com/Git-HimanshuRathi/microcks/actions/runs/25726875878

Yes, but you didn't introduce new classes in - let's say - `util` module and then use them in webapp. Do you?

The thing is don't get here is the one you explained above:

> The explicit actions/cache step then layers on top, but the resulting state can miss the SNAPSHOT artifacts build-install just installed.

Why? If snapshots have been `installed`, then they must be present in  `~/.m2/repository` and thus can be cached, restored, and made available for the downstream jobs...

lbroudoux:
From my understanding, it should be possible to manage 2 different caches:
* 1 global, shared between GHA runs - until pom.xml change - that contains all the pulled dependencies
* 1 local, shared between jobs within the same GHA run that contains the local`~/.m2/repository` folder initialized from a global cache restoration + the SNAPSHOTS layered on top...

That's what I was trying to achieve previously.

If we remove the cache from the JDK setup step, I'm afraid we'll have to fetch all the project dependencies on each and every run, and this will take longer than reinstalling with tests skipped. Don't you think?

Heyy-Himanshuu:
@lbroudoux You were right original diagnosis was wrong. Pushed an experimental probe class in `commons/util` referenced from `webapp` to actually reproduce the bug, then iterated:

- **Pass A** (both caches + probe): red. Bug reproduced, but `cache: maven` is present same as before.
- **Pass B** (drop `cache: maven`, keep probe): still red. Refutes the dual-cache theory.
- **Pass C** (keep `cache: maven`, move `actions/cache` **before** `mvn install` in `build-install`): green.

Real root cause: in `build-install`, the `actions/cache` step was the **last** step. `actions/cache` restores at the position it's encountered, not at job start. With the step after `mvn install`, the primary key `maven-repo-${run_id}` didn't exist yet, so `restore-keys` fell back to a *prior* run's cache and extracted stale SNAPSHOTs on top of the freshly installed ones. Job-end save then captured the stale state exactly your log:

```
Cache restored from key: maven-repo-147564147   ← prior run, not the current one
```

The workaround was forcing a fresh `mvn install` to overwrite the stale cache in each downstream job.

Fix: move the cache step before the build in `build-install` (matching every downstream job's order). `cache: maven` stays global Maven Central cache preserved. Workaround removed from all 4 downstream jobs. Squashed to one clean commit (`8d04bbfd`), force-pushed. Probe deleted. Net diff is 10 lines in one file. Sorry for the noise.

Heyy-Himanshuu:
@lbroudoux anything wrong with this PR ?

lbroudoux:
> @lbroudoux anything wrong with this PR ?

Nothing wrong, but not time to review this in details for this week. Thanks for your patience!

.github/workflows/build-verify.yml
```diff
@@ -70,7 +70,6 @@ jobs:
 
       - name: Build Angular app
         run: |
-          mvn -B -DskipTests install
           cd ${{github.workspace}}/webapp
           mvn -B -Pprod -DskipTests install
```
lbroudoux, line 74: As the full install in prod mode is executed here, and if the result is correctly stored in the cache, couldn't we also remove this block here: https://github.com/microcks/microcks/blob/master/.github/workflows/build-container.yml#L44 ?

lbroudoux (commented):
I am wondering if we couldn't also remove the `install` done in build-containers? Can you check this?

.github/workflows/build-verify.yml
Heyy-Himanshuu: The `Package webapp in prod mode` step in `build-container.yml` can't be removed. Unlike the workarounds we eliminated from `build-verify.yml`, this step serves two genuine purposes:

1. **Produces `webapp/target/`** for the Docker build context (`context: ${{github.workspace}}/webapp`). Each job runs on a fresh runner — the Maven cache restores `~/.m2/repository` but not workspace `target/` directories. Without this step, the Dockerfile has nothing to package.
2. **Installs the prod webapp JAR** to `~/.m2/repository` so `distro/uber` can resolve it as a dependency when it runs `mvn -Pprod package` later in the same job.

What _does_ improve is the speed: with the warm cache (all deps + SNAPSHOT JARs already in `~/.m2/repository`), this install is much faster than before — no network calls, just a compile + Angular build.

lbroudoux: Makes a lot of sense. Thanks for these explanations. Ready to merge!

PR #2137, opened 2026-05-27
fix: OAS 3.0 exclusiveMinimum/Maximum parsing exception

### LINKED ISSUE #2129: OAS 3.0 exclusiveMinimum/Maximum parsing exception
### Describe the bug

When Microcks parses an OAS 3.0 specification with `exclusiveMinimum` and/or `exclusiveMaximum` set as `booleans` (as expected for OAS 3.0), it throws a `JsonSchemaException` exception.

### Expected behavior

Microcks should be able to parse the OAS 3.0 specification.

### Actual behavior

Microcks throws an exception:

```
08:29:12.808 ERROR 1 --- [    task-3] i.g.m.service.TestRunnerService          : Throwable while testing operation GET /v0/locations

com.networknt.schema.JsonSchemaException: exclusiveMinimum value is not a number
        at com.networknt.schema.ExclusiveMinimumValidator.<init>(ExclusiveMinimumValidator.java:45)
        at com.networknt.schema.ValidatorTypeCode.newValidator(ValidatorTypeCode.java:162)
        at com.networknt.schema.JsonMetaSchema.newValidator(JsonMetaSchema.java:496)
        at com.networknt.schema.ValidationContext.newValidator(ValidationContext.java:63)
        at com.networknt.schema.JsonSchema.read(JsonSchema.java:542)
        at com.networknt.schema.JsonSchema.getValidators(JsonSchema.java:1316)
        at com.networknt.schema.JsonSchema.<init>(JsonSchema.java:151)
```

### How to Reproduce?

1. Start Microcks with an OpenAPI specification using exclusiveMinimum set to `true/false`
2. Trigger the `OPEN API SCHEMA` tests
3. Microcks throws a `com.networknt.schema.JsonSchemaException: exclusiveMinimum value is not a number`

1.14.0

Docker / Quarkus Extension

### Additional information

It may be a downstream issue with the `json-schema-validator` library?

vleon1a:
### Description

When Microcks parses an OAS 3.0 specification with `exclusiveMinimum` and/or `exclusiveMaximum` set as `booleans` (as expected for OAS 3.0), it throws a `JsonSchemaException`.

The fix is the following:
- Update `JsonSchemaValidator` to allow providing a schema version, keeping `V202012` as default
- Update `OpenAPISchemaValidator` to extract the `openapi` field, and force the schema `V4` if it starts with `3.0.`

Fix tested locally using compose, adding my application to it, and triggering the `OPEN API SCHEMA` conformance test.
Attaching the logs before/after the fix:
- [microcks-logs-before-fix.txt](https://github.com/user-attachments/files/28308420/microcks-logs-before-fix.txt)
- [microcks-logs-after-fix.txt](https://github.com/user-attachments/files/28308419/microcks-logs-after-fix.txt)

Small bonus: The project requires Java 25, but the `BUILDING.md` was mentioning 21

### Related issue(s)

Fixes #2129

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -17,7 +17,7 @@ Don't forget to indicate your Java, Maven and/or Docker version.
 
 ### Build the whole project
 
-You need to have [Apache Maven](https://maven.apache.org) (version >= 3.5) up and running as well as a valid Java Development Kit (version >= 21) install to build the project.
+You need to have [Apache Maven](https://maven.apache.org) (version >= 3.5) up and running as well as a valid Java Development Kit (version >= 25) install to build the project.
 
 ```
 $ git clone https://github.com/microcks/microcks.git
--- a/commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
+++ b/commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
@@ -151,7 +151,7 @@ public class JsonSchemaValidator {
     * @return The list of validation failure messages. If empty, json object is valid !
     */
    public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode, String namespace) {
-      return validateJson(schemaNode, jsonNode, namespace, SpecVersion.VersionFlag.V202012);
+      return validateJson(schemaNode, jsonNode, namespace, JsonSchemaDialect.DRAFT_2020_12);
    }
 
    /**
@@ -160,17 +160,17 @@ public class JsonSchemaValidator {
     * semantics regarding additional or unknown attributes: schema must explicitly set <code>additionalProperties</code>
     * to false if you want to consider unknown attributes as validation errors. It returns a list of validation error
     * messages.
-    * @param schemaNode  The Json schema specification as a Jackson node
-    * @param jsonNode    The Json object as a Jackson node
-    * @param namespace   Namespace definition to resolve relative dependencies in Json schema
-    * @param versionFlag The Json schema draft version to use for validation (e.g. V4 for OAS 3.0, V202012 for OAS 3.1)
+    * @param schemaNode The Json schema specification as a Jackson node
+    * @param jsonNode   The Json object as a Jackson node
+    * @param namespace  Namespace definition to resolve relative dependencies in Json schema
+    * @param dialect    The JSON Schema dialect to validate against (DRAFT_4 for OAS 3.0, DRAFT_2020_12 for OAS 3.1)
     * @return The list of validation failure messages. If empty, json object is valid !
     */
    public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode, String namespace,
-         SpecVersion.VersionFlag versionFlag) {
+         JsonSchemaDialect dialect) {
       List<String> errors = new ArrayList<>();
 
-      final JsonSchema jsonSchemaNode = extractJsonSchemaNode(schemaNode, namespace, versionFlag);
+      final JsonSchema jsonSchemaNode = extractJsonSchemaNode(schemaNode, namespace, toVersionFlag(dialect));
 
       Set<ValidationMessage> messages = jsonSchemaNode.validate(jsonNode, executionContext -> {
          executionContext.getExecutionConfig().setFormatAssertionsEnabled(true);
@@ -216,6 +216,13 @@ public class JsonSchemaValidator {
          case V7 -> JsonMetaSchema.getV7();
          case V201909 -> JsonMetaSchema.getV201909();
          case V202012 -> JsonMetaSchema.getV202012();
+      };
+   }
+
+   private static SpecVersion.VersionFlag toVersionFlag(JsonSchemaDialect dialect) {
+      return switch (dialect) {
+         case DRAFT_4 -> SpecVersion.VersionFlag.V4;
+         case DRAFT_2020_12 -> SpecVersion.VersionFlag.V202012;
       };
    }
 }
--- a/commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
+++ b/commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
@@ -37,7 +37,7 @@ import java.util.Set;
 
 /**
  * Helper class for validating Json objects against their Json schema. Supported version of Json schema is
- * http://json-schema.org/draft-07/schema.
+ * http://json-schema.org/draft/2020-12/schema.
  * @author laurent
  */
 public class JsonSchemaValidator {
@@ -45,8 +45,6 @@ public class JsonSchemaValidator {
    /** A commons logger for diagnostic messages. */
    private static final Logger log = LoggerFactory.getLogger(JsonSchemaValidator.class);
 
-   public static final String JSON_V4_SCHEMA_IDENTIFIER = "http://json-schema.org/draft-04/schema#";
-   public static final String JSON_V7_SCHEMA_IDENTIFIER = "http://json-schema.org/draft-07/schema#";
    public static final String JSON_V12_SCHEMA_IDENTIFIER = "http://json-schema.org/draft/2020-12/schema#";
    public static final String JSON_SCHEMA_IDENTIFIER_ELEMENT = "$schema";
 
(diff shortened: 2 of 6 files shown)


SebastienDegodez (approved):
LGTM

vleon1a:
> LGTM

Thanks @SebastienDegodez !
Is it possible to include it in the 1.15.0 milestone?

SebastienDegodez:
> > LGTM
> 
> Thanks @SebastienDegodez ! Is it possible to include it in the 1.15.0 milestone?

Yes !
Another review @lbroudoux ?

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
```diff
@@ -145,8 +146,25 @@ public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode)
...
+    * <code>additionalProperties</code> to false if you want to consider unknown attributes as validation errors. It
+    * returns a list of validation error messages.
+    * @param schemaNode  The OpenAPI schema specification as a Jackson node
+    * @param jsonNode    The Json object as a Jackson node
+    * @param namespace   Namespace definition to resolve relative dependencies in Json schema
+    * @param versionFlag The Json schema draft version to validate against (V4 for OAS 3.0, V202012 for OAS 3.1)
+    * @return The list of validation failures. If empty, json object is valid !
+    */
+   public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode, String namespace,
+         SpecVersion.VersionFlag versionFlag) {
```
lbroudoux: I would prefer `SpecVersion.VersionFlag` not appear as a public method parameter as it's an implementation detail. We happened to change the validator implementation in the past and may also change it in the future, so being sticky to an implementation is not that great.

I also wonder why this has to be visible at the OpenAPISchemaValidator level? Should we have the valdator analyse the openapi file and choose the correct version of Json Schema?

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
```diff
@@ -213,7 +231,23 @@ public static List<String> validateJsonMessage(JsonNode specificationNode, JsonN
...
+    * OpenAPI 3.0.x relies on a schema dialect derived from JSON Schema draft-04 (boolean
+    * <code>exclusiveMinimum</code>/<code>exclusiveMaximum</code>) — V4 is the only networknt
+    * <code>SpecVersion.VersionFlag</code> that accepts that form. For OpenAPI 3.1.x and any other / missing version, we
+    * keep the default V202012.
+    */
+   private static SpecVersion.VersionFlag resolveSpecVersion(JsonNode specificationNode) {
+      String openApiVersion = specificationNode.path("openapi").asText("");
+      if (openApiVersion.startsWith("3.0.")) {
+         log.info("Detected OpenAPI specification version {}, parsing it using schema V4", openApiVersion);
+         return SpecVersion.VersionFlag.V4;
```
lbroudoux: I'm a bit concerned here because OpenAPI 3.0.x officially supports [JSON Schema Specification Wright Draft 00](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-00) (aka Json Schema Draft 05) but not `V4`. It seems that  `SpecVersion.VersionFlag` don't have a V5...

So I wonder what are the actual diffs between V5 and V4 and could we be confident sticking with V4?

commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
```diff
@@ -153,9 +151,26 @@ public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode)
...
+    * to false if you want to consider unknown attributes as validation errors. It returns a list of validation error
+    * messages.
+    * @param schemaNode  The Json schema specification as a Jackson node
+    * @param jsonNode    The Json object as a Jackson node
+    * @param namespace   Namespace definition to resolve relative dependencies in Json schema
+    * @param versionFlag The Json schema draft version to use for validation (e.g. V4 for OAS 3.0, V202012 for OAS 3.1)
+    * @return The list of validation failure messages. If empty, json object is valid !
+    */
+   public static List<String> validateJson(JsonNode schemaNode, JsonNode jsonNode, String namespace,
+         SpecVersion.VersionFlag versionFlag) {
```
lbroudoux: Got the same concern here about making `SpecVersion.VersionFlag` visible as it's an implementation details that may switch in the future... Maybe we can reintroduce the `JSON_V4_SCHEMA_IDENTIFIER` and others above as an enumerated type and do the conversion to `SpecVersion.VersionFlag` internally?

I think it would safer for the future.

BUILDING.md
```diff
@@ -17,7 +17,7 @@ Don't forget to indicate your Java, Maven and/or Docker version.
 
 ### Build the whole project
 
-You need to have [Apache Maven](https://maven.apache.org) (version >= 3.5) up and running as well as a valid Java Development Kit (version >= 21) install to build the project.
+You need to have [Apache Maven](https://maven.apache.org) (version >= 3.5) up and running as well as a valid Java Development Kit (version >= 25) install to build the project.
```
lbroudoux, line 20: Nice catch! Thanks!

lbroudoux (changes requested):
Thanks for pushing this! I have 2 concerns:
* One regarding the visibility of `SpecVersion.VersionFlag` => I would prefer to hide the underlying implementation,
* One regarding the actual schema version used for OpenAPI 3.0.x => the spes says Draft-05 but it's not supported by the validation library so you proposed fallabcking to Draft-04. Can the differences be neglected?

lbroudoux:
> Is it possible to include it in the 1.15.0 milestone?

Yes, I'd like to!

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
```diff
@@ -213,7 +231,23 @@ public static List<String> validateJsonMessage(JsonNode specificationNode, JsonN
...
+    * Resolve the JSON schema draft version to use when validating a message against the given OpenAPI specification.
+    * OpenAPI 3.0.x relies on a schema dialect derived from JSON Schema draft-04 (boolean
+    * <code>exclusiveMinimum</code>/<code>exclusiveMaximum</code>) — V4 is the only networknt
+    * <code>SpecVersion.VersionFlag</code> that accepts that form. For OpenAPI 3.1.x and any other / missing version, we
+    * keep the default V202012.
+    */
+   private static SpecVersion.VersionFlag resolveSpecVersion(JsonNode specificationNode) {
+      String openApiVersion = specificationNode.path("openapi").asText("");
+      if (openApiVersion.startsWith("3.0.")) {
+         log.info("Detected OpenAPI specification version {}, parsing it using schema V4", openApiVersion);
```
lbroudoux: This log can be lowered to debug imho

vleon1a:
> Thanks for pushing this! I have 2 concerns:
> * One regarding the visibility of `SpecVersion.VersionFlag` => I would prefer to hide the underlying implementation,
> * One regarding the actual schema version used for OpenAPI 3.0.x => the spes says Draft-05 but it's not supported by the validation library so you proposed fallabcking to Draft-04. Can the differences be neglected?

Thanks for reviewing! 
Let me see how I can rework the visibility aspect, it makes sense
As for the draft version, I did not see any any issue but I can't claim that this is exhaustive, I'll try to do some research on that, but for sure the V4 is the only one working from the library perspective

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
vleon1a: I pushed some changes in https://github.com/microcks/microcks/commit/0abf8d431af89f61a8b2a54fd75e1d08fb6e64d9

> I also wonder why this has to be visible at the OpenAPISchemaValidator level?

My idea was that it's up to the specific validator to deduce which schema should be used, as it depends on the specification itself (it may be different for AsyncAPI or GraphQL for instance).
So the JsonSchemaValidator remains agnostic

commons/util/src/main/java/io/github/microcks/util/JsonSchemaValidator.java
vleon1a: I pushed some changes in https://github.com/microcks/microcks/commit/0abf8d431af89f61a8b2a54fd75e1d08fb6e64d9

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
vleon1a: That's a bit verbose indeed, included in https://github.com/microcks/microcks/commit/0abf8d431af89f61a8b2a54fd75e1d08fb6e64d9

commons/util/src/main/java/io/github/microcks/util/openapi/OpenAPISchemaValidator.java
vleon1a: Still digging, so far I found that the library maintainers did a fix for my initial issue explicitly on V4: https://github.com/networknt/json-schema-validator/pull/1127
I also found in the OpenAPI specification repository some mention that they use a subset of V4

Another option may be to change my implementation: currently, only the latest specification is used, we could imagine having a fallback mechanism? If the validation fails we fallback to older versions down to V4?

lbroudoux: Making AI work on this, I get the feeling that there's no syntax or functional difference between Draft 4 (draft-zyp-json-schema-04) and Draft-Wright-00 (Draft 5). Draft 5 was published only as an editorial revision. Though OpenAPI 3.0 still implements only a subset of Draft 5, so we absolutely need to keep our conversion method in `OpenAPISchemaValidator`.

lbroudoux (approved):
It looks all good to me. Thanks!

PR #2141, opened 2026-05-28
fix: return proper HTTP status codes for import failures

### LINKED ISSUE #2135: bug: ImportController always returns HTTP 201 even on failure or empty/invalid file
### Describe the bug

The POST /api/import endpoint in ImportController.java always returns HTTP 201 (Created) 
regardless of what actually happened. If the uploaded file is empty, has the wrong content type, or the import throws an exception 
the API still returns 201 with no error message, making it impossible for clients to detect failures.

### Expected behavior

1) Empty file → 400 Bad Request
2) Wrong content type (not application/json) → 415 Unsupported Media Type
3) Import throws an exception → 500 Internal Server Error
4) Successful import → 201 Created (unchanged)

### Actual behavior

The endpoint always returns HTTP 201 Created in all cases including failures. 
Exceptions are also silently swallowed .  only e.getMessage() is logged,  losing the full stack trace.

### How to Reproduce?

1. Send a POST request to /api/import with an empty file
2. Send a POST request to /api/import with a file that has content type text/plain
3. Send a POST request to /api/import with a malformed JSON file

In all three cases the response is HTTP 201 with no error body.

Code reference:
webapp/src/main/java/io/github/microcks/web/ImportController.java line 65

docker-compose

### Additional information

Line 58 also uses new String(bytes) without specifying UTF-8 charset explicitly,  which is platform-dependent and can cause encoding issues on non-UTF-8 systems.
Suggested fix: return ResponseEntity with appropriate HttpStatus based on each  condition, and re-throw or return the exception details in the response body.

dipak0000812:
## Description

This changes the `/api/import` endpoint so it no longer returns `201 Created` for every request, even when the import fails.

Before this change:

* empty uploads still returned `201`
* unsupported content types still returned `201`
* exceptions during import were swallowed and the API still responded with success

With this update:

* empty files now return `400 Bad Request`
* unsupported content types return `415 Unsupported Media Type`
* import exceptions return `500 Internal Server Error`
* successful imports still return `201 Created`

I also updated the controller to use `StandardCharsets.UTF_8` when converting uploaded bytes to String.

## Tests

Added integration tests covering:

* empty file upload
* invalid content type
* import failure path returning `500`

Test command used:

```bash
mvn -pl webapp -am -Dtest=ImportControllerIT -Dsurefire.failIfNoSpecifiedTests=false test
```

Fixes #2135

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/ImportController.java
+++ b/webapp/src/main/java/io/github/microcks/web/ImportController.java
@@ -26,6 +26,8 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.multipart.MultipartFile;
 
+import java.nio.charset.StandardCharsets;
+
 /**
  * A Controller for importing new definitions into microcks repository.
  * @author laurent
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
Looks all good to me. Thanks for pushing it.

PR #2147, opened 2026-06-05
fix(api): redact secret search results

### LINKED ISSUE #2146: GET /api/secrets/search returns unredacted credentials to non-admin users
### Describe the bug

`GET /api/secrets/search` returns full `Secret` objects without applying the sensitive-data redaction used by the other secret read endpoints.

For non-admin users, `GET /api/secrets` and `GET /api/secrets/{id}` clear sensitive fields before returning secrets. The search endpoint currently skips that authorization check and returns the repository result directly, which can expose fields such as `username`, `password`, `token`, and `caCertPem` to any authenticated user allowed to call the endpoint.

### Expected behavior

`searchSecrets()` should behave consistently with `listSecrets()` and `getSecret()`:

- admins can receive the complete secret data
- non-admin users should receive secrets with sensitive fields redacted

### Current behavior

In `SecretController.java`, `listSecrets()` applies `filterSensitiveData()` when the caller is not an admin:

```java
if (!authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)) {
   return secretRepository.findAll(...)
         .stream().map(this::filterSensitiveData).toList();
}
```

`getSecret()` follows the same pattern.

However, `searchSecrets()` does not accept `UserInfo`, does not check the caller role, and returns raw repository results:

```java
@GetMapping(value = "/secrets/search")
public List<Secret> searchSecrets(@RequestParam(value = "name") String name) {
   log.debug("Searching secrets corresponding to {}", name);
   return secretRepository.findByNameLike(name);
}
```

### How to reproduce

1. Deploy Microcks with authentication enabled.
2. Create a secret containing credential data.
3. Authenticate as a user with the `user` role, but without the `admin` role.
4. Call:

   ```text
   GET /api/secrets/search?name=<secret-name-or-pattern>
   ```

5. Observe that the response includes sensitive values from the matching `Secret`.
6. Compare with:

   ```text
   GET /api/secrets?page=0&size=20
   ```

   The list endpoint correctly returns the same secret with sensitive fields redacted for the same non-admin caller.

### Suggested fix

Add `UserInfo` to `searchSecrets()` and apply the same redaction logic used by the other read endpoints:

```java
@GetMapping(value = "/secrets/search")
public List<Secret> searchSecrets(@RequestParam(value = "name") String name, UserInfo userInfo) {
   log.debug("Searching secrets corresponding to {}", name);
   List<Secret> secrets = secretRepository.findByNameLike(name);
   if (!authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)) {
      return secrets.stream().map(this::filterSensitiveData).toList();
   }
   return secrets;
}
```

### Additional context

`SecurityConfiguration` already gates `GET /api/secrets/*` behind authenticated roles, so this is not unauthenticated access. The issue is that users who are allowed to discover secret names can also receive credential values through the search endpoint.

Verified against the current `master` / `1.15.0-SNAPSHOT` code path.

gyanranjanpanda:
## Summary

This updates `GET /api/secrets/search` so it applies the same sensitive-data redaction already used by the other secret read endpoints.

Non-admin users can still search and discover matching secrets, but the response no longer includes credential values such as `username`, `password`, `token`, or `caCertPem`. Admin users continue to receive complete secret data.

## Root cause

`listSecrets()` and `getSecret()` both check whether the caller has the admin role before returning secret data. `searchSecrets()` did not receive `UserInfo`, skipped that role check, and returned raw repository results directly.

## Validation

Added `SecretControllerTest` coverage for both non-admin and admin search responses.

```text
mvn -pl webapp -am -Dtest=SecretControllerTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Fixes #2146

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/SecretController.java
+++ b/webapp/src/main/java/io/github/microcks/web/SecretController.java
@@ -77,9 +77,14 @@ public class SecretController {
    }
 
    @GetMapping(value = "/secrets/search")
-   public List<Secret> searchSecrets(@RequestParam(value = "name") String name) {
+   public List<Secret> searchSecrets(@RequestParam(value = "name") String name, UserInfo userInfo) {
       log.debug("Searching secrets corresponding to {}", name);
-      return secretRepository.findByNameLike(name);
+      List<Secret> secrets = secretRepository.findByNameLike(name);
+      if (!authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)) {
+         return secrets.stream().map(this::filterSensitiveData).toList();
+      }
+      // We're admin, so we can return all secrets with sensitive data.
+      return secrets;
    }
 
    @GetMapping(value = "/secrets/count")
--- a/webapp/src/test/java/io/github/microcks/web/SecretControllerTest.java
+++ b/webapp/src/test/java/io/github/microcks/web/SecretControllerTest.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.microcks.web;
+
+import io.github.microcks.domain.Secret;
+import io.github.microcks.repository.SecretRepository;
+import io.github.microcks.security.AuthorizationChecker;
+import io.github.microcks.security.UserInfo;
+
+import org.assertj.core.api.SoftAssertions;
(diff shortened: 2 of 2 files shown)


minions/async/src/main/java/io/github/microcks/minion/async/consumer/WebSocketClient.java
lbroudoux, line 44: Unrelated change => please rebase your branch

webapp/src/main/java/io/github/microcks/util/ai/OpenAICopilot.java
lbroudoux, line 349: Unrelated change => please rebase your branch

webapp/src/main/java/io/github/microcks/util/metadata/ExamplesExporter.java
lbroudoux, line 174: Unrelated change => please rebase your branch

lbroudoux (changes requested):
Changes related to `SecretControlle` are ok but there are other changes because your branch is probably not up-to-date with `master`. Please rebase your branch before resubmitting this PR.

gyanranjanpanda:
check this now @lbroudoux

lbroudoux:
> check this now @lbroudoux

All good! Thanks for quick reply.

lbroudoux (approved):
LGTM. Thanks!

PR #2150, opened 2026-06-06
feat(async): propagate correlation ID header in Kafka request-reply handler

### LINKED ISSUE #2149: Enhancement (TODO): Propagate Correlation ID Header in Kafka Request-Reply Handler
### Reason/Context

In `KafkaRequestReplyHandler.java`, the `buildReplyHeaders()` method contains an open `TODO`:
```
// TODO: Add correlation ID header when we parse correlation object from AsyncAPI spec
```
Currently, reply messages are sent without any correlation ID, making it impossible for consumers to match a reply back to its originating request, a fundamental requirement of async request-reply messaging.

### Description

When an incoming Kafka request carries a `correlationId` header (per the AsyncAPI correlation ID object), that value should be forwarded on the outgoing reply. If no `correlationId` header is present, a UUID should be auto-generated to ensure every exchange is traceable.

**File:** `minions/async/src/main/java/io/github/microcks/minion/async/handler/KafkaRequestReplyHandler.java`

### Implementation ideas

- Read `correlationId` from incoming request headers in `handleRequest()`.
- If absent, auto-generate a `UUID`.
- Pass it to `buildReplyHeaders()` and add it as a `RecordHeader` on the reply.
- Define a constant `CORRELATION_ID_HEADER = "correlationId"` for the header name.

Vaishnav88sk:
### Description

- Reads `correlationId` from incoming Kafka request headers.
- Falls back to a generated UUID if the header is absent.
- Forwards the correlation ID on the outgoing reply so consumers can match replies to requests.
- Removes the open `TODO` in `buildReplyHeaders()`.

### Related issue(s)
Fixes #2149

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/handler/KafkaRequestReplyHandler.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/handler/KafkaRequestReplyHandler.java
@@ -36,6 +36,7 @@ import org.eclipse.microprofile.config.Config;
 import org.jboss.logging.Logger;
 
 import java.nio.charset.StandardCharsets;
+import java.util.UUID;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


Vaishnav88sk:
Not stale - Active
@lbroudoux

PR #2156, opened 2026-06-08
fix: return 404 for missing import jobs

### LINKED ISSUE #2155: JobController.deleteJob throws NPE when deleting a non-existent job
### Describe the bug

Calling `DELETE /api/jobs/{id}` with an ID that does not exist returns a `500 Internal Server Error` instead of `404 Not Found`.

In `JobController.deleteJob`, the repository lookup can return `null`, but the result is passed directly to `authorizationChecker.hasRoleForImportJob(...)`. For a manager user, this eventually calls `job.getMetadata()` and throws a `NullPointerException`.

This does not happen for admin users because the admin role check short-circuits before `hasRoleForImportJob(...)` is called.

The related `activateJob` endpoint has a similar inconsistency: it avoids the NPE, but returns `403 Forbidden` when the job does not exist. `startJob` and `stopJob` already return `404 Not Found` in this case.

### Steps to reproduce

1. Enable Keycloak authentication with repository tenancy.
2. Authenticate as a manager user without the admin role.
3. Send `DELETE /api/jobs/nonexistent-id`.
4. Observe a `500 Internal Server Error` and an NPE in `AuthorizationChecker.hasRoleForImportJob`.

### Expected behavior

Both `deleteJob` and `activateJob` should return `404 Not Found` when the requested job does not exist, consistently with `startJob` and `stopJob`. Authorization should only be checked after confirming that the job was found.

### Environment

Microcks `master` at `e6b5a0d7`.

### Relevant code

- [`JobController.deleteJob`](https://github.com/microcks/microcks/blob/e6b5a0d7/webapp/src/main/java/io/github/microcks/web/JobController.java#L207-L217)
- [`AuthorizationChecker.hasRoleForImportJob`](https://github.com/microcks/microcks/blob/e6b5a0d7/webapp/src/main/java/io/github/microcks/security/AuthorizationChecker.java#L95-L96)

gyanranjanpanda:
## Summary

- return `404 Not Found` when `deleteJob` receives an unknown job ID
- return the same `404 Not Found` response from `activateJob`
- add regression tests for both endpoints

## Root cause

`deleteJob` passed the nullable repository result to `hasRoleForImportJob`. For manager users, that method dereferenced the missing job and caused a `NullPointerException`.

Both endpoints now check whether the job exists before performing authorization. This also makes their behavior consistent with `startJob` and `stopJob`.

## Validation

- `mvn -pl webapp -am -Dtest=JobControllerTest -Dsurefire.failIfNoSpecifiedTests=false test`
- Spotless checks run by the commit hook

Fixes #2155

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/JobController.java
+++ b/webapp/src/main/java/io/github/microcks/web/JobController.java
@@ -155,8 +155,13 @@ public class JobController {
    public ResponseEntity<ImportJob> activateJob(@PathVariable("id") String jobId, UserInfo userInfo) {
       log.debug("Activating job with id {}", jobId);
       ImportJob job = jobRepository.findById(jobId).orElse(null);
-      if (job != null && (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
-            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job))) {
+
+      if (job == null) {
+         return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+      }
+
+      if (authorizationChecker.hasRole(userInfo, AuthorizationChecker.ROLE_ADMIN)
+            || authorizationChecker.hasRoleForImportJob(userInfo, AuthorizationChecker.ROLE_MANAGER, job)) {
          job.setActive(true);
          initMetadataIfMissing(job);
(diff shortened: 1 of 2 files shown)


gyanranjanpanda:
could @lbroudoux  review this

lbroudoux (approved):
LGTM. Thank you!

PR #2158, opened 2026-06-09
fix: prevent StringIndexOutOfBoundsException in HARImporter for URLs without file extension

### LINKED ISSUE #2157: bug: HARImporter crashes with StringIndexOutOfBoundsException on URLs with no file extension
### Describe the bug

When importing a HAR file that contains REST API requests with URLs that have 
no file extension (e.g. /api/users or /api/v1/products), the HARImporter 
crashes with a StringIndexOutOfBoundsException.

webapp/src/main/java/io/github/microcks/util/har/HARImporter.java
Method: filterValidEntries()     line 387

String extension = url.substring(url.lastIndexOf("."));

If the URL contains no "." character, lastIndexOf(".") returns -1, 
and substring(-1) throws StringIndexOutOfBoundsException, causing 
the entire HAR import to fail.

### Expected behavior

URLs with no file extension (e.g. /api/users, /api/v1/orders) should be 
treated as valid API entries and pass through the filter without crashing. 
Only known static resource extensions like .ico .html .css .js .png 
should be filtered out.

### Actual behavior

StringIndexOutOfBoundsException is thrown when any HAR entry contains 
a URL with no "." character. The full import fails and no services are created.

### How to Reproduce?

1. Capture or create a HAR file with a request to a clean REST endpoint:
   GET http://localhost:8080/api/users

2. Import the HAR file into Microcks via Upload

3. Observe the crash  - StringIndexOutOfBoundsException in filterValidEntries()

docker-compose

### Additional information

Suggested fix in filterValidEntries():

  int dotIndex = url.lastIndexOf(".");
  if (dotIndex == -1) return true;
  String extension = url.substring(dotIndex);
  return !INVALID_ENTRY_EXTENSIONS.contains(extension);

This ensures extension-less API URLs are always treated as valid 
entries instead of crashing the importer.

Harishrs2006:
Summary
1) Fixed a crash in `HARImporter.filterValidEntries()` where URLs with no 
  file extension (e.g. `/api/users`) caused `lastIndexOf(".")` to return `-1`, 
  making `substring(-1)` throw `StringIndexOutOfBoundsException`
2)Added a guard so extension-less URLs are treated as valid API entries
3) HAR import no longer fails when any entry has a clean REST URL

Changes
`webapp/src/main/java/io/github/microcks/util/har/HARImporter.java` line 387

Test plan
[ ] Import a HAR file with a request to `/api/users` (no extension)   should succeed
[ ] Verify existing HAR files with extensions (.ico, .js) still get filtered out

Fixes #2157

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/har/HARImporter.java
+++ b/webapp/src/main/java/io/github/microcks/util/har/HARImporter.java
@@ -384,7 +384,11 @@ public class HARImporter implements MockRepositoryImporter {
    private List<JsonNode> filterValidEntries(Iterator<JsonNode> entries) {
       return Stream.generate(() -> null).takeWhile(x -> entries.hasNext()).map(next -> entries.next()).filter(entry -> {
          String url = entry.path(REQUEST_NODE).path("url").asText();
-         String extension = url.substring(url.lastIndexOf("."));
+         int dotIndex = url.lastIndexOf(".");
+         if (dotIndex == -1) {
+            return true;
+         }
+         String extension = url.substring(dotIndex);
          return !INVALID_ENTRY_EXTENSIONS.contains(extension);
       }).filter(entry -> {
          if (apiPrefix != null) {


lbroudoux:
Hi @Harishrs2006 
This looks good but the DCO check fails. Can you check the details and apply the commands to fix this?
I'll be ready to merge just after.

lbroudoux (approved):
LGTM. Thank you!

PR #2163, opened 2026-06-11
fix: add null check for GenericResource payload in ResourceController

### LINKED ISSUE #2162: bug: NullPointerException silently swallowed in ResourceController when GenericResource payload is null
### Describe the bug

In ResourceController.java line 123-124, getPayload() is called on a GenericResource 
without a null check. If getPayload() returns null, calling reference.append() 
throws a NullPointerException.

The exception is caught by a broad catch(Exception e) block that only logs 
a warning, silently swallowing the error. The API returns an incomplete 
response with no schema instead of a proper error, making this very hard 
to debug.

webapp/src/main/java/io/github/microcks/web/ResourceController.java
Lines: 123-128

### Expected behavior

If getPayload() returns null, the code should explicitly handle it with 
a null check and either skip schema building or return a meaningful error 
response   not silently fail.

### Actual behavior

NullPointerException is thrown at reference.append() and caught silently 
by a generic catch(Exception e) block. Only a warn log is emitted. 
The referenceSchema stays null and the response is incomplete 
with no indication of failure to the caller.

### How to Reproduce?

1. Create a GenericResource with a null payload in the database
2. Call GET /api/resources/{serviceId}/{resourceType} for that service
3. Observe only a WARN log   no error returned to client

Code reference:
webapp/src/main/java/io/github/microcks/web/ResourceController.java lines 123-128

Document reference = genericResources.get(0).getPayload(); // can be null
reference.append(ID_FIELD, genericResources.get(0).getId()); // NPE here

docker-compose

### Additional information

Suggested fix  : add a null check before calling .append():

   Document reference = genericResources.get(0).getPayload();
   if (reference != null) {
      reference.append(ID_FIELD, genericResources.get(0).getId());
      referenceSchema = OpenAPISchemaBuilder.buildTypeSchemaFromJson(reference.toJson());
   } else {
      log.warn("Reference payload is null for serviceId: {}", serviceId);
   }

Harishrs2006:
Summary
1) `getPayload()` can return `null` on a `GenericResource`, causing a silent 
  `NullPointerException` at `reference.append()` 
2) The exception was being swallowed by a generic `catch(Exception e)` block 
  with only a warn log, making it very hard to debug
3)Added an explicit `null` guard with a descriptive warning log

 Changes:
`webapp/src/main/java/io/github/microcks/web/ResourceController.java` lines 123–128

Test plan
1)[ ] Call `GET /api/resources/{serviceId}/resourceType` for a service 
  with a GenericResource that has a null payload
2)[ ] Verify no NullPointerException is thrown
3) [ ] Verify a clear warning log is emitted instead

Fixes #2162

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/ResourceController.java
+++ b/webapp/src/main/java/io/github/microcks/web/ResourceController.java
@@ -121,8 +121,12 @@ public class ResourceController {
          if (genericResources != null && !genericResources.isEmpty()) {
             try {
                Document reference = genericResources.get(0).getPayload();
-               reference.append(ID_FIELD, genericResources.get(0).getId());
-               referenceSchema = OpenAPISchemaBuilder.buildTypeSchemaFromJson(reference.toJson());
+               if (reference != null) {
+                  reference.append(ID_FIELD, genericResources.get(0).getId());
+                  referenceSchema = OpenAPISchemaBuilder.buildTypeSchemaFromJson(reference.toJson());
+               } else {
+                  log.warn("Reference payload is null for serviceId: {}", serviceId);
+               }
             } catch (Exception e) {
                log.warn("Exception while building reference schema", e);
             }


lbroudoux (approved):
LGTM. Thank you!

PR #2166, opened 2026-06-11
fix: run Keycloak token refresh outside Angular zone to avoid unnecessary change detection

### LINKED ISSUE #2124: Optimize Keycloak token refresh by running it outside Angular zone
Problem
The Keycloak auth service refreshes tokens on a fixed interval inside Angular's zone.
That timer can cause unnecessary change detection even though no UI state has changed.

Expected behavior
- Refresh token in the background without triggering Angular change detection.
- Keep auth state correct and reliable.
- Avoid impacting app responsiveness during idle refresh cycles.

Proposed fix
- Run the periodic `keycloak.updateToken(...)` call inside `NgZone.runOutsideAngular(...)`
- Only re-enter Angular when the UI or application state actually needs to update

Why it matters
This change reduces unnecessary frontend work and makes the auth refresh process less intrusive.

Harishrs2006:
Summary
- Wrapped the periodic `setInterval` token refresh in `NgZone.runOutsideAngular()`
- Prevents Angular from triggering change detection every 30 seconds when no UI state has changed
- Added `NgZone` as a constructor dependency in `KeycloakAuthenticationService`

Changes
`webapp/src/main/webapp/src/app/services/auth-keycloak.service.ts`

Test plan
- [ ] Login to Microcks with Keycloak auth enabled
- [ ] Verify token still refreshes correctly after 30 seconds
- [ ] Verify no unnecessary Angular change detection cycles triggered during idle time

Fixes #2124

### DIFF
--- a/webapp/src/main/webapp/src/app/services/auth-keycloak.service.ts
+++ b/webapp/src/main/webapp/src/app/services/auth-keycloak.service.ts
@@ -18,6 +18,7 @@ import { Observable, BehaviorSubject } from 'rxjs';
 import { User } from '../models/user.model';
 import { ConfigService } from './config.service';
 import { HttpClient } from '@angular/common/http';
+import { NgZone } from '@angular/core';
 
 /**
  * A version of the authentication service that uses keycloak.js to provide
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
It looks all good. Thanks for pushing this.

PR #2187, opened 2026-06-25
fix: preserve null data in mock error responses

### LINKED ISSUE #2176: GraphQL mock returns HTTP 500 "Exception while filtering response JSON" for error responses (data: null); no support for the GraphQL errors array
### Describe the bug

Microcks' GraphQL mocking only models the `data` field of a response (it projects `data` onto the consumer's selection set). It has no support for the GraphQL `errors` array, and when a stored response has `data: null` (the standard shape for a GraphQL error response), serving it fails with **HTTP 500 `Exception while filtering response JSON`** instead of returning the stored payload.

This is effectively two issues:

1. **Bug:** the field-filtering code throws when the `data` node is `null` (or absent), rather than degrading gracefully.
2. **Missing feature:** even when the response *can* be served (e.g. `data: { alert: null }`), the top-level `errors` array is dropped, so a faithful GraphQL error response cannot be mocked.

### How to Reproduce

1. Import a GraphQL schema, e.g.:
   ```graphql
   type Query { alert(id: String): Alert }
   type Alert { id: String }
   ```
2. Import an `APIExamples` artifact with an error example whose response body is a standard GraphQL error:
   ```json
   { "data": null, "errors": [ { "message": "Access Denied", "path": ["alert"], "extensions": { "classification": "DataFetchingException" } } ] }
   ```
3. Make the mock serve that example (e.g. a `SCRIPT` dispatcher returning its name, or query args that match it).
4. Query the mock: `{ alert(id: "x") { id } }`.

### Expected behavior

The mock returns the stored GraphQL error response (HTTP 200 with `data: null` and the `errors` array), as a real GraphQL server would.

### Actual behavior

- `{"data": null, "errors": [...]}` → **HTTP 500 `Exception while filtering response JSON`**.
- `{"errors": [...]}` (no `data` key) → HTTP 500.
- `{"data": {"alert": null}, "errors": [...]}` → HTTP 200, but returns only `{"data":{"alert":null}}` — the `errors` array is silently dropped.

### Microcks version

1.14.0

### Additional context

The [GraphQL conventions docs](https://microcks.io/documentation/references/artifacts/graphql-conventions/) only cover successful data responses; error responses aren't mentioned. Supporting the `errors` array in mock responses (and handling `data: null` without throwing) would let teams mock error scenarios for contract testing.

Nachiket-Roy:
### Description

- Handle GraphQL mock responses with null or missing data fields without failing.
- Preserve GraphQL error payloads when aggregating mock responses.
- Add integration tests for GraphQL error response scenarios.

### Related issue

Fixes #2176

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/GraphQLController.java
+++ b/webapp/src/main/java/io/github/microcks/web/GraphQLController.java
@@ -249,17 +249,50 @@ public class GraphQLController {
       // ensures that aliasing applies consistently for both multi and single queries, matching actual
       // GraphQL behavior.
       ObjectNode aggregated = mapper.createObjectNode();
-      ObjectNode dataNode = aggregated.putObject("data");
-      for (GraphQLQueryResponse response : graphqlResponses) {
-         dataNode.set(StringUtils.defaultIfBlank(response.getAlias(), response.getOperationName()),
-               response.getJsonResponse().path("data").path(response.getOperationName()).deepCopy());
+
+      boolean hasDataKey = true;
+      boolean isDataNull = false;
+
+      if (graphqlResponses.size() == 1) {
+         JsonNode responseJson = graphqlResponses.get(0).getJsonResponse();
+         if (responseJson == null) {
+            hasDataKey = false;
+         } else if (!responseJson.has("data")) {
+            hasDataKey = false;
+         } else if (responseJson.get("data").isNull()) {
+            isDataNull = true;
+         }
+      }
+
+      if (hasDataKey) {
+         if (isDataNull) {
+            aggregated.set("data", mapper.nullNode());
+         } else {
+            ObjectNode dataNode = aggregated.putObject("data");
+            for (GraphQLQueryResponse response : graphqlResponses) {
+               JsonNode responseJson = response.getJsonResponse();
+               String key = StringUtils.defaultIfBlank(response.getAlias(), response.getOperationName());
+               if (responseJson == null) {
+                  dataNode.set(key, mapper.nullNode());
+               } else {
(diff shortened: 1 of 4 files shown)


lbroudoux:
It looks like the build is broken with tests failing => https://github.com/microcks/microcks/actions/runs/28181410001/job/83479399948?pr=2187#step:5:2004
Would you mind looking at it? Also, remove the `Draft` status when ready.

Nachiket-Roy:
@lbroudoux Should be fixed now, PTAL

lbroudoux:
Sorry but we still have failing integration tests here: https://github.com/microcks/microcks/actions/runs/28241308985/job/83683523044#step:5:1979

Be sure to have them pass locally using `mvn clean -it test` in the `/webapp` folder.

Nachiket-Roy:
@lbroudoux I ran the `webapp` module tests locally and all 420 tests pass.
While looking into the build, I also noticed a ~30s hang during CI shutdown. It comes from the gRPC server's lifecycle: the shutdown hook calls `grpcServer.shutdown()` but never releases the `CountDownLatch` the await thread is parked on, so Surefire's exit-timeout ends up being the only thing that actually terminates it.
This is unrelated to this PR, but if it's worth fixing I'm happy to open a separate PR.

lbroudoux:
> Surefire's exit-timeout ends up being the only thing that actually terminates it.
This is unrelated to this PR, but if it's worth fixing I'm happy to open a separate PR.

Oh yes, that's a good catch. Please do submit another PR for this.

lbroudoux (approved):
It looks all good to me. Thank you @Nachiket-Roy !

PR #2191, opened 2026-06-26
fix: handle keyed AsyncAPI 3 examples

### LINKED ISSUE #2045: NPE in Async Minion with AsyncAPI v3 inline examples - EventMessage.getContent() is null
Body:
## Description

When importing an AsyncAPI v3 spec with inline `examples` in the message definition, the Async Minion crashes with a NullPointerException every 3 seconds. This also blocks the Sync-to-Async triggers feature (#1956) from working.

## Affected versions

- `nightly` (tested April 2026)
- Likely `1.14.0-rc1` and `1.14.0-rc2`

## Steps to reproduce

1. Start Microcks with `docker-compose-devmode.yml` using `nightly` images
2. Import an AsyncAPI v3 spec with inline examples:

```yaml
asyncapi: 3.0.0
info:
  title: My API
  version: 1.0.0
channels:
  myChannel:
    messages:
      MyMessage:
        $ref: '#/components/messages/MyMessage'
operations:
  sendMessage:
    action: send
    channel:
      $ref: '#/channels/myChannel'
    messages:
      - $ref: '#/channels/myChannel/messages/MyMessage'
components:
  messages:
    MyMessage:
      payload:
        type: object
      examples:
        - myExample:
            payload: |-
              {"id": "123", "name": "test"}
```

3. Observe Async Minion logs
Actual behavior
Two NPEs crash the Async Minion continuously:
1. Scheduled producer crash (every 3s):
java.lang.NullPointerException: Cannot invoke "String.contains(java.lang.CharSequence)" 
because the return value of "io.github.microcks.domain.EventMessage.getContent()" is null
  at io.github.microcks.minion.async.producer.ProducerManager.lambda$getPureEventMessages$3(ProducerManager.java:447)
2. Service update consumer crash:
Cannot invoke "io.github.microcks.domain.ServiceType.equals(Object)" 
because the return value of "io.github.microcks.domain.Service.getType()" is null
3. Trigger consumer crash (blocks #1956):
application-io.github.microcks.minion.async.AsyncMockProducerTrigger#onAsyncAPITriggerCommand: [KO]
Health check reports all channels DOWN:
```
{"status":"DOWN","checks":[{"name":"SmallRye Reactive Messaging - readiness check","status":"DOWN","data":{"microcks-services-updates":"[KO]","microcks-asyncapi-triggers":"[KO]"}}]}
```
Expected behavior
• EventMessage.content should be populated from the inline example payload during import
• Service type should be set to EVENT for AsyncAPI v3 specs
• Scheduled producer should skip messages with null content (or not store them with null)
• Sync-to-Async trigger should publish the event to the output Kafka topic

Root cause analysis
The AsyncAPI v3 importer appears to not properly map the v3 examples format (examples: [{name: {payload: ...}}]) to EventMessage.content. The content is stored as null in MongoDB, causing the NPE in ProducerManager.getPureEventMessages() at line 447.
Impact
•
All async mock production broken for AsyncAPI v3 specs with inline examples
•
Sync-to-Async triggers (#1956) completely blocked for AsyncAPI v3
•
Trigger fires correctly (message appears in microcks-asyncapi-triggers internal topic) but Async Minion cannot process it

Environment
•
macOS, Docker Desktop
•
Microcks nightly (quay.io/microcks/microcks:nightly)
•
Async Minion nightly (quay.io/microcks/microcks-async-minion:nightly)
•
AsyncAPI 3.0.0

arnabnandy7:
### Description

- Added support for keyed/map-style AsyncAPI 3 examples, such as `- myExample: { payload: ... }`, when importing event messages.
- Reused the keyed example name as the imported `EventMessage` name and unwrapped the nested example before reading payload and headers.
- Hardened async producer message filtering/rendering to avoid NPEs when an `EventMessage` has null content.
- Added regression tests for keyed AsyncAPI 3 examples and null-content event messages.

### Related issue(s)

Fixes #2045

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/ProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/ProducerManager.java
@@ -479,7 +479,7 @@ public class ProducerManager {
 
    /** Get the event messages that are not contextualized with request or response part. */
    private List<EventMessage> getPureEventMessages(AsyncMockDefinition definition) {
-      return definition.getEventMessages().stream()
+      return definition.getEventMessages().stream().filter(eventMessage -> eventMessage.getContent() != null)
             .filter(eventMessage -> !eventMessage.getContent().contains("request.")
                   && !eventMessage.getContent().contains("response."))
             .toList();
@@ -487,7 +487,7 @@ public class ProducerManager {
 
    /** Get the event messages that are contextualized with request or response part. */
    private List<EventMessage> getContextualizedMessages(AsyncMockDefinition definition) {
-      return definition.getEventMessages().stream()
+      return definition.getEventMessages().stream().filter(eventMessage -> eventMessage.getContent() != null)
             .filter(eventMessage -> eventMessage.getContent().contains("request.")
                   || eventMessage.getContent().contains("response."))
             .toList();
@@ -496,7 +496,7 @@ public class ProducerManager {
    /** Render event message content from definition applying template rendering if required. */
    private String renderEventMessageContent(EventMessage eventMessage) {
       String content = eventMessage.getContent();
-      if (content.contains(TemplateEngine.DEFAULT_EXPRESSION_PREFIX)) {
+      if (content != null && content.contains(TemplateEngine.DEFAULT_EXPRESSION_PREFIX)) {
          logger.debug("EventMessage contains dynamic EL expression, rendering it...");
          TemplateEngine engine = TemplateEngineFactory.getTemplateEngine();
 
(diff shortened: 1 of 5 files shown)


lbroudoux:
Thanks but ... keyed AsyncAPI 3 examples haven't to be supported as they are not correct regarding the spec! Please check https://www.asyncapi.com/docs/reference/specification/v3.1.0#messageExampleObject 

They were introduced only at the beginning of AsyncAPI 2.0 because it was not present in the spec but syntactically valid using the extension capabilities. Examples have been fully described for AsyncAPI 2.1 and are not supposed to be keyed elements.

PR #2200, opened 2026-06-30
fix: gRPC server shutdown hang

Nachiket-Roy:
### Description

- Added a finally block in the JVM shutdown hook to count down the CountDownLatch (latch), ensuring the grpc-server-awaiter thread is unblocked and completes.


lbroudoux (approved):
LGTM. Thank you @Nachiket-Roy 🙏

Issue #2201, opened 2026-07-01
Support nested fragment spreads selection in GraphQL


lbroudoux:
### Reason/Context

As of today, Microcks does not support nested fragment spread selection in GraphQL queries. This is a very common pattern that arises quickly when schemas get more complex. 

### Description

Consider the following query:
```graphql
query film($id: String) {
  film(id: "ZmlsbXM6MQ==") {
    ...FilmCoreFields
  }
}
fragment FilmCoreFields on Film {
  ...FilmIdentityFields
  episodeID
}
fragment FilmIdentityFields on Film {
  id
  title
}
```
Today's return is 
```json
{
   "data": {
     "film": {}
   }
 }
```
where it is expected to be:
```json
{
   "data": {
     "film": {
       "id": "ZmlsbXM6MQ==",
       "title": "A New Hope",
       "episodeID": 4
     }
   }
 }
```

### Implementation ideas

We've got this result because - even if we implemented recursion correctly - the nested `...FilmIdentityFields` fragment operates at the same level as `episodeID`. Because we're applying the fragment filter first, only the `id` and `title` are retained in the response. Later on, we try to retain `episodeID` but this one is absent and thus it empties the result.

We must then change this and operate with collecting and flatening all field selections (simple and fragment-based ones) of the same level before applying the retain on result.

lbroudoux:
Now available in `nightly` tagged container image.

PR #2203, opened 2026-07-01
fix: #2199 GraphQL 500 when fragment definitions precede the operation

### LINKED ISSUE #2199: GraphQL: named fragment spreads cause HTTP 500; inlining them exposes a second bug where Microcks synthesises null for unrecorded fields
## Summary

Two related bugs surface when a GraphQL client sends a query that uses **named fragment spreads** (`...FragmentName`) to a Microcks mock endpoint:

1. **Bug 1 — HTTP 500 on named fragment spreads.** Microcks returns a 500 Internal Server Error for any query that contains a `FragmentSpread` node. Replacing the spread with a semantically equivalent `InlineFragment` (`... on TypeName { ... }`) works correctly.

2. **Bug 2 — Null synthesis for fields not present in the recorded example.** When you work around Bug 1 by rewriting spreads to inline fragments, the query now requests **more fields** than were recorded in the example. Microcks responds with `null` for each of those extra fields instead of omitting them or returning an error. For real-world components that dereference nested objects without optional chaining, this causes frontend crashes that are hard to trace back to the mock layer.

---

## Environment

| Item | Version |
|------|---------|
| Microcks | 1.14.0 (observed; likely affects earlier versions too) |
| GraphQL transport | HTTP POST, `application/json` |

---

### Minimal reproduction

**This query fails with HTTP 500:**

```graphql
query GetExample($id: ID!) {
  example(id: $id) {
    ...ExampleCoreFields
    someSource {
      ...SomeSourceFields
    }
  }
}

fragment ExampleCoreFields on Example {
  id
  name
  status
}

fragment SomeSourceFields on someSource {
  product
}
```

**This semantically identical query succeeds:**

```graphql
query GetExample($id: ID!) {
  example(id: $id) {
    ... on Example {
      id
      name
      status
    }
    someSource {
      ... on SomeSourceFields {
        product
      }
    }
  }
}
```

The only difference is `...FragmentName` (named spread + separate fragment definition) vs `... on TypeName { ... }` (inline fragment). Both are valid GraphQL per the spec and must produce the same result.

### Why this matters

Almost every production GraphQL client (Apollo Client, urql, Relay) generates queries with named fragment spreads. This means any real-world application that uses code-generated queries will hit this bug immediately when pointed at Microcks. The workaround of rewriting spreads to inline fragments requires intercepting and AST-transforming every outgoing query on the client side — a significant burden.

### Client-side workaround (what we had to implement - Not what we want to have)

```typescript
import { Kind, parse, print } from 'graphql';
import type { DocumentNode, FragmentDefinitionNode, SelectionSetNode } from 'graphql';

/**
 * Recursively replaces every FragmentSpread with an InlineFragment,
 * then removes the now-redundant FragmentDefinition nodes.
 * Applied to every query before forwarding to Microcks.
 */
function inlineFragments(queryStr: string): string {
  let doc: DocumentNode;
  try { doc = parse(queryStr); } catch { return queryStr; }

  const fragmentMap: Record<string, FragmentDefinitionNode> = {};
  for (const def of doc.definitions) {
    if (def.kind === Kind.FRAGMENT_DEFINITION) fragmentMap[def.name.value] = def;
  }
  if (Object.keys(fragmentMap).length === 0) return queryStr;

  function inlineSelectionSet(ss: SelectionSetNode, visited = new Set<string>()): SelectionSetNode {
    const selections: SelectionSetNode['selections'][number][] = [];
    for (const sel of ss.selections) {
      if (sel.kind === Kind.FRAGMENT_SPREAD) {
        const frag = fragmentMap[sel.name.value];
        if (!frag || visited.has(sel.name.value)) {
          selections.push(sel);
        } else {
          selections.push({
            kind: Kind.INLINE_FRAGMENT,
            typeCondition: frag.typeCondition,
            directives: sel.directives ?? [],
            selectionSet: inlineSelectionSet(frag.selectionSet, new Set([...visited, sel.name.value])),
          });
        }
      } else if ('selectionSet' in sel && sel.selectionSet) {
        selections.push({ ...sel, selectionSet: inlineSelectionSet(sel.selectionSet, visited) } as typeof sel);
      } else {
        selections.push(sel);
      }
    }
    return { ...ss, selections };
  }

  return print({
    ...doc,
    definitions: doc.definitions
      .filter((d) => d.kind !== Kind.FRAGMENT_DEFINITION)
      .map((d) =>
        d.kind === Kind.OPERATION_DEFINITION
          ? { ...d, selectionSet: inlineSelectionSet(d.selectionSet) }
          : d,
      ),
  });
}
```

This runs in Node.js (Playwright test runner) and rewrites every query before it reaches Microcks. No changes on the Microcks side.

---

## Bug 2 — Null synthesis for fields not present in the recorded example

This bug is a consequence of working around Bug 1. Once fragment spreads are inlined, the forwarded query may request **fields that were not present in the original recording**. For example:

- The recorded example was captured with a minimal integration-test query requesting 10 fields.
- The real application query (after inlining its 5 fragments) requests 40 fields.
- Microcks returns a response with the 10 recorded fields populated, and **synthesises `null`** for the remaining 30.

### Why this is problematic

Returning `null` for fields that were not requested in the original recording is incorrect. The correct behaviours would be either:
- **Omit** the fields (since they were not in the recorded request), or
- **Return an error** indicating no example matches the requested shape, so the caller knows to provide a more complete recording.

Silently returning `null` is especially harmful for nested objects. A component that does `example.realTime.id` will throw `TypeError: Cannot read properties of null` rather than getting a "no match" signal from the mock layer. This is very hard to debug because the failure manifests in the UI, not in the network layer.

### Secondary workaround required

Because of Bug 2, large queries that contain fragments which expose many new fields after inlining cannot be fixed by inlining alone. We had to additionally:

1. Send a **hand-crafted minimal query** (only the fields present in the recording) to Microcks instead of the real application query.
2. **Backfill the missing fields** on the response before returning it to the browser.

```typescript
// Because of Bug 1 + Bug 2, we cannot forward the real GetExample query.
// Instead we send a minimal query that exactly matches the recorded fields,
// then patch the response with safe defaults for fields the FE needs.
const MINIMAL_GET_EXAMPLE_QUERY = `
  query GetExample($id: ID!, $skipSelfLink: Boolean! = false) {
    example(id: $id) {
      id name status createdAt detectedAt description
      selfLink @skip(if: $skipSelfLink)
    }
  }
`;
```

This is a significant maintenance burden: every time the recording is updated with new fields, the minimal query and the backfill logic must also be updated manually.

---

## Expected behaviour

1. Microcks should accept queries that use named fragment spreads and match them against recorded examples using the **resolved field set** (i.e., after expanding fragments), not the raw query string.
2. When a query requests fields that are not present in a recorded example, Microcks should either omit those fields or return an appropriate "no match" / "unknown field" signal — it should **not** synthesise `null`.

---

## Impact

Both bugs together make it impractical to use Microcks for mocking against real production GraphQL clients, since those clients almost universally use named fragments (often generated by tools like GraphQL Code Generator). The current behaviour forces consumers to maintain a non-trivial client-side query transformer and response patcher — work that should not be necessary.

DKey96:
## Problem

GraphQL clients (Apollo, urql, Relay) emit **fragment definitions before the operation** in the query document. `GraphQLController` assumed the operation is the first definition and blindly cast `getDefinitions().get(0)` to `OperationDefinition`, so any real client query using named fragment spreads throws:

```
java.lang.ClassCastException: class graphql.language.FragmentDefinition
cannot be cast to class graphql.language.OperationDefinition
    at io.github.microcks.web.GraphQLController.execute(...)
```

This is a different code path from the one fixed in #2201 (ab184ed) — that change hardened field *retention* in `filterFieldSelection`, but `execute()` throws before ever reaching it, so the 500 remains for real-world queries.

## Reproduction

Same query, only the definition order differs:

| Variant | Result |
|---|---|
| Named fragment spreads, **fragments defined first** (Apollo's native output) | **HTTP 500** |
| Same query, **operation moved before the fragments** | **HTTP 200** |
| Same query, fragments inlined (`... on Type {}`) | HTTP 200 |

Reordering the operation to the front makes the identical named-fragment query succeed, confirming the failure is the `get(0)` cast rather than the spreads themselves.

## Fix

- Select the operation via `getDefinitionsOfType(OperationDefinition.class)`, honoring the request `operationName` when present (mirrors the existing `getDefinitionsOfType(FragmentDefinition.class)` call just below).
- Harden two sibling casts that assumed a top-level selection is always a `Field`: guard the introspection check with `instanceof`, and flatten top-level fragment spreads via `collectFields()` before processing each queried field.

## Test

Adds a `GraphQLControllerIT` case asserting a query whose `fragment` definition precedes the operation returns 200 (verified failing before the change, passing after).

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/GraphQLController.java
+++ b/webapp/src/main/java/io/github/microcks/web/GraphQLController.java
@@ -36,7 +36,6 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import graphql.ExecutionResult;
 import graphql.GraphQL;
 import graphql.language.Argument;
-import graphql.language.Definition;
 import graphql.language.Document;
 import graphql.language.Field;
 import graphql.language.FragmentDefinition;
@@ -149,17 +148,31 @@ public class GraphQLController {
          return new ResponseEntity<>("Error parsing GraphQL request: " + e.getMessage(), HttpStatus.BAD_REQUEST);
       }
 
-      Definition<?> graphqlDef = graphqlRequest.getDefinitions().get(0);
-      OperationDefinition graphqlOperation = (OperationDefinition) graphqlDef;
+      // GraphQL clients (Apollo, urql, Relay...) emit fragment definitions before the operation, so the
+      // operation cannot be assumed to be the first definition of the document. Select it explicitly -
+      // by name when the request carries an operationName, otherwise the first operation found.
+      List<OperationDefinition> operationDefinitions = graphqlRequest.getDefinitionsOfType(OperationDefinition.class);
+      String requestedOperationName = graphqlHttpReq.getOperationName();
+      OperationDefinition graphqlOperation = operationDefinitions.stream().filter(op -> requestedOperationName == null
+            || requestedOperationName.isEmpty() || requestedOperationName.equals(op.getName())).findFirst()
+            .orElse(null);
+
+      if (graphqlOperation == null) {
+         log.error("No matching operation definition found in GraphQL request");
+         return new ResponseEntity<>("No matching operation definition found in GraphQL request",
+               HttpStatus.BAD_REQUEST);
+      }
 
       log.debug("Got this graphqlOperation: {}", graphqlOperation);
 
       // Operation type is direct but name depends on syntax (it can be a composite one). Better to use the names of selections...
       String operationType = graphqlOperation.getOperation().toString();
 
-      // Check is it's an introspection query to handle first.
-      if ("QUERY".equals(operationType) && INTROSPECTION_SELECTION
-            .equals(((Field) graphqlOperation.getSelectionSet().getSelections().get(0)).getName())) {
+      // Check is it's an introspection query to handle first. The first selection may be a fragment spread
+      // rather than a Field, so guard the cast.
+      Selection<?> firstSelection = graphqlOperation.getSelectionSet().getSelections().get(0);
+      if ("QUERY".equals(operationType) && firstSelection instanceof Field firstField
+            && INTROSPECTION_SELECTION.equals(firstField.getName())) {
          log.info("Handling GraphQL schema introspection query...");
          Resource graphqlSchema = resourceRepository
                .findByServiceIdAndType(service.getId(), ResourceType.GRAPHQL_SCHEMA).get(0);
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
Awesome! This looks great. Thank you.

PR #2204, opened 2026-07-02
feat: Support Workload Identity (ADC) for Google Pub/Sub authentication

### LINKED ISSUE #2120: Support Workload Identity for Google Pub/Sub Authentication
### Reason/Context

When running Microcks on Google Kubernetes Engine (GKE), Workload Identity is the recommended way to authenticate to Google Cloud services. Right now we have to provide JSON key of service account with Pub/Sub permissions. The same service account could be used for Workload Identity, then we wouldn't have to provide any json key to Microcks :)

### Description

Using Application Default Credentials when initializing connection to Pub/Sub. It may be an addition to current SA json key authentication, then it's not a breaking change.

### Implementation ideas

_No response_

priyamkarn:
Closes #2120

Adds support for GKE Workload Identity / Application Default Credentials
(ADC) as an authentication option for Google Pub/Sub, in both the async
minion's producer (mocking) and consumer (test) paths.

Previously, if no service-account JSON key was configured, the code
fell back straight to NoCredentialsProvider, which only works against
the local emulator and fails against real GCP Pub/Sub.

Now it first attempts GoogleCredentials.getApplicationDefault(), which
transparently picks up Workload Identity credentials on GKE (or any
other ADC source), and only falls back to NoCredentialsProvider if
that also fails — preserving existing behavior for local/dev setups.

This is non-breaking: existing JSON-key-based configuration is
unaffected and still takes priority when present.

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/consumer/GooglePubSubMessageConsumptionTask.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/consumer/GooglePubSubMessageConsumptionTask.java
@@ -178,7 +178,15 @@ public class GooglePubSubMessageConsumptionTask implements MessageConsumptionTas
          credentialsProvider = FixedCredentialsProvider
                .create(GoogleCredentials.fromStream(is).createScoped(CLOUD_OAUTH_SCOPE));
       } else {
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
LGTM. Thanks a lot!

PR #2209, opened 2026-07-06
feat: show error feedback in UI on early test failure

### LINKED ISSUE #1056: Provide an error message when the test fails for Authorization Reasons
### Reason/Context

When the test fails for OAuth reasons I see no feedback in the UI, it would improve the overall experience if the error was shown write after the test fails.

### Description

Whenever a test fail for known reasons, provide the feedback text back to the user.

### Implementation ideas

_No response_

alokkumardalei-wq:
## What does this PR do?
This PR provides clear UI feedback to the user whenever a test fails during its early setup phase (e.g., due to OAuth negotiation failures, malformed URIs, or runner setup errors).

Previously, if a test failed for known reasons like an AuthorizationException during the OAuth2 token flow, the failure was silently logged in the backend and the UI provided no immediate feedback. With these changes, the backend now explicitly captures these early failures and passes a runnerMessage to the frontend, which is then rendered as a prominent error alert directly in the UI.

Fixes #1056

## How was this tested?

- Backend: Verified that TestRunnerService properly populates the runnerMessage inside TestResult during AuthorizationException, URISyntaxException, and runner retrieval failures. Ran the full backend test suite (mvn clean test) to ensure no regressions.
- Frontend / UI:

- - Spun up the Microcks instance locally using Docker Compose.
- - Imported a sample API (Petstore API).
- -  Attempted to launch a new test with intentionally invalid OAuth credentials (wrong token URI, client ID, etc.).
- - Verified that a red alert banner displaying Test failed: OAuth2 token flow failed: ... appears immediately at the top of both the Test Details and Test Runner pages.

test_failure_message_1783345304967

### DIFF
--- a/commons/model/src/main/java/io/github/microcks/domain/TestResult.java
+++ b/commons/model/src/main/java/io/github/microcks/domain/TestResult.java
@@ -47,6 +47,7 @@ public class TestResult {
    private TestRunnerType runnerType;
    private OperationsHeaders operationsHeaders;
    private OAuth2AuthorizedClient authorizedClient;
+   private String runnerMessage;
 
    private List<TestCaseResult> testCaseResults = new ArrayList<>();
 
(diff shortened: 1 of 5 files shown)


lbroudoux:
Makes sense. Thanks for the contribution. I'm gonna review it now.

lbroudoux (approved):
LGTM. Thank you!

PR #2214, opened 2026-07-09
feat: Support External References in OpenAPI MCP Tool Converter

### LINKED ISSUE #2210: Enhancement (TODO): Support External References in OpenAPI MCP Tool Converter
### Reason/Context

**Why we need this improvement:**
Currently, when Microcks generates an MCP tool schema from an OpenAPI specification, it only supports internal references (i.e., those starting with `#/`). If an enterprise uses a modular, multi-file OpenAPI specification that relies on external `$ref`s (e.g. `$ref: './paths/problemDetail.yaml'`), the `OpenAPIMcpToolConverter.java` method `getNodeForRef()` fails to resolve it and returns `null` because of an open TODO: `// TODO: handle external references reusing imported resources?`

**How will this change help:**
It ensures that MCP tools generated from multi-file OpenAPI specs include all required parameters and properties, preventing AI agents from failing due to incomplete schema definitions.

**What is the motivation:**
MCP (Model Context Protocol) is rapidly becoming the standard for AI tool integrations. Supporting robust, enterprise-grade OpenAPI definitions that rely on external references is essential for a seamless AI agent experience in Microcks.

### Description

During the initial OpenAPI import, `AbstractJsonRepositoryImporter.resolveExternalReferences` successfully fetches these external schemas, normalizes their names, and stores them in the database as secondary `Resource` objects (linked to the same `Service`). 

However, `OpenAPIMcpToolConverter` currently does not have access to these resources. This enhancement will provide the converter access to the repository so it can retrieve these already-imported resources to complete the schema resolution. 

This will not introduce any breaking changes.

### Implementation ideas

I have been researching the codebase and have a clear implementation plan for how to solve this:

1. Modify `OpenAPIMcpToolConverter.java` and its constructor to accept `ResourceRepository`.
2. Update the `buildMcpToolConverter` factory method in `McpController.java` to pass its already-injected `ResourceRepository` down to the converter.
3. Inside `getNodeForRef()` in `OpenAPIMcpToolConverter`, implement the TODO for external references by querying `resourceRepository.findByServiceIdAndType(service.getId(), ResourceType.OPEN_API_SPEC)`.
4. Match the path of the external reference against the returned resources' `path` or encoded `name`.
5. Once matched, parse the resource content into a `JsonNode` (caching it in a `HashMap` to prevent re-parsing during the same request) and return the specific json pointer within that file.
6. Add comprehensive unit tests in `OpenAPIMcpToolConverterTest.java` with a mocked `ResourceRepository` to verify the external resolution.

I would love to be assigned to this issue and submit a PR for it if you agree with this approach!

dolliecoder:
### Description

This PR implements the missing capability for the Microcks MCP Server to resolve and extract schemas from external `$ref` definitions (such as separate JSON or YAML schema files) when converting OpenAPI specifications into MCP tools.

- **Injected `ResourceRepository`:** Updated the `McpController` and `OpenAPIMcpToolConverter` constructors to inject the `ResourceRepository`, allowing the converter to fetch externally attached `Resource` files linked to the `Service`.
- **Implemented External Reference Resolution:** Ported the `getNodeForExternalRef` logic from the `reshaprio/reshapr` project. The algorithm correctly parses external file paths and JSON pointers (e.g., `models.yaml#/components/schemas/UserModel`), locates the corresponding resource in the repository, and extracts the exact schema slice required.
- **Added Caching:** Implemented lazy parsing and caching of JSON/YAML ASTs via `externalResourcesContent` (`Map<Resource, JsonNode>`) to ensure high-performance schema resolution without redundant parsing on subsequent tool calls.
- **Added Integration Test:** Added the `testExternalReferenceResolution` test case to `OpenAPIMcpToolConverterTest` which successfully mimics a multipart OpenAPI specification with external `$ref`s and asserts that the final `McpSchema.JsonSchema` contains the correctly mapped properties.

### Related issue(s)
Resolves #2210

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
+++ b/webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
@@ -66,7 +66,7 @@ public class OpenAPIMcpToolConverter extends McpToolConverter {
    private final ResourceRepository resourceRepository;
 
    private JsonNode schemaNode;
-   private final Map<Resource, JsonNode> externalResourcesContent = new HashMap<>();
+   private final Map<String, JsonNode> externalResourcesContent = new HashMap<>();
    private ObjectMapper yamlMapper;
 
    /**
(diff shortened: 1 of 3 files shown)


dolliecoder:
@lbroudoux PTAL when you get time

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
```diff
@@ -260,7 +270,61 @@ private JsonNode getNodeForRef(String reference) {
       if (reference.startsWith("#/")) {
          return schemaNode.at(reference.substring(1));
       }
-      // TODO: handle external references reusing imported resources?
+      return getNodeForExternalRef(reference);
+   }
+
+   /** Get the JsonNode for reference that is localed in external resource. */
+   private JsonNode getNodeForExternalRef(String externalReference) {
```
SebastienDegodez: Could we refactor this method into smaller units?
Now this combine:
- external reference parsing
- resource lookup
- JSON/YAML parsing
etc ...

Breaking this into focused private methods would make the logic easier to review and safer to evolve.

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
```diff
@@ -260,7 +270,61 @@ private JsonNode getNodeForRef(String reference) {
...
+         for (Resource attachedResource : attachedResources) {
+            // Skip the main resource
+            if (attachedResource.getId() != null && attachedResource.getId().equals(resource.getId())) {
+               continue;
+            }
+            // Path direct equality is for absolute ref ("http://raw.githubusercontent.com/...")
+            // Path equality with resource name if for relative refs that have been re-normalized ("Service name+Service version+...))
+            if (attachedResourceName.equals(attachedResource.getPath()) || attachedResourceName
+                  .equals(URLEncoder.encode(attachedResource.getName(), StandardCharsets.UTF_8))) {
+               JsonNode resourceNode = externalResourcesContent.computeIfAbsent(attachedResource, k -> {
```
SebastienDegodez: Could we switch the cache key from Resource object instances to a stable key (resource id/path) ?

SebastienDegodez (changes requested):

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
dolliecoder: Good call. I've split getNodeForExternalRef into parseExternalReference, findMatchingResource, and parseAndCacheResource each with a single responsibility. PTAL.

webapp/src/main/java/io/github/microcks/util/openapi/OpenAPIMcpToolConverter.java
dolliecoder: Agreed. Switched from Resource instances to resource.getId() as the cache key with a null-ID fallback.

dolliecoder:
@SebastienDegodez PTAL

lbroudoux (approved):
This looks all good! Thanks for this contribution and for your patience!

PR #2215, opened 2026-07-09
refactor: graphql test return message

A-Deepak1610:
### Description
Fixed incorrect newline formatting in GraphQL test validation error messages in `GraphQLTestRunner`.

- Previously, validation errors were concatenated using the literal string `"/n"` instead of a proper newline character, producing malformed output.
- Replaced manual `StringBuilder` concatenation with `String.join(System.lineSeparator(), ...)` for clearer, more idiomatic error message construction.
- Wrapped the logic in a `finally` block to guarantee `lastValidationErrors` is reset after consumption, avoiding potential state leakage between test runs.

### Testing
Verified locally via `mvn clean install` and `mvn spotless:check` — build passes with no formatting violations.

### Related issue(s)
No related issue — found while reviewing the codebase.


A-Deepak1610:
Is there any update on this pr

PR #2225, opened 2026-07-16
fix: escape single quotes in curl command payload #2224

### LINKED ISSUE #2224: Bug report:  "Copy curl command to clipboard" breaks when the body contains a single quote
## Describe the bug

The "Copy curl command to clipboard" action on a Service/Operation detail page generates a broken command when the example request body (or a header value) contains a single quote (`'`).

The body is wrapped in single quotes but inner single quotes are not escaped. As soon as the payload contains an apostrophe — very common in French text or JSON string values — the first inner `'` closes the shell-quoted string and the rest of the payload is parsed by the shell as separate arguments, so the command cannot be run as-is.

## Expected behavior

The copied curl command should be runnable as-is, regardless of the characters contained in the body, URL or headers. Single quotes must be escaped for the shell (POSIX idiom `'\''`).

## Actual behavior

The generated command is:

```bash
curl -X POST '.../chat/completions?api-version=2024-06-01' \
  -H 'Content-Type: application/json' \
  -d '{ "content": "Tu es un assistant d'extraction de donnees" }'
```

Running it fails because the `'` in `d'extraction` terminates the `-d '...'` argument:

```text
curl: (6) Could not resolve host: de
curl: (6) Could not resolve host: donnees...
curl: (3) unmatched close brace/bracket in URL position ...
```

## How to Reproduce?

Reproducer:

1. Import any service whose example request body contains an apostrophe, e.g.:

   ```json
   { "content": "Tu es un assistant d'extraction de donnees" }
   ```

2. Open the Service detail page, select the operation, then select the example exchange.
3. Click **Copy curl command to clipboard**.
4. Paste the command in a terminal and run it — it fails.

Root cause:

In `webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.ts`, method `formatCurlCmd()`, values are interpolated into single-quoted shell segments without escaping:

```ts
cmd = 'curl -X ' + verb + ' \'' + mockUrl + '\'';
// ...
cmd += ` -H '${header.name}: ${header.values.join(', ')}'`;
// ...
cmd += ' -d \'' + normalizedRequestContent.replace(/\n/g, '') + '\'';
```

Suggested fix: escape single quotes with the POSIX idiom `'\''` via a small helper applied to the URL, header values and body, e.g.:

```ts
const shellSingleQuote = (s: string): string =>
  `'${s.replace(/'/g, `'\\''`)}'`;

cmd += ' -d ' + shellSingleQuote(normalizedRequestContent.replace(/\n/g, ''));
```

## Environment

- **Microcks version or git rev**: `1.15.0-SNAPSHOT`
- **Install method**: `docker-compose` / `podman-compose` (microcks-uber image)

## Additional information

- Component: Web UI (Angular) — `service-detail.page.ts`, `formatCurlCmd()`.
- The same escaping issue applies to header values and the mock URL, not only the body.

SebastienDegodez:
This pull request improves the handling and testing of cURL command generation in the `ServiceDetailPageComponent`. The main focus is on correctly escaping apostrophes in JSON payloads to ensure the generated cURL commands remain valid and functional, especially when the payload contains single quotes.

**Bug Fixes and Improvements:**

* Properly escape apostrophes in request payloads when generating cURL commands, ensuring that payloads with single quotes produce valid shell commands. (`service-detail.page.ts`, [webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.tsL939-R940](diffhunk://#diff-b000fd75abf6789c0bc64a9641e829920aa2b8d8f8b7308efa0cc17df7f8e46aL939-R940))

**Testing Enhancements:**

* Add a new unit test suite (`service-detail.page.spec.ts`) that verifies cURL command generation, specifically testing that JSON payloads containing apostrophes are correctly escaped and can be parsed back to their original content.

### DIFF
--- a/webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.spec.ts
+++ b/webapp/src/main/webapp/src/app/pages/services/{serviceId}/service-detail.page.spec.ts
@@ -1,0 +1,66 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { ServiceType } from '../../../models/service.model';
+import { ServiceDetailPageComponent } from './service-detail.page';
+
+describe('ServiceDetailPageComponent formatCurlCmd', () => {
+  function createComponentForCurl(): ServiceDetailPageComponent {
+    const component = Object.create(ServiceDetailPageComponent.prototype) as ServiceDetailPageComponent;
+
+    (component as any).resolvedServiceView = {
+      service: {
+        type: ServiceType.REST,
+        name: 'OpenAI API Documentation',
+        version: '1.0.0'
+      }
+    };
+
+    (component as any).formatMockUrl = () =>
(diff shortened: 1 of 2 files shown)


lbroudoux (approved):
LGTM. Thanks @SebastienDegodez

PR #2229, opened 2026-07-18
fix(web): make test runner connect timeout configurable

### LINKED ISSUE #2223: Hardcoded 200ms connect timeout in TestRunnerService causes false ConnectTimeoutException on higher-latency networks
## Description

When running a Test Runner test (e.g. `OPEN_API_SCHEMA` runner) against a test endpoint that is reachable but has moderate network latency (e.g. reached via corporate VPN routing to an internal environment), all test calls fail immediately with:

```
org.apache.hc.client5.http.ConnectTimeoutException: Connect to https://<test-endpoint>:443 failed: Connect timed out
```

The reported response time for the failed call was only ~600ms, which is far too fast for a genuine network-level connection timeout (those typically take several seconds to multiple tens of seconds to fail at the OS level). This pointed to an artificially short timeout being enforced somewhere in Microcks itself, rather than an actual connectivity problem.

## Root cause

In `TestRunnerService.java`, the HTTP client used for test execution has a **hardcoded connect timeout of 200 milliseconds**, while only the response timeout is configurable via `runnerTimeout`:

https://github.com/microcks/microcks/blob/5bc56ec2d1ea4615092c9ce07ba928d6c6ae607c/webapp/src/main/java/io/github/microcks/service/TestRunnerService.java#L380-L452

```java
CloseableHttpClient httpClient = HttpClients.custom().setConnectionManager(connectionManager)
      .setDefaultRequestConfig(RequestConfig.custom().setConnectTimeout(Timeout.ofMilliseconds(200))
            .setResponseTimeout(Timeout.ofMilliseconds(runnerTimeout)).build())
      .build();
```

200ms is too aggressive for many real-world scenarios, e.g.:
- Test endpoints reached through corporate VPNs
- Endpoints on internal/staging networks with extra routing hops
- Any environment where TCP+TLS handshake naturally takes longer than 200ms, even though the endpoint is perfectly reachable

## Steps to Reproduce

1. Run Microcks (e.g. via `quay.io/microcks/microcks-uber:latest` in Docker).
2. Import an OpenAPI spec for a service reachable at an internal/VPN-routed HTTPS endpoint whose connection setup (DNS + TCP + TLS handshake) takes more than 200ms but well under any reasonable timeout (confirmed reachable and functional via a manual `curl` call from inside the same container, which succeeds with a normal HTTP response).
3. Create a New Test in the UI targeting that endpoint, with runner type `OPEN_API_SCHEMA` (or others using `HttpTestRunner`).
4. Launch the test.

## Expected Behavior

The test should be able to connect to the endpoint successfully, since the endpoint is reachable (as verified independently via `curl` from within the same container), given a reasonable amount of time to establish the connection.

## Actual Behavior

All test operations fail immediately (~600ms) with:
```
Connect to https://<test-endpoint>:443 failed: Connect timed out
```

Manually running `curl -v https://<test-endpoint>` **from inside the same running Microcks container** succeeds without issue, confirming the endpoint is reachable and the failure is due to Microcks' internal timeout configuration rather than actual network/DNS/firewall problems.

## Suggested Fix

Make the connect timeout configurable (e.g. via an environment variable or application property, similar to how `runnerTimeout`/response timeout is already configurable), rather than hardcoding it to 200ms. Alternatively, raise the default to a more realistic value (e.g. 5000ms) to better accommodate real-world network conditions such as VPN-routed or higher-latency internal endpoints.

## Environment

- Microcks image: `quay.io/microcks/microcks-uber:latest`
- Deployment: Local Docker (`docker run`)
- Test Runner type: `OPEN_API_SCHEMA` (issue likely applies to all runners extending `HttpTestRunner`)

dipak0000812:
### Description

- Introduce a configurable `tests-runner.connect-timeout` property with a default value of `5000` ms.
- Replace the previously hardcoded HTTP client connect timeout (`200` ms) with the configured value when creating the test runner HTTP client.
- Expose the property in both the default application configuration and the Kubernetes ConfigMap so it can be overridden through `TESTS_RUNNER_CONNECT_TIMEOUT`.

### Related issue(s)

Fixes #2223

### DIFF
--- a/webapp/src/main/resources/config/application.properties
+++ b/webapp/src/main/resources/config/application.properties
@@ -21,6 +21,7 @@ management.metrics.distribution.percentiles-histogram.http.server.requests=true
 management.metrics.distribution.slo."http.server.requests"=1ms, 5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms, 1000ms, 2500ms, 5000ms, 10000ms
 
 tests-callback.url=${TEST_CALLBACK_URL:http://localhost:8080}
+tests-runner.connect-timeout=${TESTS_RUNNER_CONNECT_TIMEOUT:5000}
 postman-runner.url=${POSTMAN_RUNNER_URL:http://localhost:3000}
 async-minion.url=${ASYNC_MINION_URL:http://localhost:8081}
 default-artifacts-repository.url=${DEFAULT_ARTIFACTS_REPOSITORY_URL:#{null}}
--- a/install/kubernetes/microcks/templates/configmap.yaml
+++ b/install/kubernetes/microcks/templates/configmap.yaml
@@ -54,6 +54,7 @@ data:
   application.properties: |-
     # Application configuration properties
     tests-callback.url=${TEST_CALLBACK_URL}
+    tests-runner.connect-timeout=${TESTS_RUNNER_CONNECT_TIMEOUT:5000}
     postman-runner.url=${POSTMAN_RUNNER_URL}
     async-minion.url=${ASYNC_MINION_URL|http://localhost:8081}
 
--- a/webapp/src/main/java/io/github/microcks/service/TestRunnerService.java
+++ b/webapp/src/main/java/io/github/microcks/service/TestRunnerService.java
@@ -102,6 +102,9 @@ public class TestRunnerService {
    @Value("${tests-callback.url}")
    private String testsCallbackUrl = null;
 
+   @Value("${tests-runner.connect-timeout:5000}")
+   private long runnerConnectTimeout;
+
    @Value("${postman-runner.url}")
    private String postmanRunnerUrl = null;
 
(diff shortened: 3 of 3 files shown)


SebastienDegodez (approved):

webapp/src/main/resources/config/application.properties
```diff
@@ -21,6 +21,7 @@ management.metrics.distribution.percentiles-histogram.http.server.requests=true
 management.metrics.distribution.slo."http.server.requests"=1ms, 5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms, 1000ms, 2500ms, 5000ms, 10000ms
 
 tests-callback.url=${TEST_CALLBACK_URL:http://localhost:8080}
+tests-runner.connect-timeout=${TESTS_RUNNER_CONNECT_TIMEOUT:5000}
```
lbroudoux, line 24: Could we just change this env to `TEST_RUNNER_CONNECT_TIMEOUT`? I think it would be more consistent with `TEST_CALLBACK_URL`.

lbroudoux (changes requested):
It looks good but I suggest changing the name of environment variable for better consistency.

PR #2244, opened 2026-07-30
feat: compare OpenAPI response examples

### LINKED ISSUE #2243: JSON comparison of OpenAPI response examples with actual
### Reason/Context

With @Basuchi, we noticed that some of our OpenAPI response examples did not match the actual JSON from the implementation, yet the contract tests were passing in our CI.
The mismatch was caught by an agent during the code review, but it may not have been spotted by humans.

The motivation would be to detect incoherent examples, based on the integration tests.

But maybe this is out-of-scope of conformance testing & Microcks, which covers JSON schema, HTTP status & content-type.

### Description

Add an optional parameter in the /tests API to do assertions that the response examples are contained in the response body.

There could be 2 modes:
* STRICT - exact match, but array ordering doesn't matter.
* LENIENT - partial match: for expected elements, the value has to be an exact match, but ignore extra fields in JSONObject or extra elements in JSONArray.

### Implementation ideas

In JUnit, the library https://github.com/skyscreamer/JSONassert/blob/master/src/main/java/org/skyscreamer/jsonassert/JSONAssert.java does this.
The modes are described in https://github.com/skyscreamer/JSONassert/blob/master/src/main/java/org/skyscreamer/jsonassert/JSONCompareMode.java

ME-Massine:
## What does this PR do?

  Adds optional JSON response-example comparison to the `/tests` API for `OPEN_API_SCHEMA` tests.

  Two comparison modes are supported:

  - `STRICT`: requires the same JSON structure and values while ignoring array ordering.
  - `LENIENT`: requires all expected values while allowing additional object fields and array elements.

  The comparison runs after the existing status-code, content-type, and schema validation. When the option is omitted, existing behavior remains unchanged.

  The OpenAPI specification and frontend TypeScript model are also updated.

  Closes #2243

  ## How was this tested?

  - Added matcher tests covering strict matching, lenient matching, unordered arrays, duplicate elements, and ambiguous array matches.
  - Added OpenAPI runner tests covering both modes and backward compatibility.
  - Added controller tests covering JSON deserialization and option propagation.
  - Java 25 Maven reactor build completed successfully.
  - 11 focused tests passed with 0 failures.
  - Spotless formatting and checks passed.
  - OpenAPI YAML parsing passed.
  
  > FOSSA reports 5 License Compliance issues, but the same check also fails on the PR’s base commit. This PR does not modify any dependency manifests or introduce new dependencies, so the findings appear to be pre-existing baseline issues.

### DIFF
--- a/api/microcks-openapi-v1.15.yaml
+++ b/api/microcks-openapi-v1.15.yaml
@@ -1536,6 +1536,10 @@ components:
         operationsHeaders:
           $ref: '#/components/schemas/OperationHeaders'
           description: This test operations headers override
+        responseExampleComparison:
+          $ref: '#/components/schemas/ResponseExampleComparisonMode'
+          description: Optional JSON response example comparison for OPEN_API_SCHEMA
+            tests
     Resource:
       description: "Resource represents a Service or API artifacts such as specification,\
         \ contract"
@@ -1581,6 +1585,15 @@ components:
       - ASYNC_API_SCHEMA
       - GRPC_PROTOBUF
       - GRAPHQL_SCHEMA
+      type: string
+    ResponseExampleComparisonMode:
+      description: "JSON response example comparison mode. STRICT requires the same\
+        \ structure and values while ignoring array ordering. LENIENT requires the\
+        \ example structure and values while allowing additional object fields and\
+        \ array elements."
+      enum:
+      - STRICT
+      - LENIENT
       type: string
     ArtifactUpload:
       description: |-
--- a/commons/model/src/main/java/io/github/microcks/domain/ResponseExampleComparisonMode.java
+++ b/commons/model/src/main/java/io/github/microcks/domain/ResponseExampleComparisonMode.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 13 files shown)


SebastienDegodez:
I think it would be ideal if the 'enum' were never null. That would simplify passing arguments to constructors by initializing it directly to the default behavior. 

What do you think?

lbroudoux:
Thanks for jumping on this but the request for enhancement has to be carefully analysed first. As described here: https://github.com/microcks/microcks/issues/2243#issuecomment-5129124906 I'm not sure this is the right approach and I'm opening the discussion on other (IMHO better) alternatives. Please join us on the issue threads if you have some opinion on this!

ME-Massine:
> I think it would be ideal if the 'enum' were never null. That would simplify passing arguments to constructors by initializing it directly to the default behavior.
> 
> What do you think?

@SebastienDegodez  Good point. null currently means “disabled” to preserve existing behavior. If we keep this approach, I can replace it with an explicit DISABLED default.

ME-Massine:
> Thanks for jumping on this but the request for enhancement has to be carefully analysed first. As described here: [#2243 (comment)](https://github.com/microcks/microcks/issues/2243#issuecomment-5129124906) I'm not sure this is the right approach and I'm opening the discussion on other (IMHO better) alternatives. Please join us on the issue threads if you have some opinion on this!

@lbroudoux Thanks for the context. I see the distinction now. I’ll pause changes here and continue the discussion in #2243 first.

PR #2247, opened 2026-08-03
Feature/amqps support 2233

### LINKED ISSUE #2233: async-minion: support `amqps://` for Amazon MQ (port 5671) and non-fatal AMQP startup
### Reason/Context

Why we need this improvement?

I deploy async-minion on AWS ECS Fargate against Amazon MQ for RabbitMQ (AMQP 0-9-1 over TLS on port 5671). Amazon MQ requires the amqps:// URI scheme; plain amqp:// cannot connect on this port. Today the task never becomes healthy on ECS.

How will this change help?

Correct URI scheme detection would allow async-minion to connect to managed RabbitMQ brokers that enforce TLS (Amazon MQ, and similar :5671 deployments) using existing amqp.server configuration. A non-fatal AMQP startup would also let Kafka and Sync-to-Async keep running when the broker is temporarily unavailable.

What is the motivation?

In Microcks 1.14.0, AMQPProducerManager.createConnection() always builds factory.setUri("amqp://" + amqpServer). This is not configurable via application.properties. Consequences:

Connection to Amazon MQ on :5671 fails (client attempts plain AMQP without TLS).

The @PostConstruct connection failure aborts Quarkus startup → ECS reports 1 tasks failed to start, deploy timeout.

Version: 1.14.0.

### Description

Update AMQPProducerManager to support TLS AMQP brokers (e.g. Amazon MQ on port 5671) and resilient startup when the broker is temporarily unavailable.

Changes to introduce:

Auto-detect or honor amqps:// in amqp.server instead of hardcoding amqp://.

Make AMQP broker connection failure non-fatal at startup so Kafka / Sync-to-Async remain available.

Add unit tests for URI construction.

Breaking change? No for typical non-TLS setups (amqp:// on standard ports). Behavior change only when port 5671 or an explicit amqps:// scheme is used (intended fix). Making AMQP startup non-fatal is a resilience improvement: the process starts even if AMQP is down; AMQP mocking is disabled until the broker is reachable.

### Implementation ideas

Suggested changes in minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java:

```
@PostConstruct
public void create() {
    try {
        amqpConnection = createConnection();
    } catch (Exception e) {
        logger.errorf("Cannot connect to AMQP broker %s: %s", amqpServer, e.getMessage());
        amqpConnection = null;
    }
}

protected Connection createConnection() throws Exception {
    ConnectionFactory factory = new ConnectionFactory();
    String uri = amqpServer.contains("://") ? amqpServer
            : (amqpServer.contains(":5671") ? "amqps://" : "amqp://") + amqpServer;
    factory.setUri(uri);
    factory.setConnectionTimeout(15000);
    factory.setHandshakeTimeout(15000);
    // ... credentials ...
    return factory.newConnection(amqpClientId);
}

public void publishMessage(...) {
    if (amqpConnection == null) {
        logger.warnf("Skipping AMQP publish to {%s} — broker not connected", destinationName);
        return;
    }
    // ...
}
```

Suggested tests:

unit test: amqp.server=b.example.com:5671 → URI amqps://b.example.com:5671

unit test: amqp.server=amqps://b.example.com:5671 → no double scheme

optional integration test (Testcontainers RabbitMQ TLS)

I have already validated this patch in production (async-minion on ECS against Amazon MQ for RabbitMQ). If you agree with this approach, I'm happy to submit a PR with the minimal patch above and unit tests.

Gautam-aman:
## Summary

This PR improves the AMQP producer used by the async minion by:

- Supporting explicit `amqp://` and `amqps://` URIs.
- Automatically using `amqps://` when the configured broker uses port `5671`.
- Allowing the async minion to start even when the AMQP broker is temporarily unavailable.
- Skipping message publication when no AMQP connection is available instead of failing.
- Adding unit tests covering AMQP URI construction.

## Changes

- Added `buildAmqpUri()` to centralize AMQP URI creation.
- Updated `createConnection()` to use the generated URI.
- Made AMQP connection failures during startup non-fatal by logging the error and continuing startup.
- Added a null guard in `publishMessage()` to safely skip publishing when no broker connection exists.
- Added unit tests for:
  - Default `amqp://` URI generation
  - Standard port (`5672`)
  - TLS port (`5671`)
  - Explicit `amqp://` URIs
  - Explicit `amqps://` URIs

## Testing

- Ran `mvn spotless:apply`
- Verified Spotless checks pass
- Added unit tests for URI construction

Fixes #2233

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
@@ -75,7 +75,7 @@ public class AMQPProducerManager {
       } catch (Exception e) {
          logger.errorf("Cannot connect to AMQP broker %s", amqpServer);
          logger.errorf("Connection exception: %s", e.getMessage());
-         throw e;
+         amqpConnection = null;
       }
    }
 
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
@@ -69,27 +69,26 @@ public class AMQPProducerManager {
     * @throws Exception If connection to AMQP Broker cannot be done.
     */
    @PostConstruct
-   public void create() throws Exception {
+   public void create() {
       try {
          amqpConnection = createConnection();
       } catch (Exception e) {
-         logger.errorf("Cannot connect to AMQP broker %s", amqpServer);
-         logger.errorf("Connection exception: %s", e.getMessage());
+         logger.errorf(e, "Cannot connect to AMQP broker %s", amqpServer);
          amqpConnection = null;
       }
    }
 
-  private String buildAmqpUri() {
-    if (amqpServer.contains("://")) {
-      return amqpServer;
-    }
+   static String buildAmqpUri() {
+      if (amqpServer.startsWith("amqp://") || amqpServer.startsWith("amqps://")) {
+         return amqpServer;
+      }
 
-    if (amqpServer.endsWith(":5671")) {
-      return "amqps://" + amqpServer;
-    }
+      if (amqpServer.endsWith(":5671")) {
+         return "amqps://" + amqpServer;
+      }
 
-    return "amqp://" + amqpServer;
-  }
+      return "amqp://" + amqpServer;
+   }
 
    /**
     * @return A newly created connection to configured broker
@@ -120,8 +119,8 @@ public class AMQPProducerManager {
    public void publishMessage(String destinationType, String destinationName, String routingKey, String value,
          Set<Header> headers) {
       if (amqpConnection == null) {
-        logger.warn("Skipping AMQP publish because broker is not connected");
-        return;
+         logger.warnf("Skipping AMQP publish to {%s}: broker is not connected", destinationName);
+         return;
       }
       logger.infof("Publishing on destination {%s}, message: %s ", destinationName, value);
       try (Channel channel = amqpConnection.createChannel()) {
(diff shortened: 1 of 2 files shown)


dipak0000812 (commented):
One concern with making the initial connection failure non-fatal: setAutomaticRecoveryEnabled(true) does not retry an initial newConnection() failure. If the broker is unavailable when the minion starts, amqpConnection remains null, and the new guard in publishMessage() will skip subsequent publishes even after the broker becomes available.

Could we add a retry/reconnect path for the initial connection, or otherwise clarify how the producer is expected to recover in this case?

dipak0000812 (commented):
What happens if an established AMQP connection is lost while publishMessage() is running? The current guard only checks for null. Could we also handle a closed connection explicitly so a broker interruption doesn't result in an unexpected failure from createChannel()?

minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
```diff
@@ -63,20 +63,40 @@ public class AMQPProducerManager {
...
-         logger.errorf("Cannot connect to AMQP broker %s", amqpServer);
-         logger.errorf("Connection exception: %s", e.getMessage());
-         throw e;
+         logger.warnf(e, "Cannot connect to AMQP broker %s", amqpServer);
+         amqpConnection = null;
+         return false;
+      }
+   }
+
+   String buildAmqpUri() {
```
lbroudoux: Please add some visbility here - protected or private

lbroudoux:
> Could we add a retry/reconnect path for the initial connection, or otherwise clarify how the producer is expected to recover in this case?

I think it's better not to recover here. If the connection fails, that means that it's either a configuration or a unavailability issue. In both case, it's better to fail early and as we are in a `@PostConstruct` method to have a direct impact on the minion and have its probe fail.

> What happens if an established AMQP connection is lost while publishMessage() is running? 

I tend to think that we should keep it simple here: publication fails, probe fails and the minion container restarts until the connection is OK again. It's better to die early and keep recovery simple IMHO. Don't bother with this here.

lbroudoux:
@Gautam-aman Any follow-up on the previous review? Let us know.

Gautam-aman:
No i have stopped working on this.

lbroudoux:
> No i have stopped working on this.

Ok noted. Thanks for the quick response. I'll close the PR to let someone pick up the topic again.

PR #2274, opened 2026-08-16
fix: retain examples from all messages of an AsyncAPI 3 operation

### LINKED ISSUE #2273: AsyncAPI 3.x import: only the last message's examples are retained for multi-message operations
## Describe the bug

When importing an AsyncAPI 3.x document where a single operation references **multiple messages** (the v3 equivalent of an AsyncAPI 2.x `oneOf` message), only the examples of the **last** referenced message are retained. All other messages' examples are silently dropped, so the async minion mock publishes only that last message's example instead of cycling through all of them.

The AsyncAPI 2.x importer handles this correctly: `AsyncAPIImporter` walks `oneOf`/`anyOf`/`allOf` and aggregates examples from every message. The v3 importer does not.

Note this is a different path from multiple *examples on one message* (as in the [first AsyncAPI mock tutorial](https://microcks.io/documentation/tutorials/first-asyncapi-mock/)), which works fine — the bug is in the loop across *messages*.

## Likely cause

In `AsyncAPI3Importer.buildEventMessages` (tag 1.10.1, ~line 441; still present on current master), the result list is **reassigned** on each loop iteration instead of accumulated:

```java
List<EventMessage> eventMessages = null;
while (messages.hasNext()) {
   ...
   if (messageName != null && messageNode.has(EXAMPLES_NODE)) {
      eventMessages = buildEventMessageFromExamples(...);   // reassigned each iteration
   }
}
return eventMessages;
```

## To Reproduce

Import this minimal spec as a primary artifact (`POST /api/artifact/upload?mainArtifact=true`):

```yaml
asyncapi: 3.0.0
info:
  title: Repro Multi Message Examples
  version: 1.0.0
defaultContentType: application/json
channels:
  repro-channel:
    address: repro/channel
    bindings:
      ws:
        bindingVersion: 0.1.0
    messages:
      firstEvent:
        $ref: "#/components/messages/FirstEvent"
      secondEvent:
        $ref: "#/components/messages/SecondEvent"
operations:
  repro/channel:
    action: send
    channel:
      $ref: "#/channels/repro-channel"
    x-microcks-operation:
      frequency: 3
    messages:
      - $ref: "#/channels/repro-channel/messages/firstEvent"
      - $ref: "#/channels/repro-channel/messages/secondEvent"
components:
  messages:
    FirstEvent:
      name: FirstEvent
      payload:
        type: object
        properties:
          kind:
            type: string
      examples:
        - name: firstExample
          payload:
            kind: first
    SecondEvent:
      name: SecondEvent
      payload:
        type: object
        properties:
          kind:
            type: string
      examples:
        - name: secondExample
          payload:
            kind: second
```

Observed on `quay.io/microcks/microcks-uber:1.15.0` (latest) and `1.10.1`, each with the matching `microcks-uber-async-minion` (docker compose).

`GET /api/services/{id}?messages=true` shows only the last message's example survived the import:

```json
{
  "operation": "SEND repro/channel",
  "examples": ["secondExample"]
}
```

The async minion publishes only that example — on 1.15.0, 92 out of 92 published messages were `{"kind":"second"}`; `{"kind":"first"}` never appears:

```
Publishing on channel {/api/ws/Repro+Multi+Message+Examples/1.0.0/repro/channel}, message: {"kind":"second"}
Publishing on channel {/api/ws/Repro+Multi+Message+Examples/1.0.0/repro/channel}, message: {"kind":"second"}
```

**Order-dependence (rules out spec error or intentional design):** swapping the order of the two refs in the operation's `messages` array flips the survivor — `firstExample` is then the only one retained. Last-one-wins, whichever it is:

| Operation message order | Examples surviving import |
|---|---|
| firstEvent, secondEvent | `secondExample` only |
| secondEvent, firstEvent | `firstExample` only |
| same content as AsyncAPI 2.6 `oneOf` | both ✅ |

## Expected behavior

Both `firstExample` and `secondExample` are imported and the mock cycles through both — matching the AsyncAPI 2.x behaviour, where a `subscribe` operation with `message.oneOf` aggregates examples across all messages (verified working on the same stacks with an equivalent 2.6 spec).

## Additional context

Hit while evaluating a migration of a working AsyncAPI 2.6 spec (one channel, one operation, 7 messages with named examples, WS binding) to 3.x — the 2.6 version mocks all 7 events; the 3.x equivalent mocks only the last one. Multi-message operations are the v3 representation of the common 2.x `oneOf` event-channel shape, so most 2.x→3.x migrations with per-message examples would hit this.

dataGriff:
## Description

Fixes #2273 — importing an AsyncAPI 3.x document where one operation references multiple messages retained only the last message's examples; all others were silently dropped, so async mocks published a single event type.

`AsyncAPI3Importer.buildEventMessages` reassigned its result list on each iteration of the message loop. This change initialises the list once and accumulates with `addAll`, matching the v2 importer's behaviour for `message.oneOf`.

## Changes

- `AsyncAPI3Importer.buildEventMessages`: accumulate examples across all operation messages (2 lines)
- New test `testMultiMessagesAsyncAPI3ImportYAML` + resource spec with one `send` operation referencing two messages (2 + 1 examples): fails with `expected: <3> but was: <1>` before the fix, passes after; all existing `AsyncAPI3ImporterTest` tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/webapp/src/main/java/io/github/microcks/util/asyncapi/AsyncAPI3Importer.java
+++ b/webapp/src/main/java/io/github/microcks/util/asyncapi/AsyncAPI3Importer.java
@@ -421,7 +421,7 @@ public class AsyncAPI3Importer extends AbstractJsonRepositoryImporter implements
 
    /** Build a list of EventMessages from a "messages" Json node. */
    private List<EventMessage> buildEventMessages(JsonNode messagesNode, String defaultContentType) {
-      List<EventMessage> eventMessages = null;
+      List<EventMessage> eventMessages = new ArrayList<>();
       for (JsonNode messageRefNode : messagesNode) {
          JsonNode messageInChannelNode = followRefIfAny(messageRefNode);
          JsonNode msgNode = followRefIfAny(messageInChannelNode);
@@ -438,7 +438,7 @@ public class AsyncAPI3Importer extends AbstractJsonRepositoryImporter implements
          if (messageName != null && msgNode.has(EXAMPLES_NODE)) {
             // Compute a short message name if examples have no name attribute.
             messageName = messageName.substring(messageName.lastIndexOf("/") + 1);
-            eventMessages = buildEventMessageFromExamples(messageName, contentType, msgNode.get(EXAMPLES_NODE));
+            eventMessages.addAll(buildEventMessageFromExamples(messageName, contentType, msgNode.get(EXAMPLES_NODE)));
          }
       }
       return eventMessages;
--- a/webapp/src/test/java/io/github/microcks/util/asyncapi/AsyncAPI3ImporterTest.java
+++ b/webapp/src/test/java/io/github/microcks/util/asyncapi/AsyncAPI3ImporterTest.java
@@ -553,6 +553,65 @@ class AsyncAPI3ImporterTest {
    }
 
    @Test
+   void testMultiMessagesAsyncAPI3ImportYAML() {
+      AsyncAPI3Importer importer = null;
+      try {
+         importer = new AsyncAPI3Importer(
+               "target/test-classes/io/github/microcks/util/asyncapi/user-lifecycle-asyncapi-3.0-multi-messages.yaml",
+               null);
+      } catch (IOException ioe) {
+         Assertions.fail("Exception should not be thrown");
+      }
+
+      // Check that basic service properties are there.
+      List<Service> services = null;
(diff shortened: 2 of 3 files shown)


lbroudoux (approved):
It looks all good. Thank you!

PR #2294, opened 2026-09-09
fix(async-minion): support amqps:// for Amazon MQ TLS connectionsFix/async minion amqps amazon mq

### LINKED ISSUE #2233: async-minion: support `amqps://` for Amazon MQ (port 5671) and non-fatal AMQP startup

gozen59:
## Summary
- Auto-detect `amqps://` for port 5671 and honor explicit `amqp://` / `amqps://` schemes in `AMQPProducerManager`
- Required for Amazon MQ for RabbitMQ (TLS on port 5671)
- Document Helm and ECS configuration paths for `features.async.amqp.url` / `amqp.server`
- Keep fail-fast `@PostConstruct` behavior per maintainer feedback on #2247
Fixes #2233
## Problem
`AMQPProducerManager.createConnection()` hardcoded `amqp://`, preventing TLS connections to Amazon MQ on port 5671.
## Configuration
**Helm (Kubernetes):**
```yaml
features:
  async:
    amqp:
      url: xxxx.mq.eu-west-3.amazonaws.com:5671
      # or: amqps://xxxx.mq.eu-west-3.amazonaws.com:5671

### DIFF
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
@@ -86,8 +86,6 @@ public class AMQPProducerManager {
    protected Connection createConnection() throws Exception {
       ConnectionFactory factory = new ConnectionFactory();
       factory.setUri(buildAmqpUri());
-      factory.setConnectionTimeout(15000);
-      factory.setHandshakeTimeout(15000);
 
       if (amqpUsername != null && !amqpUsername.isEmpty() && amqpPassword != null && !amqpPassword.isEmpty()) {
          logger.infof("Connecting to AMQP broker with user '%s'", amqpUsername);
@@ -99,19 +97,29 @@ public class AMQPProducerManager {
    }
 
    /**
-    * Build the AMQP connection URI from configured server address. Honors explicit {@code amqp://} or {@code amqps://}
-    * schemes; auto-detects TLS for port 5671.
+    * Build the AMQP connection URI from configured server address.
     *
     * @return The connection URI for the RabbitMQ Java client
     */
-   String buildAmqpUri() {
-      if (amqpServer.contains("://")) {
-         return amqpServer;
+   private String buildAmqpUri() {
+      return resolveAmqpUri(amqpServer);
+   }
+
+   /**
+    * Resolve the AMQP connection URI from a server address. Honors explicit {@code amqp://} or {@code amqps://}
+    * schemes; auto-detects TLS for port 5671.
+    *
+    * @param server The configured AMQP server address
+    * @return The connection URI for the RabbitMQ Java client
+    */
+   static String resolveAmqpUri(String server) {
+      if (server.contains("://")) {
+         return server;
       }
-      if (amqpServer.contains(":5671")) {
-         return "amqps://" + amqpServer;
+      if (server.contains(":5671")) {
+         return "amqps://" + server;
       }
-      return "amqp://" + amqpServer;
+      return "amqp://" + server;
    }
 
    /**
--- a/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
+++ b/minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
@@ -85,7 +85,9 @@ public class AMQPProducerManager {
     */
    protected Connection createConnection() throws Exception {
       ConnectionFactory factory = new ConnectionFactory();
-      factory.setUri("amqp://" + amqpServer);
+      factory.setUri(buildAmqpUri());
+      factory.setConnectionTimeout(15000);
+      factory.setHandshakeTimeout(15000);
 
       if (amqpUsername != null && !amqpUsername.isEmpty() && amqpPassword != null && !amqpPassword.isEmpty()) {
          logger.infof("Connecting to AMQP broker with user '%s'", amqpUsername);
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java [hunk 1]
```diff
@@ -85,7 +85,9 @@ public void create() throws Exception {
     */
    protected Connection createConnection() throws Exception {
       ConnectionFactory factory = new ConnectionFactory();
-      factory.setUri("amqp://" + amqpServer);
+      factory.setUri(buildAmqpUri());
+      factory.setConnectionTimeout(15000);
+      factory.setHandshakeTimeout(15000);
```
lbroudoux: Seems to be big numbers here... Is it necessary? Any interest in making this configurable?

minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
```diff
@@ -96,6 +98,22 @@ protected Connection createConnection() throws Exception {
       return factory.newConnection(amqpClientId);
    }
 
+   /**
+    * Build the AMQP connection URI from configured server address. Honors explicit {@code amqp://} or {@code amqps://}
+    * schemes; auto-detects TLS for port 5671.
+    *
+    * @return The connection URI for the RabbitMQ Java client
+    */
+   String buildAmqpUri() {
```
lbroudoux: Please add a `private` marker here.

lbroudoux (changes requested):
Great! Thank you. Just got one question and one minor change request in review. 

Thanks!

minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
gozen59: Good point — these timeouts are not required for the amqps:// fix. The RabbitMQ Java client defaults (60s connection, 10s handshake) are sufficient. I've removed them to keep the change minimal; we can add configurable timeouts in a follow-up if needed.

minions/async/src/main/java/io/github/microcks/minion/async/producer/AMQPProducerManager.java
gozen59: I will do this

lbroudoux (approved):
Great!

PR #2302, opened 2026-09-11
fix(webapp): keep already encoded characters in proxied URLs

### LINKED ISSUE #2291: Pre-encoded proxy URLs are double-encoded
### Describe the bug

When an operation uses the `PROXY` or `PROXY_FALLBACK` dispatcher, Microcks builds the outbound URL from the configured `proxyUrl` (plus the incoming resource path and query string). If that URL already contains percent-encoded characters, Microcks encodes them a second time, so the percent sign itself is encoded:

* `%3A` → `%253A`
* `%2F` → `%252F`
* `%20` → `%2520`

This corrupts the target path and the proxied request fails on the backend. It happens for any backend whose resource identifiers must stay percent-encoded inside the path, for example:

```
https://backend.example.com/resources/urn%3Aorg%3Aexample%3Aservice%2Fitem-ABC123
```

is sent out as:

```
https://backend.example.com/resources/urn%253Aorg%253Aexample%253Aservice%252Fitem-ABC123
```

**Root cause:* `MockControllerCommons.getProxyUrlIfProxyIsNeeded()` builds the URI without telling Spring that the URL is [already encoded](https://github.com/microcks/microcks/blob/f0eee1d17aa70dc9574916837ca853d2398921f1/webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java#L136):

```java
return Optional.of(
      UriComponentsBuilder.fromUriString(externalUrl)
            .query(request.getQueryString())
            .build()      // == build(false)
            .toUri());
```

With `build(false)`, `UriComponents.toUri()` uses the multi-argument `java.net.URI` constructor, which quotes the components again — turning `%` into `%25`.

### Expected behavior

Percent-encoded characters present in the configured `proxyUrl` are preserved as-is: the outbound request path contains `%3A` and `%2F`, not `%253A` and `%252F`.

### Actual behavior

Every `%` is re-encoded to `%25`, so the path contains `%253A` / `%252F` and the backend rejects or mis-routes the request.

### How to Reproduce?

Reproducer:

1. Import an API (REST) into Microcks.
2. Set one operation dispatcher to `PROXY` (same happens with `PROXY_FALLBACK`) with a `proxyUrl` containing percent-encoded characters:

   ```yaml
   dispatcher: PROXY
   dispatcherRules: https://backend.example.com/resources/urn%3Aorg%3Aexample%3Aservice%2Fitem-ABC123
   ```

   A minimal REST API specification that exposes one `GET /resources/{id}` operation with the `PROXY` dispatcher above is enough; no specific artifact is required.

3. Invoke the mock for this operation (`GET /rest/<service>/<version>/resources/…`).
4. Observe the path of the outbound request: in the Microcks Live Traces
5. The observed URL contains `%253A` / `%252F` instead of `%3A` / `%2F`.

1.16.0-SNAPSHOT, 1.15.0

docker-compose

### Additional information

**Real-world case where this breaks the integration: AWS Bedrock AgentCore Runtime.** Its data-plane endpoint is:

```
POST https://bedrock-agentcore.{region}.amazonaws.com/runtimes/{escaped_arn}/invocations?qualifier=DEFAULT
```

and `{escaped_arn}` is the **URL-encoded runtime ARN as a single path segment** — `arn%3Aaws%3Abedrock-agentcore%3Aeu-west-2%3A000000000000%3Aruntime%2Fexample-ABC123`. 

References:

* Runtime HTTP protocol contract, which documents `/runtimes/{ESCAPED_ARN}/invocations`: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-http-protocol-contract.html
* `InvokeAgentRuntime` API reference: "When you use the agent ID instead of the full ARN, you don't need to URL-encode the identifier" — i.e. with the full ARN, encoding is required: https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_InvokeAgentRuntime.html
* AWS-published samples build the URL with `urllib.parse.quote(agent_arn, safe='')`, which yields `%3A` for `:` and `%2F` for `/`.

Consequences of the current Microcks behaviour for such a target:

1. The escaped ARN is sent as `arn%253Aaws%253A…%252Fexample-ABC123`, so AWS resolves a different (non-existent) runtime → `404` / `ResourceNotFoundException` ("Occurs when the specified agent runtime cannot be found").
2. Using the decoded ARN is not a workaround either: the `/` of `runtime/example-ABC123` becomes a real path separator, so the request no longer targets `/runtimes/{escaped_arn}/invocations`.

PunKHS:
## Description

Fixes #2291

With the `PROXY` or `PROXY_FALLBACK` dispatcher, a `proxyUrl` that already contains percent-encoded characters was
encoded a second time on the outbound request: `%3A` became `%253A`, `%2F` became `%252F`, and the query string had the same problem. Backends whose resource identifier must stay encoded inside a single path segment then resolve a
non-existent resource and return `404`.

`MockControllerCommons.getProxyUrlIfProxyIsNeeded()` built the URI with `UriComponentsBuilder.build()`, which marks the components as *not* encoded, so `toUri()` uses the multi-args `java.net.URI` constructor that quotes `%` again.

## Changes

- `MockControllerCommons.java`: build with `build(true)` the `proxyUrl`, the incoming resource path and the query string are already URI encoded, so they are assembled as-is. 

Fixes `PROXY` and `PROXY_FALLBACK` for the REST, SOAP and GraphQL invocation processors at once. Proxy URLs that are not valid URI syntax now hit the existing `Invalid external URL in the dispatcher` branch instead of being silently encoded.

## Tests

- `MockControllerCommonsTest.ProxyUrlTests` (7 tests): encoded `PROXY` url, resource path appending, encoded query
  string, encoded `PROXY_FALLBACK` url and the three no-proxy cases (response found, no proxy dispatcher, external url
  equal to the incoming one).
- `RestInvocationProcessorTest.shouldProxyRequestWithEncodedProxyUrl`: asserts the `URI` handed to
  `ProxyService.callExternal()` keeps its raw path and raw query.

```bash
mvn -pl webapp -am -Dtest=MockControllerCommonsTest,RestInvocationProcessorTest -Dsurefire.failIfNoSpecifiedTests=false test
```

Prepared with the assistance of an AI coding agent. Reviewed, tested and owned by the human author.

### DIFF
--- a/webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java
+++ b/webapp/src/main/java/io/github/microcks/web/MockControllerCommons.java
@@ -132,8 +132,8 @@ public class MockControllerCommons {
          externalUrl = externalUrl.replaceFirst("/$", "") + resourcePath;
          if (!externalUrl.contentEquals(request.getRequestURL())) {
             try {
-               return Optional.of(
-                     UriComponentsBuilder.fromUriString(externalUrl).query(request.getQueryString()).build().toUri());
+               return Optional.of(UriComponentsBuilder.fromUriString(externalUrl).query(request.getQueryString())
+                     .build(true).toUri());
             } catch (IllegalArgumentException ex) {
                log.warn("Invalid external URL in the dispatcher - {}", externalUrl);
             }
--- a/webapp/src/test/java/io/github/microcks/web/MockControllerCommonsTest.java
+++ b/webapp/src/test/java/io/github/microcks/web/MockControllerCommonsTest.java
@@ -15,11 +15,23 @@
  */
 package io.github.microcks.web;
 
+import io.github.microcks.domain.Response;
+import io.github.microcks.util.DispatchStyles;
 import io.github.microcks.util.delay.DelaySpec;
+import io.github.microcks.util.dispatcher.JsonMappingException;
+import io.github.microcks.util.dispatcher.ProxyFallbackSpecification;
+import jakarta.servlet.http.HttpServletRequest;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.springframework.http.HttpHeaders;
 
+import java.net.URI;
+import java.util.Optional;
+
 import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 /**
  * This is a test case for the MockControllerCommons class.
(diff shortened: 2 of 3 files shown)


SebastienDegodez (approved):

lbroudoux (approved):
This looks great! Thanks for taking the time to push the fix!!
However, it seems that one integration test is now failing... I re-launched the CI build to check if this is not a flaky test but maybe it's a side effect...

lbroudoux:
No errors ... that was a flaky integration test. Ready to merge! Thanks!

PR #2313, opened 2026-09-18
fix(webapp): avoid duplicate /services/search calls when filtering

### LINKED ISSUE #2297: Services list triggers duplicate search API calls when filtering by domain label
### Describe the bug

When filtering the services list by a `domain` label, the UI triggers redundant API calls and does not pass pagination parameters to the search endpoint.
Without a filter, the list correctly calls:
    GET /api/services?page=0&size=20
Once a domain filter is applied, the UI additionally calls:
    GET /api/services/search?labels.domain=<value>
This request is sent twice instead of making a single paginated search request.

### Expected behavior

When filtering by domain, a **single** request should be made:
    GET /api/services/search?labels.domain=example_domain

### Actual behavior

- `GET /api/services/search?labels.domain=example_domain` is called twice instead of once.
- Pagination parameters (`page`, `size`) are not passed to the `/search` endpoint when filtering.
- Inefficient behavior: duplicate requests and mixed usage of `/api/services` vs `/api/services/search`.

### How to Reproduce?

**Reproducer:**
1. Open the Microcks UI and navigate to the services list.
2. With no filter applied, open DevTools → Network tab.
3. Confirm a single call: `GET /api/services?page=0&size=20`.
4. Apply a filter on the `domain` label (e.g. `example_domain`).
5. Observe the network calls.

1.14.0

_No response_

### Additional information

**Root cause :**

In `webapp/src/main/webapp/src/app/pages/services/services.page.ts`:
- `route.queryParams.subscribe()` calls `filterServices()`
- `filterServices()` calls the API then `router.navigate()` with query params
- Navigation re-triggers `queryParams`, causing a duplicate `/api/services/search` call

celidrissi:
Signed-off-by: CHAFIK EL IDRISSI <chafik.elidrissi@outlook.fr>

### Description

- Fix duplicate `GET /api/services/search` call when filtering the services list by domain label: `filterServices()` both fetched the results and navigated to update the URL, and that navigation re-triggered the `queryParams` subscription which called `filterServices()` a second time with the same filter.
- Same issue when clearing an active filter (duplicate `GET /api/services` call) — fixed the same way.
- Fetching and URL navigation are now separated: `handleFilter()` only updates the URL, and the `queryParams` subscription remains the single place that triggers the actual API call.
- Extracted a `buildLabelsFilter()` helper to remove the duplicated label-map construction between `filterServices()` and `handleFilter()`.

**Testing**: Verified manually against a local backend (`docker-compose-devmode.yml`) + `ng serve` — checked the Network tab before/after for: applying a domain filter, changing the filter, and clearing it. Each action now triggers exactly one API call instead of two. `ng build --configuration=production` passes.

**AI usage disclosure**: I used Claude to help explain the root cause and review the code I wrote myself; no code was generated by AI.

### Related issue(s)

Fixes #2297


SebastienDegodez (approved):
LGTM

